PR #28, opened 2024-11-19
registers: autogen from files on-the-fly and I3C peripheral

swenson:
A common request is for people to want to check that RDL files work with our parser and generator but without having to mess with generator itself. So, we've added the ability to run additional RDL files (and addrmap instances) from the command-line.

We also added some documentation on how to use the RDL parser and generator, and how to use them for developing emulator peripherals.

As an example, we start writing the I3C emulator peripheral. This was originally written by @ArthurHeymans using manual Tock registers, but I converted it to use the new autogenerated root bus method.

This required us to modify the autogenerated code to contain the size of the read/write, since the I3C peripheral needs to know if 1, 2, or 4 bytes are being addressed.

We also deleted some extra autogenerated code that was left behind and ensure that we always clean up the directory in the future.


swenson:
Thanks!

PR #29, opened 2024-11-21
Compile with unaligned-scalar-mem feature

swenson:
This optimizes memory reads and writes so that they don't have to be aligned, as this is not required for our CPU.

This saves abot 3 KB of instruction space in the runtime binary.

The VeeR CPU also supports bit manipulation instructions, but we don't have complete support for those in the emulator, so I left that as a TODO. That saves a few hundred bytes more (for now).


helloxiling (approved):

swenson:
Thanks!

PR #30, opened 2024-11-21
Remove atty from dependencies

swenson:
This is built into the standard library now.

This resolves https://github.com/chipsalliance/caliptra-mcu-sw/security/dependabot/1


mlvisaya (approved):

swenson:
Thanks!

PR #31, opened 2024-11-21
Add xtask command to generate and verify flash image

mlvisaya:
Following the flash layout defined in
https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/docs/src/flash_layout.md

- Add new command to specify the firmware files and soc manifest to create 1 flash image file that can be burned to the SPI flash.

- Add new command to verify a flash image file

Example:

cargo run --bin flash-image create --caliptra-fw caliptra_fw.bin \
             --soc-manifest soc_manifest.bin \
             --mcu-runtime mcu_runtime.bin \
             --soc-images soc_image1.bin soc_image2.bin soc_image3.bin \
             --output flash_image.bin

cargo run --bin flash-image verify flash_image.bin


xtask/src/flash_image.rs
```diff
@@ -0,0 +1,504 @@
...
+    fn align_to_4_bytes(data: &mut Vec<u8>) {
+        let padding = (4 - (data.len() % 4)) % 4;
+        data.extend(vec![0; padding]);
+    }
+
+    pub fn write_to_file(&self, filename: &str) -> io::Result<()> {
+        let mut file = File::create(filename)?;
+
+        // Write header
+        file.write_all(&self.header.magic_number.to_le_bytes())?;
```
swenson, line 109: Rather than manually serializing (and deserializing), it would be better to use a library like `zerocopy` or `serde`. `zerocopy` is preferable, because we will want to move this into a library eventually that the firmware can use (where we are `no_std`), which means we will need to change the `Vec`s to slices as well.

xtask/src/flash_image.rs
```diff
@@ -0,0 +1,504 @@
...
+
+        Self {
+            header,
+            checksum,
+            payload,
+        }
+    }
+
+    fn align_to_4_bytes(data: &mut Vec<u8>) {
+        let padding = (4 - (data.len() % 4)) % 4;
```
swenson, line 101: For readability, I might suggest:

```suggestion
        let padding = data.len().next_multiple_of(4) - data.len();
```

xtask/src/flash_image.rs
```diff
@@ -0,0 +1,504 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::PROJECT_ROOT;
+    use std::fs::{self, File};
+    use std::io::Write;
+
+    /// Helper function to create a temporary file with specific content
+    fn create_temp_file(content: &[u8], file_name: &str) -> io::Result<String> {
```
swenson, line 299: Instead of managing these files yourself, I'd recommend using `tempfile::NamedTempFile` -- it should simplify things and ensure that the files are cleaned up.

xtask/src/flash_image.rs
```diff
@@ -0,0 +1,504 @@
...
+        if header_checksum != calculated_header_checksum {
+            return Err("Header checksum mismatch.")?;
+        }
+
+        if payload_checksum != calculated_payload_checksum {
+            return Err("Payload checksum mismatch.")?;
+        }
+
+        // Parse and verify image info and data
+        let mut offset = 16; // Start after header and checksums
```
swenson, line 166: Rather than having a `mut` here, could you just calculate `let offset = 16 + i * 12` in the `for`-loop?

xtask/src/flash_image.rs
```diff
@@ -0,0 +1,504 @@
...
+
+#[derive(Clone)]
+pub struct FirmwareImage {
+    identifier: u32,
+    data: Vec<u8>,
+}
+
+impl FirmwareImage {
+    pub fn new(identifier: u32, filename: &str) -> io::Result<Self> {
+        let mut file = File::open(filename)?;
```
swenson, line 53: It would be useful to have an error message printed here so that we know which file failed to open when running this.

xtask/src/flash_image.rs
```diff
@@ -0,0 +1,504 @@
+// Licensed under the Apache-2.0 license
+
+use crate::DynError;
+use crc32fast::Hasher;
+use std::fs::File;
+use std::io::{self, Read, Write};
+use std::io::{Error, ErrorKind};
```
swenson, line 7: I think you can can combine these imports.

xtask/src/flash_image.rs
```diff
@@ -0,0 +1,504 @@
...
+
+const FLASH_IMAGE_MAGIC_NUMBER: u32 = 0x464C5348;
```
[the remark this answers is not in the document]
mlvisaya: that's true, this should be big endian

xtask/src/flash_image.rs
mlvisaya: thanks. i'll try this one.

xtask/src/flash_image.rs
mlvisaya: ack

xtask/src/flash_image.rs
mlvisaya: ack

xtask/src/flash_image.rs
mlvisaya: yes, will do

xtask/src/flash_image.rs
mlvisaya: I'll use zerocopy

xtask/src/flash_image.rs
```diff
@@ -151,8 +118,8 @@ impl FlashImage {
         // Parse and verify checksums
         let header_checksum = u32::from_le_bytes(image[8..12].try_into().unwrap());
         let payload_checksum = u32::from_le_bytes(image[12..16].try_into().unwrap());
-        let calculated_header_checksum = FlashImageChecksum::calculate_checksum(&image[0..8]);
-        let calculated_payload_checksum = FlashImageChecksum::calculate_checksum(&image[16..]);
+        let calculated_header_checksum = calculate_checksum(&image[0..8]);
+        let calculated_payload_checksum = calculate_checksum(&image[16..]);
```
swenson, line 122: If these were parsed with `zerocopy` you could just use their `calculate_checksum()` methods.

xtask/src/flash_image.rs
```diff
@@ -190,85 +158,116 @@ impl FlashImage {
...
+
+    Ok(buffer)
 }
 
 pub(crate) fn flash_image_create(
     caliptra_fw_path: &str,
     soc_manifest_path: &str,
     mcu_runtime_path: &str,
-    soc_image_paths: &Option<Vec<String>>,
+    soc_image_paths: &Option<Vec<&str>>,
```
swenson, line 217: I think it's okay if this is `&Option<Vec<String>>` or `&Option<Vec<PathBuf>>` so that we don't need to have the conversion function below.

xtask/src/flash_image.rs
```diff
@@ -3,90 +3,75 @@
...
 use crc32fast::Hasher;
 use std::fs::File;
-use std::io::{self, Read, Write};
-use std::io::{Error, ErrorKind};
+use std::io::{self, Error, ErrorKind, Read, Write};
+use zerocopy::Immutable;
+use zerocopy::{byteorder::U32, FromBytes, IntoBytes};
 
-const FLASH_IMAGE_MAGIC_NUMBER: u32 = 0x464C5348;
+const FLASH_IMAGE_MAGIC_NUMBER: u32 = u32::from_be_bytes([b'F', b'L', b'S', b'H']);
```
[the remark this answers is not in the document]
mlvisaya: good suggestion

xtask/src/flash_image.rs
mlvisaya: sounds good. i can definitely parse these using zerocopy, although the payload itself can not be parsed since it has variable length.

xtask/src/flash_image.rs
```diff
@@ -3,90 +3,75 @@
...
 
-pub struct FlashImage {
+pub struct FlashImage<'a> {
     header: FlashImageHeader,
     checksum: FlashImageChecksum,
-    payload: FlashImagePayload,
+    payload: FlashImagePayload<'a>,
 }
 
+#[repr(C)]
```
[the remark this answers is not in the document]
mlvisaya: yeah, looking at the documentation, looks like we need the packed trait to stop it from adding padding between fields.

xtask/src/flash_image.rs
```diff
@@ -97,46 +82,28 @@ impl FlashImage {
...
+            file.write_all(image.data)?;
         }
 
         Ok(())
     }
 
     pub fn verify_flash_image(image: &[u8]) -> Result<(), DynError> {
         // Parse and verify header
-        let magic_number = u32::from_le_bytes(image[0..4].try_into().unwrap());
+        let magic_number = u32::from_be_bytes(image[0..4].try_into().unwrap());
```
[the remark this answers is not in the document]
mlvisaya: ack

xtask/src/flash_image.rs
mlvisaya: ack

swenson (approved):
LGTM

PR #32, opened 2024-11-22
i3c: Start basic driver

swenson:
This does the initialization sequence, routes interrupts, and does passes reads and writes to the clients.

I haven't had a chance to do much testing yet, but that will be easier once the peripheral is more complete.


parvathib (changes requested):

parvathib (approved):

swenson:
Thanks!

PR #33, opened 2024-11-25
emulator: Add Zba,Zbc,Zbs instruction sets

swenson:
And enables support in code generation to use them. This saves ~500 bytes or so of instruction space for now (and more in the future).

This has the negative side effect of making tests a little slower to run due to the quantity of macros used in generating all of the test cases.

This work extends the previous Zbb bit manipulation instruction set we supported to include the rest of the `b` family, all of which are supported by our VeeR CPU.

We also add missing tests for the Zbb instruction set.

We added support for the disassembler as well, including for RV64 (even though the emulator is only RV32).


helloxiling (approved):

swenson:
Thanks!

PR #34, opened 2024-11-25
Add dummy flash controller emulation

helloxiling:
This PR introduces the emulation of a dummy flash controller peripheral, with backup storage implemented as a file. The registers are defined in `flash_ctrl.rdl` and the peripheral trait is generated using `cargo xtask registers-autogen`. It supports single-page based read, write, and erase operations. The primary purpose of this peripheral is to aid in the development and testing of the flash driver capsule under TockOS.


hw/flash_ctrl.rdl
```diff
@@ -0,0 +1,83 @@
...
+            desc = "This register ensures the contents of CONTROL cannot be changed by software once a flash operation has begun. \
+                    This bit defaults to 1 and is set to 0 by hardware when flash operation is initiated. When the controller completes the flash operation, \
+                    It unlocks whenever the existing flash operation completes, regardless of success or error.";
+
+        } EN[0:0];
+    } CTRL_REGWEN @ 0x1C;
+};
+
+addrmap flashctrl_dummy {
+    flash_ctrl flash_ctrl @ 0x7000_0000;
```
swenson, line 82: To keep the address space a little neater, let's put this somewhere in in `0x2000_0000` or `0x3000_0000`.

hw/flash_ctrl.rdl
```diff
@@ -0,0 +1,83 @@
...
+            desc = "The page number for read, write, erase operations";
+        } PAGE_NUM[31:0];
+    } PAGE_NUM @ 0x0C;
+
+    reg {
+        field {
+            sw = rw;
+            desc = "The page buffer address for read/write operations";
+        } PAGE_ADDR[31: 0];
+    } PAGE_ADDR @ 0x10;
```
swenson, line 44: So, this is like a DMA controller? We program the address to read to / write from, then use `CONTROL` to start the read, write, or erase, then check the status with `OP_STATUS`?

emulator/periph/src/flash_ctrl.rs
```diff
@@ -0,0 +1,1096 @@
...
+pub enum FlashOpError {
+    ReadError = 0,
+    WriteError = 1,
+    EraseError = 2,
+    InvalidOp = 3,
+}
+
+// Define a struct to represent the flash controller peripheral emulation.
+// Includes IRQs, timers, and peripheral registers, backend storage and a buffer page_size for the flash controller.
+#[derive(Bus)]
```
swenson, line 63: If you implement `FlashPeripheral`, you shouldn't need to use `derive(Bus)` -- the `FlashPeripheral` wrapper in the `AutoRootBus` will take care of the `Bus` methods for you.

That means you can probably remove the `ReadWriteRegister<u32>`, etc., below if you want and just use `u32` or `Cell<u32>` (for example)

emulator/periph/src/flash_ctrl.rs
```diff
@@ -0,0 +1,1096 @@
...
+        _size: emulator_types::RvSize,
+        _val: emulator_bus::ReadWriteRegister<
+            u32,
+            registers_generated::flash_ctrl::bits::Control::Register,
+        >,
+    ) {
+        self.control.reg.set(_val.reg.get());
+
+        // Set the operation_pending flag to true
+        self.operation_pending = true;
```
swenson, line 427: You'll also want to set the `timer` here so that you are sure that the `poll()` method is called at the appropriate time.

hw/flash_ctrl.rdl
helloxiling: Sure, will update in next patch.

hw/flash_ctrl.rdl
helloxiling: Yes, exactly! I had intended to use the FIFO register for data transfer, but that adds extra complexity for this kind of dummy controller.

emulator/periph/src/flash_ctrl.rs
```diff
@@ -0,0 +1,1096 @@
...
+    fn write_page(&mut self) -> Result<(), FlashOpError> {
+        // Get the page number from the register
+        let page_num = self.page_num.reg.get();
+
+        // Sanity check for the page number, page size and file
+        if page_num >= Self::MAX_PAGES
+            || self.page_size.reg.get() < Self::PAGE_SIZE as u32
+            || self.file.is_none()
+        {
+            println!("Write page failed sanity check");
```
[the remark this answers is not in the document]
helloxiling: yeah, will remove in next patch.

emulator/periph/src/flash_ctrl.rs
```diff
@@ -0,0 +1,1096 @@
...
+    fn write_ctrl_regwen(
+        &mut self,
+        _size: emulator_types::RvSize,
+        _val: emulator_bus::ReadWriteRegister<
+            u32,
+            registers_generated::flash_ctrl::bits::CtrlRegwen::Register,
+        >,
+    ) {
+        // Unreacheable
+        panic!("Write to read-only ctrl_regwen register");
```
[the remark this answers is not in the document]
helloxiling: This routine is still generated automatically. Another thing I observed is that other peripherals with same register name such as `CONTROL` got modified when running `cargo xtask registers-autogen`.  I revert those changes manually. Will debug and fix that in the script later.

emulator/periph/src/flash_ctrl.rs
helloxiling: Hi @swenson, thanks for catching this issue.  
Do we prefer normal variable `u32` to `ReadWriteRegister<u32>`  in this context?
It seems that `ReadWriteRegister<u32>` etc. still works by importing `use emulator_bus::{ReadOnlyRegister, ReadWriteRegister};` event without `derive(bus)`.  Please advise if I have to refactor it to `u32`.

emulator/periph/src/flash_ctrl.rs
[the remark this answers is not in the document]
swenson: Names need to be globally unique, otherwise the script can get a bit confused. So, maybe call it `FLASH_CONTROL`?

emulator/periph/src/flash_ctrl.rs
swenson: I would say use whichever is easier for you.

emulator/periph/src/flash_ctrl.rs
```diff
@@ -58,50 +57,21 @@ pub enum FlashOpError {
...
-
-    /// Internal buffer for read, write operations
     buffer: Vec<u8>,
-
-    /// Operation pending flag
-    operation_pending: bool,
-
-    /// Error interrupt
+    operation_start: Option<ActionHandle>,
     _error_irq: Irq,
```
swenson, line 74: Since we're using these, we can remove the underscore from the beginning. (The underscore is just to prevent the warnings when they are unused.)

emulator/periph/src/flash_ctrl.rs
```diff
@@ -306,74 +299,78 @@ impl DummyFlashCtrl {
...
     ) {
+        if self.interrupt_state.reg.is_set(FlInterruptState::Error)
+            && _val.reg.is_set(FlInterruptEnable::Error)
+        {
+            self._error_irq.set_level(true);
+            self.timer.schedule_poll_in(1);
+        }
+
+        if self.interrupt_state.reg.is_set(FlInterruptState::Event)
+            && _val.reg.is_set(FlInterruptEnable::Event)
```
swenson, line 368: If you're using the `_val`, best to rename it to `val`.

emulator/periph/src/flash_ctrl.rs
```diff
@@ -306,74 +299,78 @@ impl DummyFlashCtrl {
...
         _val: emulator_bus::ReadWriteRegister<
             u32,
-            registers_generated::flash_ctrl::bits::InterruptEnable::Register,
+            registers_generated::flash_ctrl::bits::FlInterruptEnable::Register,
         >,
     ) {
+        if self.interrupt_state.reg.is_set(FlInterruptState::Error)
+            && _val.reg.is_set(FlInterruptEnable::Error)
+        {
+            self._error_irq.set_level(true);
```
swenson, line 363: You also want to check if the IRQ is current set, but the `val` is now `CLEAR`, you want to `set_level(false)` to stop raising the interrupt.

swenson (approved):
LGTM

A few minor things.

PR #35, opened 2024-11-26
emulator: Add RISC-V compliance nightly tests

swenson:
This enables the RISC-V compliance tests to be run nightly.

It also enables Zbb (bit manipulation) tests.


.github/workflows/compliance.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Install RISC-V Python tools
+      run: |
+        pip install git+https://github.com/riscv/riscv-isac.git@dev && \
+        pip install git+https://github.com/riscv/riscof.git
+    - name: Install riscv-isa-sim
+      run: |
+        git clone https://github.com/riscv-software-src/riscv-isa-sim.git && \
+        cd riscv-isa-sim && \
+        mkdir build && \
+        cd build && \
```
mlvisaya, line 33: not a blocker: but could use pushd and popd instead of cd, so that we don't have to care about the relative paths when going back.

swenson: Good call. I just copied this from a previous repo, but I can make that modification.

mlvisaya (approved):
LGTM

mlvisaya:
all good, thanks

swenson:
Thanks!

PR #36, opened 2024-11-27
Add Mux, transport binding and ctrl msg handling layers of MCTP capsule

parvathib:
This PR adds the basic structure for the Mux MCTP driver layer, the transport binding layer, and the handling of MCTP control messages. 
At this stage, the code remains untested, and the modules have yet to be instantiated.
The implementation includes the Set EID and Get EID commands, with plan to add other mandatory managed MCTP device commands once the basic commands have been tested and validated.


runtime/capsules/runtime/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+# Licensed under the Apache-2.0 license
```
swenson, line 1: I noticed the directory is `runtime/capsules/runtime`. That seems a bit redundant :)

Maybe `runtime/capsules` or, we think we will have different kinds of capsules in the future, `runtime/capsules/devices` or `drivers` (or something).

runtime/capsules/runtime/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[package]
+name = "capsules-runtime"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+
+[dependencies]
```
swenson, line 9: I would put the dependencies under `[target.'cfg(target_arch = "riscv32")'.dependencies]` so that the build stays clean when you run `cargo build`. You might have to put a target config attribute above `pub mod mctp` though.

runtime/capsules/runtime/src/mctp/mctp_binding.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        crc = MCTPI3CBinding::crc8(crc, addr);
+
+        for byte in buf.iter().take(len) {
+            crc = MCTPI3CBinding::crc8(crc, *byte);
+        }
+        crc
+    }
+
+    fn crc8(crc: u8, data: u8) -> u8 {
```
swenson, line 95: I would definitely like to see some unit tests for this. If there are SMBus test vectors, that would be ideal.

runtime/capsules/runtime/src/mctp/mctp_binding.rs
```diff
@@ -0,0 +1,204 @@
...
+pub const MCTP_I3C_MAXBUF: usize = 69; // 4 MCTP header + 64 baseline payload + 1 (PEC)
+
+pub const MCTP_I3C_MAXMTU: usize = MCTP_I3C_MAXBUF - 1; // 64 bytes
+pub const MCTP_I3C_MINMTU: usize = 4 + 64; // 4 MCPT header + 64 baseline payload
+
+/// This trait contains the interface definition
+/// for sending the MCTP packet through MCTP transport binding layer.
+pub trait MCTPTransportBinding<'a> {
+    /// Set the client that will be called when the packet is transmitted.
+    fn set_tx_client(&self, client: &'a dyn TxClient);
```
swenson, line 22: I wonder if we should use our own separate `TxClient` and `RxClient` interfaces for MCTP that are different from the I3C ones.

Specifically, the I3C ones have the weird thing where reads and writes are backwards (since for I3C, we transmit read responses and receive writes).

It also helps keeps the layers a little more independent. (In software engineering terms, we call the layers blending together too much to be a "leaky abstraction".)

runtime/capsules/runtime/src/mctp/mctp_control.rs
```diff
@@ -0,0 +1,276 @@
...
+    SetEid,
+    GetEid,
+    GetMsgTypeSupport,
+    Unsupported,
+}
+
+impl From<u8> for MCTPCtrlCmd {
+    fn from(val: u8) -> MCTPCtrlCmd {
+        match val {
+            1 => MCTPCtrlCmd::SetEid,
```
swenson, line 50: I always think "Eid Mubarak" when I see this :)

swenson (commented):
Looking good so far. A few comments.

runtime/capsules/runtime/Cargo.toml
parvathib: Right! changed to `runtime/capsules`

runtime/capsules/runtime/Cargo.toml
parvathib: done

runtime/capsules/runtime/src/mctp/mctp_control.rs
parvathib: Haha! I could use EID instead of Eid. Or maybe call it EndpointID, but that seems a bit too long.

runtime/capsules/runtime/src/mctp/mctp_binding.rs
parvathib: Although they look similar, I introduced new traits for the transport binding layer to distinguish them from those in HIL.

runtime/capsules/runtime/src/mctp/mctp_binding.rs
parvathib: added unit tests for PEC calculation with some dummy MCTP over I3C packets

runtime/capsules/runtime/src/mctp/mctp_control.rs
swenson: `Eid` is fine :) Just funny.

PR #41, opened 2024-11-28
I3C peripheral updates and integration tests

swenson:
This progresses more on the I3C peripheral, correctly queueing writes how the hardware does.

We also start on writing a socket protocol so that another thread or program can send reads and writes to the I3C bus. This hasn't been teste yet though.

I noticed two bugs that I will address in separate PRs:
* The I3C interrupts are not quite working yet. When enabled, they are routed correctly to the kernel handler, but keep getting retriggered and not handled properly. I've disabled the interrupts for now, so the read/write methods need to be called manually.
* Alarms are not firing properly. I tried to work around the interrupts not handling correctly by scheduling an alarm to call the read/write methods periodically, but the alarms never fire. I've noticed something similar to this before. I will debug this separately.

The biggest piece though is adding new integration tests for the kernel:

* A test can be added to `tests/integration/src/lib.rs`
* This will build the emulator and firmware with the feature equal to the test name, and then run them.
* The test can exit with a non-zero exit code to signal failure.

The tests need to built sequentially so that they don't conflict when writing the `layout.ld` linker script (this has to have the same name each time to avoid modifying `RUSTFLAGS` with every invocation; otherwise dependencies are recompiled every single time).

### DIFF
--- a/emulator/periph/src/i3c.rs
+++ b/emulator/periph/src/i3c.rs
@@ -152,6 +152,21 @@ impl I3cPeripheral for I3c {
         self.interrupt_status.clone()
     }
 
+    fn write_i3c_ec_tti_interrupt_status(
+        &mut self,
+        _size: emulator_types::RvSize,
+        val: emulator_bus::ReadWriteRegister<
+            u32,
+            registers_generated::i3c::bits::InterruptStatus::Register,
+        >,
+    ) {
+        let current = self.interrupt_status.reg.get();
+        let new = val.reg.get();
+        // clear the interrupts that are set
+        self.interrupt_status.reg.set(current & !new);
+        self.check_interrupts();
+    }
+
     fn write_i3c_ec_tti_interrupt_enable(
         &mut self,
         _size: emulator_types::RvSize,
--- a/emulator/periph/src/i3c.rs
+++ b/emulator/periph/src/i3c.rs
@@ -12,10 +12,10 @@ use emulator_bus::{Clock, ReadWriteRegister, Timer};
 use emulator_cpu::Irq;
 use emulator_registers_generated::i3c::I3cPeripheral;
 use emulator_types::{RvData, RvSize};
-use registers_generated::i3c::bits::InterruptEnable;
-use registers_generated::i3c::bits::InterruptStatus;
-use registers_generated::i3c::bits::StbyCrDeviceAddr;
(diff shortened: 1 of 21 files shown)


emulator/periph/src/i3c.rs
```diff
@@ -7,12 +7,17 @@ Abstract:
...
 use crate::i3c_protocol::{I3cController, I3cTarget, I3cTcriResponseXfer, ResponseDescriptor};
-use emulator_bus::{Clock, Timer};
+use crate::DynamicI3cAddress;
+use emulator_bus::{Clock, ReadWriteRegister, Timer};
 use emulator_cpu::Irq;
 use emulator_registers_generated::i3c::I3cPeripheral;
 use emulator_types::{RvData, RvSize};
+use registers_generated::i3c::bits::InterruptEnable;
+use registers_generated::i3c::bits::InterruptStatus;
+use registers_generated::i3c::bits::StbyCrDeviceAddr;
```
helloxiling, line 17: minor:  is it better to combine them into a single use statement?

emulator/periph/src/i3c.rs
```diff
@@ -23,15 +28,20 @@ pub struct I3c {
...
     /// RX DATA in u8
-    tti_rx_data_raw: VecDeque<u8>,
+    tti_rx_data_raw: VecDeque<Vec<u8>>,
+    /// RX DATA currently being read from driver
+    tti_rx_current: Vec<u8>,
     /// TX Command in u32
     tti_tx_desc_queue_raw: VecDeque<u32>,
     /// TX DATA in u32
-    tti_tx_data_raw: VecDeque<u8>,
+    tti_tx_data_raw: VecDeque<Vec<u8>>,
```
helloxiling, line 37: Minor: comment should be updated.  `/// TX DATA in u8`

runtime/i3c/src/core.rs
```diff
@@ -273,18 +273,10 @@ impl<'a, A: Alarm<'a>> I3CCore<'a, A> {
             }
```
helloxiling, line 273: Is `interrupt-status` register a regular type or `write1toclear`?  If it is `write1toclear`, I think we should change this piece of code accordingly.

swenson: Ah, you're correct, the docs say "woclr".

helloxiling (approved):

swenson:
Thanks!

PR #42, opened 2024-11-28
Use crossterm instead of console

### LINKED ISSUE #38: Stop corrupting the terminal when running the emulator
The emulator (or the Tock kernel) seems to leave the terminal in a bad state after running.

We need to investigate what terminal setting is being manipulated so that we can restore it when the emulator exits.

swenson:
So that the terminal doesn't get corrupted after running the emulator.

Fixes #38


swenson:
@helloxiling There's a long-standing issue with `console` where it leaves the terminal in a bad state when you hit Ctrl-C https://github.com/console-rs/console/issues/172

The easiest way around this is to use a different crate that handles that edge case a little better. I did some testing and `crossterm` doesn't seem to have the same issue.

helloxiling (approved):

swenson:
Thanks!

PR #43, opened 2024-11-28
registers: Honor read-only / write-only

### LINKED ISSUE #37: Read-only registers have write methods generated automatically
In #34, the RDL includes `CTRL_REGWEN`, which is read-only, but has write methods generated in its trait.

The autogenerator should not generate the write methods if there no writable fields.

swenson:
If a field is read-only, we shouldn't generate write methods for it. Likewise if a field is write-only, we shouldn't generate read methods for it.

Fixes #37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ runtime/apps/layout.ld
 
 # mdbook
 book
+
+# flash file
+dummy_flash.bin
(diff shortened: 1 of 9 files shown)


helloxiling (approved):
Thanks for fixing this. It looks good to me.

swenson:
Thanks!

PR #45, opened 2024-12-03
Speed up compiles and clean up RUSTFLAGS

### LINKED ISSUE #39: Compiles are too slow
When compiling and building the firmware, the code is recompiled too many times.

Some of this is due to how we modify the `RUSTFLAGS` when building the firmware -- this triggers dependencies to be rebuilt.

We should figure out a better way to do these builds without constantly recompiling every dependency.

swenson:
Before:

```
cargo t  14.69s user 7.03s system 161% cpu 13.420 total
```

After:

```
cargo t  12.20s user 5.38s system 182% cpu 9.612 total
```

We consolidate RUSTFLAGS so that they are entirely in `.cargo/config.toml` -- this speeds up compiles since we aren't modifying `RUSTFLAGS` as part of `cargo xtask *-build` (which triggers recompiles). This also means we only need to worry about the flags in `.cargo/config.toml`.

Putting all of the `RUSTFLAGS` in `.cargo/config.toml` is mostly equivalent to passing them as an environment variable: there was a 100 byte difference in the application build, but it's not clear why.

We also add a filesystem cache of the result of finding the `sysroot`. For some reason, repeated calculation of it can be expensive (adds multiple seconds to the build).

Fixes #39

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,7 +8,33 @@ rustflags = [
     "-C",
     "panic=abort",
     "-C",
-    "target-feature=+relax,+unaligned-scalar-mem,+b",
+    "target-feature=+relax,+unaligned-scalar-mem,+zba,+zbb,+zbc,+zbs",
     "-C",
     "force-frame-pointers=no",
+    # See https://github.com/tock/tock/pull/2853
+    "-C",
+    "relocation-model=static",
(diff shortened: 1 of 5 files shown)


mlvisaya (approved):
lgtm

swenson:
Thanks!

PR #46, opened 2024-12-04
Add MCTP control command tests

parvathib:
This PR adds MCTP mux component and instantiate it in the board configuration

Allocate resources just enough for instantiating the MCTP MUX driver. There'll be new additions in the future.
The static reference in the platform structure is temporarily added until the driver layer is implemented.
This should set up the stack for testing the MCTP control messages with integration test framework.


emulator/app/Cargo.toml
```diff
@@ -25,8 +25,13 @@ gdbstub.workspace = true
 hex.workspace = true
 tock-registers.workspace = true
 zerocopy.workspace = true
+bitfield.workspace = true
+crc = "3.2.1"
```
swenson, line 29: minor nit: this is the same package that is used in `emulator/bmc/pldm-fw-pkg/Cargo.toml`, so we might as well consolidate them into the main `Cargo.toml` so they don't accidentally get out of sync.

emulator/app/src/i3c_socket.rs
```diff
@@ -133,26 +139,250 @@ fn handle_i3c_socket_connection(
...
+}
+
+pub(crate) fn run_tests(
+    running: Arc<AtomicBool>,
+    port: u16,
+    target_addr: DynamicI3cAddress,
+    tests: Vec<Test>,
+) {
+    let running_clone = running.clone();
+    let addr = SocketAddr::from(([127, 0, 0, 1], port));
```
swenson, line 172: This can be addressed in another PR, but since we are only communicating with the socket locally and not from another program, it might be easier to use the `mpsc` channels directly so that we don't have to worry about the ports. (Otherwise, we're going to run into issues where there is a port conflict since we hard code the port and don't check if it's already in use.)

emulator/app/src/tests/mctp_ctrl_cmd.rs
```diff
@@ -0,0 +1,212 @@
+// Licensed under the Apache-2.0 license
+
+// use crc;
```
swenson, line 3: Not needed anymore?
```suggestion
```

emulator/app/src/tests/mctp_util/ctrl_protocol.rs
```diff
@@ -0,0 +1,227 @@
+// Licensed under the Apache-2.0 license
+
+use bitfield::bitfield;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+pub const MCTP_CTRL_MSG_HDR_SIZE: usize = 2;
+
+#[macro_export]
+macro_rules! set_eid_req {
```
swenson, line 9: These macros feel like they could just be simple functions?

Like `fn set_eid_req(op: u8, eid: u8) -> Vec<u8>`?

runtime/i3c/src/core.rs
```diff
@@ -21,6 +21,9 @@ use registers_generated::i3c::I3C_CSR_ADDR;
 use tock_registers::register_bitfields;
 use tock_registers::LocalRegisterCopy;
 
+use core::fmt::Write;
+use romtime::println;
```
swenson, line 25: I don't think these imports are used anymore?

emulator/periph/src/i3c_protocol.rs
```diff
@@ -72,6 +72,7 @@ impl I3cController {
         let running = self.running.clone();
         let targets = self.targets.clone();
         let counter = self.incoming_counter.clone();
+        println!("Starting I3C controller thread");
```
swenson, line 75: Maybe we don't need this line?

emulator/periph/src/i3c.rs
```diff
@@ -179,8 +184,9 @@ impl I3c {
             }
             // we only need the first byte, which is the MDB.
             // TODO: handle more than the MDB?
+            // Drain IBI descriptor size + 4 bytes (MDB)
             self.i3c_target.send_ibi(self.tti_ibi_buffer[4]);
-            self.tti_ibi_buffer.drain(0..len + 4);
+            self.tti_ibi_buffer.drain(0..8);
```
swenson, line 189: Theoretically this length could be more, so maybe `(len + 4).next_multiple_of(4)`?

emulator/app/src/i3c_socket.rs
```diff
@@ -133,26 +139,250 @@ fn handle_i3c_socket_connection(
...
+        command: cmd_words,
+    };
+    transmute!(cmd_hdr)
+}
+
+fn prepare_private_read_cmd(to_addr: u8) -> [u8; 9] {
+    let mut read_cmd = ReguDataTransferCommand::read_from_bytes(&[0; 8]).unwrap();
+    read_cmd.set_rnw(1);
+    read_cmd.set_data_length(0);
+    let cmd_words: [u32; 2] = transmute!(read_cmd);
```
swenson, line 372: I'm a little worried about endianness biting us, especially in the (unlikely) event we ever run this code with a big endian host.

Could we add a test at least that checks that the packets generated have the bytes that we expect? (That way if something weird happens with endianness in the future, we'll at least have a test to let us know.)

emulator/app/Cargo.toml
parvathib: Done

emulator/app/src/i3c_socket.rs
parvathib: You are right! I did face a port conflict once while testing. I'll change to `mpsc` in the next PR when I add more tests.

emulator/app/src/tests/mctp_ctrl_cmd.rs
parvathib: removed.

runtime/i3c/src/core.rs
parvathib: removed

emulator/periph/src/i3c_protocol.rs
parvathib: removed

emulator/periph/src/i3c.rs
parvathib: That sounds correct! Changed!

emulator/app/src/i3c_socket.rs
parvathib: Good idea! Added tests for prepare_private_read and prepare_private_write functions.

emulator/app/src/tests/mctp_util/ctrl_protocol.rs
parvathib: changed to functions

swenson (approved):
LGTM

PR #48, opened 2024-12-05
Add documentation for Image Loading

mlvisaya:


docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+
+Pre-image loading state:
+
+* Caliptra MCU Runtime Firmware (MCU RT) is running in the MCU.
+* The SOC target components (Vendor CPU 1 and 2) are held in reset.
+
+Image Loading Steps:
+
+1. MCU RT checks for available SOC images in the [flash storage](./image_loading.md).
+2. MCU RT authorizes the [SOC manifest](https://github.com/chipsalliance/caliptra-sw/blob/main/auth-manifest/README.md) and the SOC images by following steps described in [Manifest-Based Authorization](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime/README.md#manifest-based-image-authorization-new-in-12) section of the Caliptra Runtime Specification.
```
helloxiling, line 25: Will MCU RT authorize the SOC manifest directly or forward the authentication request to Caliptra core?

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+
+```rust
+/// Retrieves an Image Metadata Entry by `firmware_id`.
+///
+/// # Arguments
+/// - `firmware_id`: The unique identifier for the image.
+///
+/// # Returns
+/// The Image Metadata Entry if it exists, or an error if it is not found.
+pub fn get_image_metadata(firmware_id: u32) -> Result<ImageMetadataEntry,DynError>;
```
helloxiling, line 75: I suppose those APIs should be async. @swenson, please help confirm.

docs/src/image_loading.md
mlvisaya: After retrieving the SOC manifest from flash, the MCU RT application will call another API to authenticate and forward the manifest to the Caliptra core.

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+The Image Loading module is a component of the MCU Runtime SDK designed for managing SOC images stored in flash memory. This module provides APIs for:
+
+- Detecting SOC images availability in the [flash storage](./image_loading.md).
+- Parsing the [SOC manifest](https://github.com/chipsalliance/caliptra-sw/blob/main/auth-manifest/README.md).
+- Loading SOC images to a target component.
+
+The diagram below illustrates an **example** how the Caliptra subsystem, integrated within the SOC, is used to load SOC images to vendor components. In this example, these are the 2 Vendor CPUs and their corresponding RAMs.
+
+<p align="center">
+    <img src="images/image_loading_sample.svg" alt="flash_config" width="50%">
```
swenson, line 14: Could we get this to match the style a bit closer to what is in the other documents (the other diagram is closer)? I can help with this if you provide me the source.

But specifically:

* Text should be larger so that is legible (it is currently a bit small)
* No gradients, just use simple, soft colors
* No white text on a dark background

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+</p>
+
+Pre-image loading state:
+
+* Caliptra MCU Runtime Firmware (MCU RT) is running in the MCU.
+* The SOC target components (Vendor CPU 1 and 2) are held in reset.
+
+Image Loading Steps:
+
+1. MCU RT checks for available SOC images in the [flash storage](./image_loading.md).
```
swenson, line 24: I think you should explicitly add steps for downloading the SOC manifest, downloading the images from flash, and uploading them to the SOC components.

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+2. MCU RT authorizes the [SOC manifest](https://github.com/chipsalliance/caliptra-sw/blob/main/auth-manifest/README.md) and the SOC images by following steps described in [Manifest-Based Authorization](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/runtime/README.md#manifest-based-image-authorization-new-in-12) section of the Caliptra Runtime Specification.
+3. If the SOC images are authorized and verified, MCU RT releases the reset of the target components.
+
+## Architecture
+
+The following diagram presents the software stack architecture where the Image Loading module resides.
+
+<p align="left">
+    <img src="images/image_loading_sw_stack.svg" alt="sw_stack" width="30%">
```
swenson, line 33: This is better. A few tips:

* Make it larger so that you can make all the text horizontal.
* Try to match the color scheme in the overview doc: gray for apps, oranges for handlers and libraries, soft yellow for SDKs, soft blue for the kernel 

I'll make a note to myself to try to formalize some of this in a style guide.

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+
+The MCU SDK exposes APIs to authorize the manifest and verify images through the "Attestation", "Anti-Rollback Protection" and "Crypto" modules.
+
+### Application Interfaces
+
+The APIs are presented as methods of the ImageLoader struct.
+
+```
+/// Image loader module
+pub struct ImageLoader;
```
swenson, line 48: Let's make this a `trait` so that we can test it more easily.

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+
+The Image Loading module provides the interface to retrieve and parse the manifest from the flash storage, and transfer SOC images from the storage to the target destination.
+
+The MCU SDK exposes APIs to authorize the manifest and verify images through the "Attestation", "Anti-Rollback Protection" and "Crypto" modules.
+
+### Application Interfaces
+
+The APIs are presented as methods of the ImageLoader struct.
+
+```
```
swenson, line 46:
```suggestion
```rust
```

docs/src/image_loading.md
swenson: Yes, they should all be `async` functions.

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+    /// ```rust
+    /// fn(status: DmaStatus);
+    /// ```
+    /// Where `DmaStatus` indicates the completion or failure status of the DMA transfer.
+    pub fn load_image_with_dma(
+        &self,
+        source_address: usize,
+        destination_address: usize,
+        size: usize,
+        callback: fn(status: DmaStatus),
```
swenson, line 114: If this is async, then we don't need a callback, it can be the return value.

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+    ///
+    /// # Callback
+    /// The callback is a function pointer or closure of type:
+    /// ```rust
+    /// fn(status: DmaStatus);
+    /// ```
+    /// Where `DmaStatus` indicates the completion or failure status of the DMA transfer.
+    pub fn load_image_with_dma(
+        &self,
+        source_address: usize,
```
swenson, line 111: Addresses should be `u64` values since they will be on the AXI bus. (`usize` is generally 32 bits for us.)

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+        destination_address: usize,
+        size: usize,
+        callback: fn(status: DmaStatus),
+    );
+```
+
+## Using Image Loading in the Application
+
+This section describes how to use the image_loader module to load an image from flash into its designated destination memory. The application interacts with the image_loader and the security modules to verify and authorize the image before loading it.
```
swenson, line 120: Try to be consistent with how we refer to the module. Maybe "Image Loading module" here.

If we are referring to a Rust package, we can say `image_loader` (in monospaced text), or a `trait` can also be monospaced, like `ImageLoader`.

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Workflow
+
+1. Retrieve the SOC manifest from flash using the image_loader module.
+2. Pass the manifest to the security module by calling its set_manifest API.
```
swenson, line 125: Function calls should be monospaced, or refer to as the "Set Manifest API".

```suggestion
2. Pass the manifest to the security module by calling its `set_manifest` API.
```

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
...
+3. Authorize the image by calling the authorize_and_stash API of the security module, providing the digest of the image.
+4. If authorization succeeds, load the image to its designated memory address.
+
+```rust
```
swenson, line 129: I don't think we need this detailed of code in the specs -- the spec should be about what procedures we follow and what APIs we are going to support, but not the actual code (which will become rapidly out of date).

Save the code for the code unless it is a small example :)

docs/src/image_loading.md
```diff
@@ -0,0 +1,203 @@
+# Image Loading
+
+## Overview
+
+The Image Loading module is a component of the MCU Runtime SDK designed for managing SOC images stored in flash memory. This module provides APIs for:
+
+- Detecting SOC images availability in the [flash storage](./image_loading.md).
```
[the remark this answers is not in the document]
mlvisaya: good catch. yes should be flash_layout.md

docs/src/image_loading.md
mlvisaya: alright, i'll add more detailed steps.

docs/src/image_loading.md
mlvisaya: I can probably just have a one line of code for each step, just to show how it's used.

docs/src/image_loading.md
mlvisaya: Updated the image. You can find the raw image in page-12 of this Visio file: https://microsoft-my.sharepoint.com/:u:/p/marcovisaya/EcaNju7LbulCjYiP533qQzIB2ZruXr3RtT_XBiOwlOCA9Q?e=DAdazr

docs/src/image_loading.md
mlvisaya: Updated the image. You can find the raw image in page-10 of this Visio file: https://microsoft-my.sharepoint.com/:u:/p/marcovisaya/EcaNju7LbulCjYiP533qQzIB2ZruXr3RtT_XBiOwlOCA9Q?e=DAdazr

docs/src/image_loading.md
```diff
@@ -0,0 +1,143 @@
...
+- Loading SOC images to target components. The SOC images could come from a [Flash Storage](./flash_layout.md) or from another platform (e.g. BMC).
+- Retrieving the SOC Image Metadata (as defined in the [SOC Manifest](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/auth-manifest/README.md) documentation)
+- Verifying and authenticating the SOC Images.
+
+The diagram below provides an **example** of how the Caliptra subsystem, integrated with custom SOC elements (highlighted in green), facilitates the loading of SOC images to vendor components.
+
+Custom SOC elements:
+
+* **External Flash** : A flash storage containing SOC manifest and the SOC images.
+* **BMC** : An external Board Mangement Controller that provides a service to source the SOC images and manifest.
```
helloxiling, line 16: I suggest not putting `BMC` under `Custom SOC elements` session.  BMC just interacts with Caliptra subsystem of SoC in streaming boot process, but not part of SOC.

docs/src/image_loading.md
```diff
@@ -0,0 +1,143 @@
...
+
+The following outlines the steps carried out by the MCU RT during the SOC boot process:
+
+1. MCU ROM reads a SOC Configuration register (implementation specific) to determine the source of the images to load (Flash/BMC).
+2. Caliptra RT authorizes and loads Caliptra RT (refer to [Caliptra Subsystem boot flow](https://github.com/chipsalliance/Caliptra/blob/main/doc/Caliptra.md#subsystem-boot-flow) for the detailed steps).
+3. Caliptra switches to Caliptra RT FW.
+4. Caliptra RT indicates to Recovery I/F that it is ready for the SOC manifest image (refer to [Caliptra Subsystem Recovery Sequence](https://github.com/chipsalliance/Caliptra/blob/main/doc/Caliptra.md#caliptra-subsystem-recovery-sequence) for the detailed steps).
+5. Retrieve SOC Manifest
+
+   1. If image is coming from BMC, BMC FW transfers manifest to Recovery I/F
```
helloxiling, line 50: Minor:  It might be better to rephrase as "If image is coming from BMC, SoC manifest is retrieved from Recovery I/F",  since we don't need to define BMC FW behavior in our spec.

docs/src/image_loading.md
```diff
@@ -0,0 +1,143 @@
...
+   2. If Image is coming from Flash, MCU ROM transfers manifest from flash to Recovery I/F
+6. Caliptra RT transfers image to Caliptra Mailbox (MB) SRAM
```
helloxiling, line 52: Does `image` refer to SoC manifest? If so, we can say it explicitly.

docs/src/image_loading.md
```diff
@@ -0,0 +1,143 @@
...
+    /// - `Err(DynError)`: An error if retrieval fails.
+    async fn get_imc(&self) -> Result<ImageMetadataCollection, DynError>;
+
+    /// Loads the specified image to a storage mapped to the AXI bus memory map.
+    /// If the image will be loaded directly to the target component, the AXI mapped load address in the image metadata can be used.
+    ///
+    /// # Returns
+    /// - `Ok()`: Image has been loaded and authorized succesfully.
+    /// - `Err(DynError)`: Indication of the failure to load or authorize the image.
+    async fn load_and_authorize(&self, image_id: u32, address: u64) -> Result<(), DynError>;
```
helloxiling, line 121: Is there a reason that we define `load_and_authorize` as one function instead splitting into two?

docs/src/image_loading.md
```diff
@@ -0,0 +1,143 @@
...
+13. Once digest is verified, the EXEC/GO bit is set
+14. The EXEC/GO bit sets a Caliptra wire to MCU (as a consequence of setting the EXEC/GO bit in the previous step)
+15. MCU ROM detects Caliptra'go' wire to the MCU got set.
+16. MCU ROM passes a parameter using the FW HandOff table to indicate the image source (i.e. the image source where it booted from)
+17. MCU switches to MCU RT
+18. MCU RT retrieves the image source from HandOff table
+19. BMC or a similar platform component will now do MCTP enumeration flow to MCU over I3C.
+20. Retrieve SOC Images Metadata
+
+    1. If Image is coming from BMC, retrieve SOC Manifest and Image Metadata Collection through PLDM T5 flow
```
helloxiling, line 71: SoC manifest has been retrieved in step(5).  We should start from image metadata collection via PLDM-T5 flow.

docs/src/image_loading.md
```diff
@@ -0,0 +1,143 @@
+# Image Loading
+
+## Overview
+
+The Image Loading module is a component of the MCU Runtime SDK designed for managing SOC images. This module provides APIs for:
+
+- Loading SOC images to target components. The SOC images could come from a [Flash Storage](./flash_layout.md) or from another platform (e.g. BMC).
+- Retrieving the SOC Image Metadata (as defined in the [SOC Manifest](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/auth-manifest/README.md) documentation)
+- Verifying and authenticating the SOC Images.
```
helloxiling, line 9: Image loading module won't authenticate SOC image directly. It interacts with Caliptra RT via mailbox interface to achieve the function. 
Can we clarify this?

docs/src/image_loading.md
swenson: I think that is how we typically do it. We generally do both steps as once to reduce back and forth, since we aren't really going to want to do one without the other, right?

docs/src/image_loading.md
mlvisaya: That is true, it's not part of the SOC. I'll remove it.

docs/src/image_loading.md
mlvisaya: yes, it should be SoC manifest

docs/src/image_loading.md
mlvisaya: On one hand, splitting them to 2 might offer some flexibility to the user, but I can't think of a use-case where the user just loads it and not authorize it immediately. I think the advantage of 1 function is just it's simpler to use, and as Chris said reduces back and forth with the application. Also i'd like to prevent them from calling authorize() first before calling load(), so making it atomic just ensures these 2 steps are followed in sequence.

docs/src/image_loading.md
mlvisaya: Actually the Image Metadata Collection (IMC) is inside the SOC Manifest (https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/auth-manifest/README.md), and the PLDM package will send the SOC Manifest (with the IMC included). However, for the flash use case, since we know the structure of the SOC Manifest, we can get the IMC directly without having to load the preamble part. I can probably reword this to retrieve the Image Metadata Collection (as part of the SOC Manifest) through PLDM-T5 flow.

docs/src/image_loading.md
mlvisaya: Yeah, I can probably reword this to mean that it provides the interface for authentication through Caliptra core but not itself doing the authentication.

docs/src/image_loading.md
mlvisaya: I think you're right about not mentioning BMC FW. Although the for the Flash use case, the SoC manifest is also retrieved from Recovery I/F.

Should I just say BMC Recovery Agent and follow the wording in step 11 of https://github.com/chipsalliance/Caliptra/blob/main/doc/Caliptra.md#subsystem-boot-flow ?

docs/src/images/image_loading_sample.svg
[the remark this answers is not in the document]
mlvisaya: Ack. How about "PLDM FW Update Agent" instead of BMC?

docs/src/image_loading.md
```diff
@@ -0,0 +1,143 @@
...
+* *Blue Arrows indicates actions taken by MCU RT*
+* *Black Arrows indicates actions taken by BMC FW*
+
+<p align="center">
+    <!--- https://www.plantuml.com/plantuml/uml/bLLlR_es47xlJt6_JffQgONrJpU95rDfrZ-LXfY0TPF6L9da0fQ2ZPnBM_hfvnmuYHlA-BsAxNly-EwvnyvDJj9IiSp4DrhW4Y5JymNDhLoBbGkeMAsa9ZWQ3oE8Z4xLlB2IbD578FFjIeYCPAPMP2K3_EGJsMCV-d7C23VUlHy6KcKpyrwY_3X43g8_EUY-hG2CH_s1ZxwKan0omjYyebt3K_UUONuXXDqE8ewwRnYupsI-u12Fmk0pwiHOfyvoPJHgObpiPq9eGmZbQPK-FHWLMYizhrPWD1o8BjHGLqUDSfCMe3gW1tVl61TbbzoYHN1VYpcIa0KPNIndQCMcTMV__60n-VVNEe-dfPmZZ4rXO_V136EQwcCkikirEX6xQZXbK0_w1IsCLHy8KYVmkv59ewW9tFzrOFqC5GA4QEGDEPcLMyY8Uz88N7JEE-V2R_U5-PJT0IYJDQJ6GkdbWTGgnPo4OpchKM7CBtkohGj3L2ivp7BSk29CpoKeUFaxwNGI270fMaNw-KPxrAHYIHZG7gRXrS4QiW4d6xTyqZ28jJJqrp647IZYLI3YuNnj5Q-5y3izatk9Ul3egC6_7sAFr7oxfGrLXgbcWjSSjcyS4-vX6IF1tTztKVTX2BUAMaY-zGy6tfH5o97Ah8FORqWO4oQm8G2Lxu3nEQoClzEWj4jD9yCq1PAp3CM-bsOYMeGZ9Alm5V-7jRPPm7_gLigEJVu0T1t1pJ-bsNNVv5bMOBFB3rKX0oIPI98GcSmTpWJSPY4oOrPiKoojnorIR0kVpBULbVEb14Bp0MIBbeVvY6wKpci3iq4QWH-X3jLlRmFmPJbDkBtiDBPKWGhAXrxyr0HnbKuHZwaV4dlL0UcVtsCk-KKb9tlyPwyxy8nMfUkDzPMkd3lLRueMWRyPCzKihMVVH_vpnvkxwKqFs97_BvtIM8NLa1TnZ7akjh7Gp3vMQx3pIXa-rQHGoJhLiR4MSpTBIdUJ-RgEQbDu0L4dKFfBtBXHiSp-0m00 -->
```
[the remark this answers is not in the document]
mlvisaya: gotcha. Yeah I can remove SOC Configuration and just mention it. Yes technically, Image Metadata Collection(IMC) is inside the SOC Manifest which PLDM transfers as 1 component. SOC Manifest is basically composed of a preamble+IMC, but for flash, we can skip through the preamble part and just retrieve directly the Image Metadata Collection. I think to avoid confusion, I can just say Get Image Data Collection for both and just mention in the step that it is part of the SOC Manifest.

docs/src/image_loading.md
helloxiling: I was thinking from the perspective of interface definition, usually one function just does one specific job.  
I agree that reducing back-and-forth interaction with app in our scenario is a valuable point.  Let's keep it as it is.

docs/src/image_loading.md
helloxiling: yeah,  it's good to reword as `BMC's Recovery Agent`.

swenson (approved):
LGTM

helloxiling (approved):

mlvisaya:
Thanks!!

PR #52, opened 2024-12-07
Vendor libtock-rs

swenson:
We vendor it for a few reasons:

* It simplifies our build, so that we no longer have to specify the LIBTOCK environment variables
* We can make more easily make changes to accommodate our user-mod testing
* We can integrate async support more seamlessly.

I also took this opportunity to rename `pldm-app` to `example-app`.

I ran `cargo clippy --fix` and did some other clippy fixes. (This also reordered some of the tests in other files.)


helloxiling (approved):
Our code change looks good to me, though I haven't fully reviewed libtock-rs code.

swenson:
Thanks! Don't worry about all of the libtock-rs code for now, just because most of it we aren't going to need, and we don't want to change it too much (in case we want to cherry pick code from the upstram later).

PR #53, opened 2024-12-09
Set and read I3C addresses in the right fields

swenson:
Also initialized the timing registers (recently added to the I3C core docs).


parvathib (approved):
Thanks for addressing this promptly :)

swenson:
Thanks!

PR #55, opened 2024-12-10
Add OTP RDL and registers

swenson:
For the OpenTitan fuse (one-time programmable) controller, and also for the related entropy sources.

These RDLs is added in the `caliptra-ss` repo at commit: https://github.com/chipsalliance/caliptra-ss/commit/984f22868c978106e8d27928a3eb82a348119396

I also slightly changed the register names to also include their most recent sub-block name so that the digest register names would be less confusing. This caused some changes to the I3C register names (which I've fixed the corresponding code for).


helloxiling (approved):

swenson:
Thanks!

PR #57, opened 2024-12-12
User mode testing example

swenson:
Reworked the example Tock user mode app to support running both in RISC-V in the emulator as well as using the `libtock_unittest`'s fake kernel. I also moved the executor and `TockSubscribe` `impl Future` into a common crate, `libtockasync`, to make it easier to use from other libraries and apps.

This allows us test directly from the host without the emulator:

```shell
cargo t -p example-app
```

And to run the app directly:
```shell
cargo run --bin example-app
```

This should make it much easier to develop and test apps with a rapid development cycle.

I added some simple tests to demonstrate how to test the async code. It's a little tricky since the fake kernel and the Embassy executor need to be in the same lexical scope as the test so that they don't get dropped and cause a segfault.

A future TODO is to have integration tests using the real Tock kernel using a similar framework to `tests/integration` for our Tock code.


mlvisaya (approved):
lgtm

parvathib (approved):
LGTMToo!

swenson:
Thanks!

PR #58, opened 2024-12-13
Use a single block of RAM

swenson:
Rather than having split instruction and data RAM (like Caliptra Core has).

This matches what our eventual hardware will eventually be.

The main challenge is that we don't necessarily know how much data RAM we're going to need yet, nor do we know the final amount of total RAM we're going to have.

Here, we tweak the build so that the RAM offsets are computed more dynamically.

For now, we allocate the last 64 KB of RAM as data RAM, but we still try to be flexible about its location.

We also move some of the constants specifying it into one place, and turned optimization to the space-saving setting (`opt-level` set to `s`).

The `example` app seems to have a little trouble: I had to adjust the stack size so that it would load after this change, which makes me suspect we have a bug somewhere, either in the app RAM offset calculations, or in the way we run integration tests.

Next I'll work on enabling PMP, which I think will help us identify these bugs more easily.


tests/hello/link.ld
```diff
@@ -20,7 +20,7 @@ MEMORY
 {
   ROM   (rx) : ORIGIN = 0x00000000, LENGTH = 32K
   ICCM (rwx) : ORIGIN = 0x40000000, LENGTH = 128K
-  DCCM (rwx) : ORIGIN = 0x50000000, LENGTH = 128K
+  DCCM (rwx) : ORIGIN = 0x40020000, LENGTH = 128K
```
helloxiling, line 23: Do we need to update `DCCM`, `ICCM` to align to the change for using single block of RAM?

swenson: Sure. This is a small example, but it's easy enough to switch to a single block.

helloxiling (approved):
Looks good to me.

swenson:
Thanks!

PR #62, opened 2024-12-17
runtime: Enable PMP

swenson:
I wasn't able to use an existing PMP layout due to our potential for multiple MMIO blocks as well as our unified RAM block: the existing PMP layouts that Tock has built in expect things to be aligned to neater boundaries with power of two sizes.

So, I modified the standard kernel EPMP layout to support what we need.

The "E" is used to indicate we support the SMEPMP extension, which provides some additional mode bits that allow more fine-grained control of shared areas between machine and user mode and allowing us to set a default deny all policy.

I moved around our memory map a little to make the PMP setup a little easier and reduce the number of regions we need, but left it flexible enough that we can still accommodate whatever memory map changes we need to make in the future. I moved the emulator-only UART and control peripherals to be in the `0x1000_0000` range, which will be accessible from user mode, while the `0x2000_0000` peripherals will only be accessible by the kernel.

This also fixes a bug in the CPU emulator's PMP calculation when using natural power of 2 regions (NAPOT). I've also ported that fix over to Caliptra core.

### DIFF
--- a/emulator/cpu/src/cpu.rs
+++ b/emulator/cpu/src/cpu.rs
@@ -35,11 +35,11 @@ pub struct CoverageBitmaps<'a> {
     pub iccm: &'a bit_vec::BitVec,
 }
 
-const ICCM_SIZE: usize = 128 * 1024;
-const ICCM_ORG: usize = 0x40000000;
+const ICCM_SIZE: usize = RAM_SIZE as usize;
+const ICCM_ORG: usize = RAM_OFFSET as usize;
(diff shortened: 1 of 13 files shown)


parvathib (commented):
LGTM! Thanks for answering my questions :)

parvathib (approved):

helloxiling (approved):
PR looks pretty good. Are we going to add any documentation in this area?

swenson:
Thanks!

Documentation is a good call. I'll add some docs about this in a separate PR.

PR #63, opened 2024-12-17
Add loop back tests for MCTP stack

parvathib:
The PR adds the following

1.  Adds loopback tests for various message sizes: send a test message to the emulator test thread, receive the same message from the thread in loopback, and compare the results. This aims to test the message assembly and packetization logic. 
2.  Refactors kernel tests for improved structure and readability. 
3. Adds couple fixes in emulator for I3C peripheral and the test threads.


swenson (approved):
LGTM

PR #65, opened 2024-12-18
Add HIL interfaces for DMA and Caliptra Mailbox

mlvisaya:
Define HIL interfaces for Caliptra Mailbox and DMA.

This is based on the mailbox functionality defined in :

https://github.com/chipsalliance/caliptra-rtl/blob/main/docs/CaliptraIntegrationSpecification.md#mailbox

and the DMA operation in:

https://github.com/chipsalliance/caliptra-ss/blob/main/docs/Caliptra%202.0%20Subsystem%20Specification%201.pdf

### DIFF
--- a/runtime/dma/Cargo.toml
+++ b/runtime/dma/Cargo.toml
@@ -1,0 +1,16 @@
+# Licensed under the Apache-2.0 license.
+
+[package]
+name = "dma-driver"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+
+[dependencies]
+kernel = { git = "https://github.com/tock/tock.git", rev = "b128ae817b86706c8c4e39d27fae5c54b98659f1" }
+
+[target.'cfg(target_arch = "riscv32")'.dependencies]
+capsules-core = { git = "https://github.com/tock/tock.git", rev = "b128ae817b86706c8c4e39d27fae5c54b98659f1" }
+registers-generated.workspace = true
+romtime.workspace = true
+tock-registers.workspace = true
--- a/runtime/dma/Cargo.toml
+++ b/runtime/dma/Cargo.toml
@@ -9,8 +9,3 @@ edition.workspace = true
 [dependencies]
 kernel = { git = "https://github.com/tock/tock.git", rev = "b128ae817b86706c8c4e39d27fae5c54b98659f1" }
 
-[target.'cfg(target_arch = "riscv32")'.dependencies]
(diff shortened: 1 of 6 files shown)


runtime/mailbox/src/hil.rs
```diff
@@ -0,0 +1,117 @@
...
+    ///
+    /// Returns:
+    /// - `Ok(response_length)` where `response_length` is the number of bytes read.
+    /// - `Err(ErrorCode)` if there was an issue reading the response.
+    fn read_response(&self, buffer: &mut [u8]) -> Result<usize, ErrorCode>;
+
+    /// Check the status of the mailbox after a command execution.
+    ///
+    /// Returns:
+    /// - `Ok(status_code)` where `status_code` represents the current mailbox status.
```
parvathib, line 51: nit: Should we write `MailboxStatus` instead of `statis_code` for clarity?

parvathib (approved):

runtime/mailbox/src/hil.rs
mlvisaya: that's a good idea. i'll change it to MailboxStatus.

runtime/dma/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition.workspace = true
+
+[dependencies]
+kernel = { git = "https://github.com/tock/tock.git", rev = "b128ae817b86706c8c4e39d27fae5c54b98659f1" }
+
+[target.'cfg(target_arch = "riscv32")'.dependencies]
+capsules-core = { git = "https://github.com/tock/tock.git", rev = "b128ae817b86706c8c4e39d27fae5c54b98659f1" }
+registers-generated.workspace = true
+romtime.workspace = true
+tock-registers.workspace = true
```
swenson, line 16: I don't think these are needed yet. (Same for mailbox.)
```suggestion
```

runtime/dma/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+// Licensed under the Apache-2.0 license.
+
+#![cfg_attr(target_arch = "riscv32", no_std)]
+
+#[cfg(target_arch = "riscv32")]
```
swenson, line 5: There's nothing RISC-V-specific about the HIL. It makes the build happier if we include it for all targets. (Same for mailbox.)

```suggestion
```

swenson (approved):
LGTM, but you can remove the RISC-V config stuff for the HILs at least.

PR #66, opened 2024-12-18
Run one integration test at a time

swenson:
We currently have files and ports that the MCU emulator is using that can conflict if we are running two emulators at once, so we force the tests to only run one at a time.


helloxiling (approved):

swenson:
Thanks!

PR #67, opened 2024-12-20
Usermode flash API (sync IO) with testing in example app

helloxiling:


helloxiling:
This PR is no longer needed for review. Will open another one for flash async API.

PR #69, opened 2024-12-24
Use SoC interface to talk to Caliptra

swenson:
`caliptra-sw`'s emulator has an "external" SoC interface that can be used to communicate with it via the standard SoC interface.

This writes a simple adapter layer so that the MCU emulator can mount that external interface as a peripheral on our root bus.

I also added some code to the ROM that reads the flow status from the SoC interface as a basic check that the functionality is working. When running the MCU with a Caliptra ROM and runtime FW, it outputs:

```
Caliptra flow status 40000000
```

When no Caliptra ROM or firmware are provided, the status is 0, so we can skip straight to Tock in that case (for now).


mlvisaya (approved):
lgtm

swenson:
Thanks!

PR #75, opened 2025-01-02
User space interface for DMA and Mailbox

mlvisaya:


runtime/apps/apis/dma/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+//! # AsyncDMA: A DMA Interface for AXI Source to AXI Destination Transfers
+//!
+//! This library provides an abstraction for performing asynchronous Direct Memory Access (DMA)
+//! transfers between AXI source and AXI destination addresses.
+
+use libtock_platform::{ErrorCode, Syscalls};
+use libtockasync::TockSubscribe;
+
+/// DMA interface.
+pub struct AsyncDMA<const DRIVER_NUM: u32, S: Syscalls>(S);
```
swenson, line 12: I think we can make this `DRIVER_NUM` just a constant in the library. I don't think we need it to be a generic parameter.

runtime/apps/apis/mailbox/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+///
+/// # Generics
+/// - `DRIVER_NUM`: The driver number for the mailbox interface.
+/// - `S`: The syscall implementation.
+/// - `W`: Configuration for writable buffers (response).
+/// - `R`: Configuration for read-only buffers (input), defaults to `DefaultConfig`.
+pub struct AsyncMailbox<
+    const DRIVER_NUM: u32,
+    S: Syscalls,
+    W: ReadWriteConfig,
```
swenson, line 19: If possible, I think we also want to eliminate these configs from the generics. I don't see us changing the values, and it makes things harder to read and write.

helloxiling (commented):
@mlvisaya, please take a look at latest mainline change: `ae37e04` add mctp userspace lib support (#76). We prefer a single package `libtock_caliptra` for userland APIs.  Please adjust for consistency.

mlvisaya:
@helloxiling , thanks I updated this PR now to be consistent with Flash and MCTP implementations

runtime/apps/syscall/src/dma.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub source: DMASource<'a>,
+    /// Destination address for the transfer.
+    pub dest_addr: u64,
+}
+
+/// Represents the source of data for a DMA transfer.
+#[derive(Debug, Clone)]
+pub enum DMASource<'a> {
+    /// A memory address as the source.
+    Address(u64),
```
swenson, line 32: Is this an AXI address or a local memory address (which would be a `u32`)?

It might be good to have a separate type for AXI addresses so that we don't mix them up. Something like `pub type AxiAddr(u64);` or `pub type AxiAddr = u64`.

mlvisaya: it is Axi address. good point, I can create a new type for AxiAddr

swenson (approved):
LGTM

helloxiling (approved):

PR #76, opened 2025-01-03
add mctp userspace lib support

parvathib:
1. Adds the MCTP user library API and includes a function in the example_app to test dummy SPDM messages in loopback mode.
2. Define a separate instance to the secure SPDM.
3. Add Alarm in the Mctp Mux layer to record the message reception time. (This is not propagated to the user yet)
4. Pass the test flags to application build to filter out the code execution.


runtime/apps/apis/src/mctp.rs
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+}
+
+pub mod message_type {
+    pub const SPDM: u8 = 0x5;
+    pub const SECURE_SPDM: u8 = 0x6;
+    pub const PLDM: u8 = 0x1;
+    pub const CALIPTRA: u8 = 0x7E;
+}
```
vsonims, line 31: Is this needed?

runtime/src/components/mod.rs
```diff
@@ -4,5 +4,5 @@
 
 pub mod flash_partition;
 pub mod mctp_driver;
-pub mod mctp_mux;
 pub mod mock_mctp;
```
vsonims, line 7: shouldn't this be under a feature flag to prevent inclusion in release?

runtime/capsules/src/mctp/driver.rs
```diff
@@ -339,61 +338,88 @@ impl<'a> MCTPTxClient for MCTPDriver<'a> {
...
-                                upcall::MESSAGE_TRANSMITTED,
-                                (
-                                    kernel::errorcode::into_statuscode(result),
-                                    dest_eid as usize,
-                                    msg_info,
-                                ),
-                            )
-                            .ok();
+
+        if self.msg_type as u8 == msg_type {
```
swenson, line 342: It might be good to remove a couple of layers of nested `if`s to make this a little easier to read.

e.g.,

```rust
if self.msg_type as u8 != msg_type {
    // panic here
}
if self.current_app.get().is_none() {
    return;
}
let process_id = self.current_app.get().unwrap();
....
```

runtime/capsules/src/mctp/driver.rs
```diff
@@ -339,61 +338,88 @@ impl<'a> MCTPTxClient for MCTPDriver<'a> {
...
+    fn receive(
+        &self,
+        src_eid: u8,
+        msg_type: u8,
+        msg_tag: u8,
+        msg_payload: &[u8],
+        msg_len: usize,
+        recv_time: u32,
+    ) {
+        if self.msg_type as u8 == msg_type {
```
swenson, line 383: Ditto here about deeply nested stuff. For this one, moving some of it out into a separate function might also help.

runtime/apps/apis/src/mctp.rs
parvathib: nope. removed

runtime/src/components/mod.rs
parvathib: The test code that creates this mock instance is compiled out in the release builds. The release binary should not have this included.

runtime/capsules/src/mctp/driver.rs
parvathib: Sure! I refactored the code to simplify. Can you check if it looks okay?

runtime/capsules/src/mctp/driver.rs
parvathib: Refactored as suggested. Can you check if it looks okay?

runtime/src/components/mod.rs
vsonims: I think the module inclusion also needs to be under feature flag

swenson (approved):
LGTM other than Vishal's comment that it would be good to put the `mod mock_mctp;` under a `#[cfg(..)]` to double-ensure that it is not included in the released binary.

PR #78, opened 2025-01-03
Update Image Loading Doc

mlvisaya:
- Update the API, user will choose which image to load by specifying the image id.
- MCU FW will not need to get the metadata of the image from flash or PLDM, instead it will issue a Mailbox command to get the load address
- Remove the example usage, since there's only API to use.


helloxiling (approved):
Looks good to me.

swenson (approved):
LGTM

PR #79, opened 2025-01-03
xtask: Enforce most dependencies are specified in the workspace

swenson:
We add an `xtask` that checks that all dependencies are specified in the root `Cargo.toml`. This will help keep things clean as we go forward.

The only exception is for architecture-specific dependencies, which might have arch-specific feature flags.

This cleans up our existing `Cargo.toml` files to be in line with this, and removes some unnecessary dependencies.


vsonims (approved):

swenson:
Thanks!

PR #80, opened 2025-01-07
rom: Populate fuses from OTP controller

swenson:
We had to shave a few yaks.

First, we added an autogenerator that takes the OTP memory map hjson file and uses it to generate a list of fuses (partitions, offsets, sizes) and some simple data structures for manipulating them, since doing so is fiddly and error-prone.

Plus, this memory map is in the process of changing quite a lot, so this helps us avoid having to reprogram everything manually later.

Then, we updated the emulator to use the autogenerated peripheral trait and the autogenerated fuse data, which will also help keep it up-to-date.

Finally, we added in a section of ROM code to read the non-secret fuses from the OTP controller and write them to Calipta Core's SoC interface.

(There is a hardware state machine that will write the secret fuses, so we don't have to worry about.)

For now, I've checked in a copy of the latest `otp_ctrl_mmap.hjson` file from `caliptra-ss` because that repository does not have a single branch that contains both it and the I3C RDL that we need at the moment.


rom/src/error.rs
```diff
@@ -0,0 +1,9 @@
+// Licensed under the Apache-2.0 license
+
+#[allow(dead_code)]
+#[repr(u32)]
+#[derive(Debug)]
+pub enum MCUError {
```
helloxiling, line 6: minor: should we prefer `McuError` for consistent naming convention?

rom/src/riscv.rs
```diff
@@ -12,41 +12,140 @@ Abstract:
...
+        I3c { registers }
+    }
+}
+
+fn recovery_flow() {
+    // TODO: implement Caliptra boot flow
+    let i3c = I3c::new(I3C_BASE);
+
+    // TODO: read this value from the fuses (according to the spec)?
+    i3c.registers.sec_fw_recovery_if_device_id_0.set(0x3a);
```
helloxiling, line 145: Is `0x3a` a magic number to indicate recovery flow?

rom/src/riscv.rs
```diff
@@ -12,41 +12,140 @@ Abstract:
...
     pub const fn new(registers: StaticRef<soc::regs::Soc>) -> Self {
         Soc { registers }
     }
+
+    fn flow_status(&self) -> u32 {
+        self.registers.cptra_flow_status.get()
+    }
+
+    fn ready_for_fuses(&self) -> bool {
+        self.flow_status() & 0x40000000 != 0
```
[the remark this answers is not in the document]
swenson: I think there is a mask defined already that I should have used. :) Will do.

rom/src/riscv.rs
swenson: This is a placeholder dynamic address. I'll add a comment about it.

helloxiling (approved):
Looks good to me.

swenson:
Thanks!

PR #86, opened 2025-01-10
Add dual flash support 

helloxiling:
…y flash


helloxiling:
This PR introduces the changes to add another identical recovery flash controller in the emulator to enable dual flash support. The flash test code were also refactored to cover the testing on recovery flash.

emulator/app/Cargo.toml
```diff
@@ -30,16 +30,18 @@ strum_macros.workspace = true
 strum.workspace = true
 tock-registers.workspace = true
 zerocopy.workspace = true
+tempfile.workspace = true
```
mlvisaya, line 33: should this be under dev-dependencies?

runtime/apps/syscall/src/flash.rs
```diff
@@ -229,8 +229,9 @@ impl<S: Syscalls> SpiFlash<S> {
 // -----------------------------------------------------------------------------
 
 pub mod driver_num {
-    pub const IMAGE_PARTITION: u32 = 0x8000_0006;
+    pub const ACTIVE_IMAGE_PARTITION: u32 = 0x8000_0006;
     pub const STAGING_PARTITION: u32 = 0x8000_0007;
+    pub const RECOVERY_IMAGE_PARTITION: u32 = 0x8000_0008;
```
mlvisaya, line 234: this might conflict with DMA driver num defined with the same value 0x8000_0008, in syscall/dma.rs

runtime/capsules/src/flash_partition.rs
```diff
@@ -11,8 +11,9 @@ use kernel::{ErrorCode, ProcessId};
 
 /// Each partition is presented to userspace as a separate driver number.
 /// Below is the temporary driver number for each partition.
-pub const IMAGE_PAR_DRIVER_NUM: usize = 0x8000_0006;
+pub const ACTIVE_IMAGE_PAR_DRIVER_NUM: usize = 0x8000_0006;
 pub const STAGING_PAR_DRIVER_NUM: usize = 0x8000_0007;
+pub const RECOVERY_IMAGE_PAR_DRIVER_NUM: usize = 0x8000_0008;
```
mlvisaya, line 16: please check with if it conflicts with DMA driver num in syscall/dma.rs

mlvisaya (commented):
looks good, just some minor questions

emulator/app/Cargo.toml
swenson: I think it needs to be under the normal dependencies since anything in `if cfg!()` still needs to be able to compile regardless of the feature flags.

emulator/app/src/main.rs
```diff
@@ -405,15 +405,46 @@ fn run(cli: Emulator, capture_uart_output: bool) -> io::Result<Vec<u8>> {
...
-        &clock.clone(),
-        Some(PathBuf::from("primary_flash")), // TODO: make this configurable
-        flash_ctrl_error_irq,
-        flash_ctrl_event_irq,
-    )
-    .unwrap();
+    let create_flash_controller = |default_path: &str, error_irq: u8, event_irq: u8| {
+        // Use a temporary file for flash storage if we're running a test
+        let flash_file = if cfg!(any(
+            feature = "test-flash-ctrl-init",
```
swenson, line 411: I wonder if we should always set `test` when running tests so that we can just use `cfg(test)` instead of checking every feature flag.

emulator/periph/src/flash_ctrl.rs
```diff
@@ -333,7 +350,180 @@ impl DummyFlashCtrl {
...
+    ) -> emulator_bus::ReadWriteRegister<
+        u32,
+        registers_generated::main_flash_ctrl::bits::FlInterruptState::Register,
+    > {
+        emulator_bus::ReadWriteRegister::new(self.interrupt_state.reg.get())
+    }
+
+    fn write_fl_interrupt_state(
+        &mut self,
+        _size: emulator_types::RvSize,
```
swenson, line 376: nit: It might be good to import some of these types to make the functions a little easier to read.

swenson (approved):
LGTM, minor nits only

runtime/apps/syscall/src/flash.rs
helloxiling: @mlvisaya, DMA driver num is defined as `0x8000_0008`  and Mailbox driver num is  defined as `0x8000_0009`.  Can we adjust them to something like `0x8000_1000`,  `0x8000_2000` to leave more spaces for growth of flash partition drivers?

mlvisaya: ok. I created https://github.com/chipsalliance/caliptra-mcu-sw/issues/96 so to remember to change it.

mlvisaya (approved):

emulator/app/src/main.rs
helloxiling: @swenson, it seems not working by only checking `cfg(test)`. I might have to keep as it is for now.

swenson: @helloxiling I think you would need to pass the `test` feature flag when running the emulator in the integration test.

PR #88, opened 2025-01-10
rom/emulator: Add small DCCM for ROM stack

### LINKED ISSUE #81: Add DCCM for ROM data
The AXI SRAM will not be available when MCU ROM starts, so it needs to reference the small DCCM RAM for its stack.

swenson:
The ROM has a small DCCM (data SRAM) mounted at `0x5000_0000` so that is has room for a stack. This is needed since the normal RAM block may not be ready yet when the ROM starts, since that normal RAM is provided over AXI through MCI.

We grab the configuration for the DCCM from the C `#define` statements that are generated from the RTL, and use it to populate the automatic root bus automatically.

While we are there, we also make those `#define` statements available as Rust constants for firmware. (I have a goal of reducing the number of constants we have to define and keep in sync with RTL by hand.)

Fixes #81


mlvisaya (approved):
i think it's good

swenson:
Thanks!

PR #90, opened 2025-01-14
Disable MCTP user loopback tests really, really hard when disabled

swenson:
Functionally there should be no difference because of this PR, as this function is not even called unless the feature is enabled.

But, before, you can see that when running the runtime that it is blocked keyboard input before the async sleeps will finish, while after this fix, it is not.

This functionality makes no sense. The `cfg!()` macro should be evaluated to `if false`, which should remove the code block entirely from the resulting binary.

Switching it to use this double `cfg` pattern removes 2 KB from the application binary, and removes this confusing keyboard behavior.

I suspect this is due to a compiler bug. It seems to be present even on the latest nightly compilers as well.

I'm going to keep investigating this and if applicable, report it to the Rust compiler devs.

### DIFF
--- a/runtime/apps/example/src/main.rs
+++ b/runtime/apps/example/src/main.rs
@@ -77,7 +77,8 @@ pub(crate) async fn async_main<S: Syscalls>() {
         writeln!(console_writer, "async sleeper woke").unwrap();
     }
 
-    if cfg!(feature = "test-mctp-user-loopback") {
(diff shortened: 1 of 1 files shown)


parvathib (approved):
LGTM

swenson:
Related I think: https://github.com/rust-lang/rust/issues/135468

swenson:
Thanks!

PR #91, opened 2025-01-14
emulator: Save interrupts to be processed later in action loop

swenson:
Otherwise, if the CPU is busy with another interrupt, then additional interrupts are dropped and never sent to the interrupt handler. This is different from how the real CPU works, where the interrupt would keep being asserted until the the interrupt handler re-enables interrupts (when it jumps out of machine mode).

This was causing problems with our test code sometimes as the timer interrupts were getting lost randomly.


helloxiling (approved):
Looks good to me.

swenson:
Thanks!

PR #92, opened 2025-01-14
Image Loading API implementation

mlvisaya:
Image Loading API Implementation

- Implement image loading API interface as described in
https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/docs/src/image_loading.md
- Import checksum API from caliptra-core
- Import mailbox API from caliptra-core only with the commands used for
Image Loading (GetImageOffset, GetImageSize, AuthorizeAndStash)
- Add Fake drivers for DMA, Mailbox and Flash to be used for unit
testing
- Add unit tests for Image Loading and Mailbox
- Fixed bug on syscall driver on DMA and Mailbox. The problem was the
allowed RO or RW buffers shared by the syscall to the driver was not
released after the upcall. It was found that there is need to put the
RO/RW buffers in the generic of share::scope block, otherwise, the
unallow closures associated with RO/RW buffers will not be called after
the call.


runtime/apps/api/src/mailbox.rs
```diff
@@ -0,0 +1,373 @@
...
+            MailboxResponse::GetImageLoadAddress(resp) => resp.verify(),
+            MailboxResponse::GetImageLocationOffset(resp) => resp.verify(),
+            MailboxResponse::GetImageSize(resp) => resp.verify(),
+            MailboxResponse::AuthorizeAndStash(resp) => resp.verify(),
+        }
+    }
+}
+
+/// Trait for mailbox responses.
+pub trait MailboxResponseTrait: FromBytes + IntoBytes + Immutable {
```
helloxiling, line 178: It is a little redundant to add `trait` into a trait name. Can we come up with a better one, maybe something like `MailboxResponder`?

runtime/apps/api/src/mailbox.rs
```diff
@@ -0,0 +1,373 @@
+// Licensed under the Apache-2.0 license
+
+use crate::checksum;
+use core::mem;
+use libsyscall_caliptra::mailbox::Mailbox;
+use libtock_platform::{ErrorCode, Syscalls};
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout, Ref, TryFromBytes};
+
+/// API for interacting with the mailbox.
+pub struct MailboxAPI<S: Syscalls> {
```
helloxiling, line 10: Do we need to add `API` into the struct name?

helloxiling (commented):
@mlvisaya, this PR also touched `flash` syscall file. Is that part of the bug fixes? If yes, can we create a separate PR to address it?

mlvisaya:
> @mlvisaya, this PR also touched `flash` syscall file. Is that part of the bug fixes? If yes, can we create a separate PR to address it?

oh, it's not. I accidentally included it. I think my cargo fmt changed the file., and only formatting is changed. I can remove syscall/flash.rs change.

runtime/apps/api/src/mailbox.rs
mlvisaya: Yeah we can change it. This is supposed to represent the type of MailboxResponse, and includes the common attributes of all mailbox response structures. For example, AuthorizeAndStashResp and GetFirmwareInfoResp implements MailboxResponse. Maybe MailboxResponseType or MailboxResponseAttr would be a better name?

runtime/apps/api/src/mailbox.rs
mlvisaya: We don't need to. I can remove it.

swenson (approved):
LGTM

mlvisaya:
Thanks!

PR #93, opened 2025-01-16
fix: correct total_size and binary_end_offset in TBF header of the app

helloxiling:
The total size calculation in the TBF header was incorrect due to a missing TBF header. This caused failures when loading multiple processes in the kernel, as the kernel relies on the total size in the TBF header to locate the next available app on flash. Without correct setting, kernel looked at the wrong offset on flash when loading the second apps and caused "TBF header not found" error.


swenson (approved):
LGTM -- good catch!

PR #95, opened 2025-01-24
[Doc] PLDM stack spec 

helloxiling:


docs/src/pldm.md
```diff
@@ -0,0 +1,268 @@
...
+
+- **PLDM message control and discovery**: This feature enables the device to be discovered by the platform management controller (typically a BMC) following MCTP enumeration. It forms the foundation for running other PLDM protocols.
+
+- **Streaming boot reminder firmware**: PLDM firmware update protocol defines standardized messages and data structures for obtaining firmware code and data. MCU leverages it to stream boot reminder firmware, which is vendor-specific SoC firmware and beyond. There are several customized amendments to PLDM firmware update specification to enable streaming boot and automatic activation. Details are available in OCP whitepaper [Flashless Boot using OCP, PCIe, and DMTF Standards](https://docs.google.com/document/d/1cjdgcKgOzcug5bBoK6k2Mw2mvsQJElp8bs0ec_xLZHc/edit?usp=sharing).
```
swenson, line 9: minor grammar and spelling fixes:
```suggestion
- **Streaming boot remainder firmware**: PLDM firmware update protocol defines standardized messages and data structures for obtaining firmware code and data. MCU leverages it to stream boot the remainder firmware, which is any vendor-specific SoC or other firmware. There are several customized amendments to the PLDM firmware update specification to enable streaming boot and automatic activation. Details are available in the OCP whitepaper, [Flashless Boot using OCP, PCIe, and DMTF Standards](https://docs.google.com/document/d/1cjdgcKgOzcug5bBoK6k2Mw2mvsQJElp8bs0ec_xLZHc/edit?usp=sharing).
```

docs/src/pldm.md
```diff
@@ -0,0 +1,268 @@
...
+
+![PLDM stack base](images/pldm_stack_base.svg)
+
+- **PLDM service**
+      - Listens for incoming PLDM requests.
+      - Extracts the command op code from the request and invokes the corresponding handler to process.
+      - Sends the response via MCTP transport layer.
+- **Command handler**
+      - Interacts with message decoding library to decode the request.
+      - Executs the command.
```
swenson, line 27:
```suggestion
      - Executes the command.
```

docs/src/pldm.md
```diff
@@ -0,0 +1,268 @@
...
+///
+/// # Methods
+///
+/// * `start_service` - Asynchronously starts the PLDM firmware update service.
+/// * `stop_service` - Asynchronously stops the PLDM firmware update service.
+///
+/// # Errors
+///
+/// Both methods return a `Result` which, on error, contains a `PldmServiceError`.
+pub trait PldmFwUpdateServiceMgr {
```
swenson, line 154: It would be good here and earlier to specify which components are in the user application, user-mode APIs, kernel capsules, and kernel drivers.

That would help understand which traits are called from which contexts as well.

docs/src/pldm.md
```diff
@@ -0,0 +1,268 @@
...
+/// This struct is responsible for handling the PLDM command interface
+/// by maintaining a collection of command handlers and the PLDM context.
+///
+/// # Fields
+///
+/// * `handlers` - A hashmap that maps command codes to their respective command handlers.
+/// * `context` - The PLDM context used for processing commands.
+/// * `other fields` - Additional fields required for the command interface responder.
+pub struct CmdInterfaceResponder {
+    handlers: HashMap<u8, Box<dyn CommandHandler>>,
```
swenson, line 260: I think we're not going to have `HashMap` and `Box` available, since they require allocations. Instead, we should probably use the `LinkList` structures that some of the virtualizers and muxes use: https://github.com/tock/tock/blob/master/capsules/core/src/virtualizers/virtual_alarm.rs#L46

docs/src/pldm.md
```diff
@@ -0,0 +1,311 @@
...
+///   - Returns the size of the processed data.
+///
+/// - `send_notification()`:
+///   Sends a notification to other modules or APIs asynchronously.
+///
+/// - `receive_notification()`:
+///   Receives a notification from other modules or APIs asynchronously.
+pub trait CommandHandler: Send + Sync {
+    async fn execute(&self, payload: &mut [u8], context: &mut PldmContext) -> usize;
+    async fn send_notification(&self);
```
mlvisaya, line 286: Can the notification registration be on PldmFwUpdateService since the API will be interfacing with the PldmFwUpdateService and needs to register some notification callbacks?
Also, we may not need to notify all PLDM commands to the application, only a handful needs to be notified. Let me know what you think or could you provide an example how API can register callbacks with the current approach.

I was thinking something like this:

```
// Define a trait Notifications that the API will implement. The trait methods will be called when PldmFwUpdateService needs to notify the specific event
trait Notifications {
	fn on_request_update_received(payload: &[u8]) -> Result<(),ErrCode>;
	fn on_image_verifypayload: &[u8]() -> Result<(),ErrCode>;
	fn on_image_apply(payload: &[u8]) -> Result<(),ErrCode>;
	fn on_update_component_received(payload: &[u8]) -> Result<(),ErrCode>;
	fn on_activate_firmware_received(payload: &[u8]() -> Result<(),ErrCode>;
}


// Add the Notifications to the service when instantiating it
pub struct PldmFwUpdateService<T: Mctp, U: MessageResponder, R: MessageRequester, N: Notifications> {
    transport: T,
    cmd_interface_responder: MessageResponder,
    cmd_interface_requester: MessageRequester,
    notifications: N
    /// other fields
}


// in RequestUpdate Command Handler, call the appropriate notifications function.
// For example
pub struct RequestUpdateCmdHandler {}
impl CommandHandler for RequestUpdateCmdHandler {
	async fn execute(&self, payload: &mut [u8], context: &mut PldmContext) -> usize {
		let result = self.notifications.on_request_update_received(payload);
		// check result
		// do other stuff

	}
}

```

helloxiling: @mlvisaya,  thanks a lot for your thoughtful suggestion on the notification mechanism. I think this approach makes perfect sense. I will update the interface to reflect this.

mlvisaya (approved):
lgtm

swenson (approved):
LGTM

PR #100, opened 2025-01-30
Start recovery flow between Caliptra Core, MCU, and fake BMC

swenson:
We start implementing a simulated recovery flow engine in a fake BMC,
and having it work with Caliptra Core and MCU bootup.

This depends on an event bus being hooked up between Caliptra Core, the
MCU, and our fake BMC, which is processed on each end to translate
events. For example, recovery block reads and writes from the BMC are
translated to recovery registers available through AXI in Caliptra Core.

This flow is a bit complex but is mostly complete.

* Emulators start and enable active mode
* MCU ROM boots and sets up Caliptra Core's fuse registers
* MCU ROM starts the recovery flow
* Calipta Core ROM boots and requests the first recovery image
  (Caliptra's own firmware)
* Caliptra Core downloads the firmware successfully
* Caliptra Core boots to its firmware
* Caliptra Core downloads the SoC manifest
* Caliptra Core loads the MCU firmware to the MCU SRAM
* MCU boots into its firmware

There is also an integration test added for this entire flow.

The flow can also be tested manually:

Assuming you have `caliptra-sw` in `../caliptra-sw`, have built the ROM
(`cd rom/dev && make build-rom`) and the signed Caliptra firmware
(`cd rom/dev && make build-fw-image`), and have a file (doesn't yet
matter the contents) `soc-manifest.bin`, you can test this whole flow
with:

```shell-session
$ cargo xtask runtime --caliptra-rom ../caliptra-sw/target/riscv32imc-unknown-none-elf/firmware/caliptra-rom.bin --caliptra-firmware ../caliptra-sw/target/riscv32imc-unknown-none-elf/firmware/caliptra-rom-test-fw --soc-manifest soc-manifest.bin --active-mode
```


tests/integration/src/lib.rs
```diff
@@ -73,13 +77,53 @@ mod test {
         output
     }
 
+    fn write_soc_manifest() -> PathBuf {
+        let path = (&*PROJECT_ROOT).join("target").join("soc-manifest");
```
mlvisaya, line 81: nit: I think this can be simplified to: PROJECT_ROOT.join("target").join("soc-manifest") because it will automatically deref when join() is called. also applies to other instances below

tests/integration/src/lib.rs
```diff
@@ -73,13 +77,53 @@ mod test {
...
+        std::fs::write(&path, &[1, 2, 3]).unwrap();
```
mlvisaya, line 82: nit: The & may not be needed for [1, 2, 3], since it's automatically converted to slice.

tests/integration/src/lib.rs
swenson: Good catches. I'm surprised clippy let me do this. :)

mlvisaya (approved):
lgtm

swenson:
Thanks!

PR #101, opened 2025-01-31
Compliance test fix for latest sail-riscv

swenson:
They switched to CMake from make, so we need to use a slightly different
way of building the official simulator.


mlvisaya (approved):

swenson:
Thanks!

PR #102, opened 2025-02-03
Fix the hello.elf build script to reference correct files

swenson:
This was slowing down builds slightly since it would force that binary to be recompiled every time.

With it referencing the correct files, the build can be cached.


mlvisaya (approved):

swenson:
Thanks!

PR #104, opened 2025-02-05
registers: Implement emulated register arrays correctly

swenson:
We were ignoring the array dimension, which caused the automatic peripherals to be missing registers.

We also simplified the `match` arms so that there would not be a ridiculous number of arms for the different addresses for large arrays.

While we're here, we can go ahead and remove the `size` paramter from peripheral methods. It's never used as we always have 32-bit registers.


mlvisaya (approved):
lgtm. (i didn't review the generated ones).

swenson:
Thanks!

PR #107, opened 2025-02-18
Add PLDM common library used by emulator and runtime fw

helloxiling:


helloxiling:
This PR adds the support on mandatory PLDM messages and encoding/decoding mechanisms for the PLDM base protocol and firmware update protocol. This library will be utilized by both the emulator and runtime firmware.

common/pldm/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+[package]
+name = "pldm-common"
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+
+[dependencies]
+zerocopy.workspace = true
+bitfield.workspace = true
+bit-vec.workspace = true
```
swenson, line 12: I don't think we can rely on `bit-vec` in the firmware. By default, it uses `std`.

common/pldm/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+// Licensed under the Apache-2.0 license
+
```
swenson, line 2: If this will be used by the firmware, we should declare it to be `no_std`.

common/pldm/src/protocol/version.rs
```diff
@@ -0,0 +1,265 @@
...
+    fn try_from(version: ProtocolVersionStr) -> Result<Self, Self::Error> {
+        // alpha can be attached to either minor or update
+
+        let mut version_parts = version.split('.').collect::<Vec<ProtocolVersionStr>>();
+        if version_parts.len() < 2 || version_parts.len() > 3 {
+            return Err("Invalid version format".to_string());
+        }
+
+        // Extract alpha, update, minor and major from the version string
+        let alpha = if version_parts[1].chars().last().unwrap().is_alphabetic() {
```
swenson, line 55: I think it would be better to check that it is not empty before we call `unwrap()` to avoid the possible panic.

swenson (commented):
Looking good; a few comments about `std` since it will be used by firmware.

common/pldm/src/protocol/base.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+bitfield! {
+    #[repr(C)]
+    #[derive(Copy, Clone, FromBytes, IntoBytes, Immutable, PartialEq)]
+    pub struct PldmMsgHeader(MSB0 [u8]);
+    impl Debug;
+    pub u8, ic, _: 0, 0;
+    pub u8, msg_type, set_msg_type: 7, 1;
```
mlvisaya, line 195: I can't seem to find this 6-bit msg_type field (bits 1-7) in the PLDM protocol spec. I also don't see ic field for bit 0. 

Can you point me which field is this?
```ascii
+//!  +--------+--------+--------+--------+----------+
+//!  | Rq | D | R |Instance ID | Hdr Ver | PLDM Type|
+//!  +----------------------------------------------+
+//!  |       PLDM Command Code       | Completion*  |
+//!  +----------------------------------------------+
+//!  |                PLDM Message Payload*         |
+//!  +----------------------------------------------+
+//! ```
+//!
+//! - **Rq** (Request bit): 1 bit
+//! - **D** (Data Integrity Check): 1 bit
+//! - **R** (Reserved): 1 bit
+//! - **Instance ID**: 5 bits
+//! - **Hdr Ver** (Header Version): 2 bits
+//! - **PLDM Type**: 6 bits
+//! - **PLDM Command Code**: 8 bits
+//! - **PLDM Completion Code**: 8 bits (optional)
+//! - **Payload**: Variable-length message payload (optional)

File: common/pldm/src/protocol/base.rs [hunk 5]
```diff
@@ -0,0 +1,343 @@
...
+    pub u8, instance_id, set_instance_id: 12, 8;
+    pub u8, reserved, _: 13, 13;
+    pub u8, datagram, set_datagram: 14, 14;
+    pub u8, rq, set_rq: 15,15;
+    pub u8, pldm_type, set_pldm_type: 21, 16;
+    pub u8, hdr_ver, set_hdr_ver: 23, 22;
+    pub u8, cmd_code, set_command_code: 31, 24;
+}
+
+impl PldmMsgHeader<[u8; PLDM_MSG_HEADER_LEN]> {
```
mlvisaya, line 205: I tested creating a message header and checked the output, but it looks like it's not what I expected.
``` rust
use pldm_common::protocol::base::{PldmMsgHeader, PldmMsgType, PldmControlCmd, PldmSupportedType};
    #[test]
    fn test_encode_header() {
        // create a header
        let header = PldmMsgHeader::new(
            0x0a,
            PldmMsgType::Request,
            PldmSupportedType::FwUpdate,
            PldmControlCmd::GetPldmTypes as u8
        );
        // Print bytes as hex
        let raw_bytes = header.0;
        println!(
            "Hex Dump: {}",
            raw_bytes.iter().map(|b| format!("{:02X}", b)).collect::<Vec<String>>().join(" ")
        );
        
        // Expected header bytes : 2a 05 04

        // Expected header to be size 3 only (from rq to command code)
        assert_eq!(raw_bytes.len(), 3);

        // Expect byte0 to be 0x2a
        assert_eq!(raw_bytes[0], 0x2a);

        // Expect byte1 to be 0x05
        assert_eq!(raw_bytes[1], 0x05);

        // Expect byte2 to be 0x04
        assert_eq!(raw_bytes[2], 0x04);


    }
```

Output:
```
---- test_packet_codec::test_encode_header stdout ----
Hex Dump: 01 51 14 04
thread 'test_packet_codec::test_encode_header' panicked at emulator/bmc/pldm/tests/test_packet_codec.rs:25:9:
assertion `left == right` failed
  left: 4
 right: 3
```

common/pldm/Cargo.toml
helloxiling: Thanks for catching this issue. Fixed.

Cargo.toml
```diff
@@ -53,6 +53,7 @@ members = [
     "tests/hello",
     "tests/integration",
     "xtask",
+    "common/pldm",
```
swenson, line 56: nit: it would be good to keep these sorted

Cargo.toml
```diff
@@ -125,6 +124,7 @@ registers-generated = { path = "registers/generated-firmware" }
 registers-generator = { path = "registers/generator" }
 registers-systemrdl = { path = "registers/systemrdl" }
 romtime = { path = "romtime" }
+pldm-common = { path = "common/pldm"}
```
swenson, line 127: nit: also would be good to keep this sorted

common/pldm/src/message/firmware_update/activate_fw.rs
```diff
@@ -0,0 +1,137 @@
...
+
+impl PldmCodec for ActivateFirmwareRequest {
+    fn encode(&self, buffer: &mut [u8]) -> Result<usize, PldmCodecError> {
+        let bytes = core::mem::size_of::<ActivateFirmwareRequest>();
+        if buffer.len() < bytes {
+            return Err(PldmCodecError::BufferTooShort);
+        };
+
+        self.write_to(&mut buffer[..bytes]).unwrap();
+        Ok(bytes)
```
swenson, line 50: You might be able to simplify this a little since `write_to_prefix` checks the length with something like this. (It also helps to avoid `.unwrap()`.)

```suggestion
        self.write_to_prefix(buffer).map_err(|_| PldmCodecError::BufferTooShort).map(|_| bytes)
```

common/pldm/src/message/firmware_update/activate_fw.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    fn decode(buffer: &[u8]) -> Result<Self, PldmCodecError> {
+        let bytes = core::mem::size_of::<ActivateFirmwareRequest>();
+        if buffer.len() < bytes {
```
swenson, line 55: Similarly I think this could be simplified a little with `read_from_prefix` and `map_err` / `map`

common/pldm/src/codec.rs
```diff
@@ -0,0 +1,36 @@
...
+    BufferTooShort,
+    Unsupported,
+}
+
+/// A trait for encoding and decoding PLDM (Platform Level Data Model) messages.
+///
+/// This trait provides methods for encoding a PLDM message into a byte buffer
+/// and decoding a PLDM message from a byte buffer. Implementers of this trait
+/// must also implement the `Debug` trait and be `Sized`.
+pub trait PldmCodec: core::fmt::Debug + Sized {
```
swenson, line 14: I think we could remove a lot of the `impl` code below if we provide a default `impl` for `FromBytes + IntoBytes`, since the code is identical for all of those (they simply call `write_to` and `read_from`), and only make custom `impl`s when needed (for things that can't use `zerocopy`).

swenson (commented):
Looking good. I made a few suggestions.

I think you will also need to remove the `caliptra-ss` submodule changes? (I don't think they are needed.)

swenson (approved):
LGTM

helloxiling:
> Also, try to fix the `caliptra-ss` update if you can

Yes, Chris. I just pushed the fix.

PR #117, opened 2025-02-28
Implement PLDM Discovery Protocol

mlvisaya:
- Added a Daemon that spawns 2 threads:
rx_thread - Receives packets from MCTP and passes it to the event thread
for handling
event_thread - Handles events generated by application, rx_thread and
state machine

- Added the PLDM discovery state machine - responsible for handling
discovery events (e.g. reception of PLDM discovery packets) and sending
out requests to device


emulator/bmc/pldm-ua/src/event_queue.rs
```diff
@@ -0,0 +1,121 @@
...
+    ///
+    /// # Returns
+    /// A new `EventQueue` instance with a cloned sender.
+    ///
+    /// **Note:** The receiver cannot be cloned, so the new instance will not have a receiver.
+    /// This means that only the original `EventQueue` instance should be used to dequeue events.
+    fn clone(&self) -> Self {
+        Self {
+            sender: self.sender.clone(),
+            receiver: None, // Receiver cannot be cloned; only the original has a receiver.
```
swenson, line 68: Since we can't clone the receiver, I'm not sure what the benefit is of having a `EventQueue`.

It seems like we get the same effect by passing around `Sender<Event>`, and passing the single `Receiver<Event>` to the function that uses it.

swenson (commented):
Also, does this need to be added to the emulator binary?

emulator/bmc/pldm-ua/tests/test_discovery.rs
```diff
@@ -0,0 +1,193 @@
...
+    // Create socket used by the PLDM daemon (update agent)
+    let ua_sock = transport.create_socket(ua_sid, fd_sid).unwrap();
+
+    // Create socket to be used by the device (FD)
+    let fd_sock = transport.create_socket(fd_sid, ua_sid).unwrap();
+
+    // Run the PLDM daemon
+    let mut daemon = PldmDaemon::run(ua_sock, Options::default());
+
+    const DEVICE_TID: u8 = 0x01;
```
helloxiling, line 74: PLDM base spec defines `UNASSIGNED_TID` as 0 and we implemented that in MCU PLDM lib.  To avoid hardcode TID, we can excise the full discovery flow and issue `set_tid` command first.

helloxiling (commented):
Overall the PR looks good!  Just added a comment to enhance the flexibility of discovery flow.

mlvisaya:
> Also, does this need to be added to the emulator binary?

yes, eventually it will be added in the main app. I plan to add it when i integrate with Xiling's code on the device side. It will be on another PR.

emulator/bmc/pldm-ua/src/event_queue.rs
mlvisaya: That is true. It's confusing to the one using the event_queue why only sender can be cloned. I'll just use the raw mpsc channels as you suggested.

emulator/bmc/pldm-ua/tests/test_discovery.rs
mlvisaya: Got it. I will send the Set_TID first before GET_TID.

swenson (approved):
LGTM

mlvisaya:
Thank you!

PR #119, opened 2025-03-01
fpga: Add the kernel modules

swenson:
For the `uio` devices (to access the FPGA memory from Linux) and the ROM backdoors (for injecting the ROMs into the FPGA).

This is modified a bit from the standard Caliptra backdoor so that we can define an additional backdoor for the MCU ROM.

These can be build and installed with `cargo xtask fpga-install-kernel-modules`.

The ROM backdoor devices are not used in the current Caliptra tests, but will be useful for MCU because we are running out of `uio` segments in our `uio` device (we're limited to 5).

These modules were tested on the Caliptra Core (2.0) FPGA -- the smoke test passes, and I was able to dump the Caliptra ROM.

(Note: if you run this on newer Linux kernels (>5.x), then it is likely to fail with error messages about `class_create` having too many arguments and a type difference in `mychardev_uevent`. The interfaces for these changed slightly. This can be fixed for Linux 6.x by removing the `THIS_MODULE` argument from `class_create` and making the first parameter of `mychardev_uevent` be `const`. However, the compiled modules will not work unless the FPGA is connected, and could possibly crash your Linux kernel.)


swenson:
Thanks!

PR #120, opened 2025-03-02
spdm: Set exit status correctly; fix target address in tests

swenson:
This fixes a few issues

* We cancel the test if it hangs for more than 30 seconds
* We correct the target address (instead of the tag) when launching the test
* We reduce the wait from 5 seconds to 0.5 after sending a message
* If scheduling an upcall fails, we panic instead of ignoring the error.
* If the test fails, we set the exit code correctly.

I believe the SPDM validator test was failing before, but the exit code was not being set, so it appeared to be passing.

However, the SPDM validator test is still failing, but I believe it is correct to still be failing, since the failure is on a message type that we do not support yet. So, I've left the test disabled for now.

(It looks like Cargo will no longer automatically install the toolchain and needed modules, so I updated GHA to do that manually.)

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -73,8 +73,10 @@ jobs:
             core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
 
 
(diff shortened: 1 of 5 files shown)


mlvisaya (approved):

swenson:
Thanks!

PR #121, opened 2025-03-03
Recovery: correctly generate the SoC manifest in integration test

swenson:
Using the Caliptra auth manifest generator.

We had to update the `caliptra-ss` registers so that the PQC key type is available to set.

(The only supported key type is LMS for the SoC auth manifest for now.)

Tested as working with when combined with https://github.com/chipsalliance/caliptra-sw/pull/1989


mlvisaya (approved):
lgtm. i didn't look at generated-* files

helloxiling (approved):
Looks good to me.

swenson:
Thanks!

swenson:
Thanks!

PR #122, opened 2025-03-03
Refactor: use caliptra-emu-types

swenson:
There's no difference between our two emulator types.

Using a single package helps some future PRs involving the HwModel and the two emulators communicating, so that we don't have to convert between the types.

We can't (yet) share the bus and cpu packages since we have slightly different implementations. The MCU has a more accurate interrupt controller, supports timer priorities, and has timer interrupts. We could potentially do a separate PR to add those features to the Caliptra Core emulator and consolidate further.


mlvisaya (approved):
seems to be straightforward replacement

swenson:
Thanks!

PR #123, opened 2025-03-03
Fix docs build: rustup needs -y and curl doesn't like it

swenson:
Also make it so that the toolchain is grabbed from `rust-toolchain.toml`.


mlvisaya (approved):

parvathib (approved):

swenson:
Thanks!

PR #125, opened 2025-03-04
RISC-V compliance tests: install rustup toolchain manually

swenson:
And also do some other workflow cleanup (use checkout instead of clone, use working-directory instead of `cd`/`pushd`, `popd` is not necessary, checkout fixed git commits, and pre-install GMP.


helloxiling (approved):
Looks good to me.

swenson:
Thanks!

PR #126, opened 2025-03-05
Add Platforms support for the ROM

swenson:
The goal is to support a Tock-like build system so that various platforms (or vendors, integrators, etc.) can build their own ROM using our SDK as a guide. This should support both in-tree and out-of-tree development using our crates and tools.

I moved the emulator-specific ROM code to `platforms/emulator/rom` to act as an example for how such a platform ROM could be structured. All of the common, non-emulator specific code stays in `rom/` (and `romtime/`).

I took a similar approach to how `Tock` uses its debug handler for the ROM fatal error handler and the `romtime::print` function handler: these handlers can be defined at ROM (and runtime) startup, and then used by the ROM and runtime code.

I added some guidelines for our platforms to the README as well.

We can refactor `runtime` to use a similar system in the near future.


mlvisaya (approved):
lgtm

swenson:
Thanks!

PR #127, opened 2025-03-05
Add pldm-lib and app in runtime for PLDM discovery flow support

helloxiling:


emulator/app/src/tests/pldm_request_response_test.rs
```diff
@@ -1,90 +1,207 @@
...
+        }
+        Ok(())
+    }
+
+    pub fn run(socket: MctpPldmSocket, running: Arc<AtomicBool>) {
+        std::thread::spawn(move || {
+            print!("Emulator: Running PLDM Loopback Test: ",);
+            let mut test = PldmRequestResponseTest::new(socket, running);
+            if test.test_send_receive().is_err() {
+                println!("Failed");
```
swenson, line 90: If we fail the test, do we need to trigger the emulator to exit with a non-zero code so that the test will register as failed by `cargo`?

runtime/apps/pldm/pldm-lib/src/control_context.rs
```diff
@@ -0,0 +1,803 @@
...
+            .any(|cap| cap.pldm_type == pldm_type)
+    }
+
+    pub fn is_supported_version(
+        &self,
+        pldm_type: PldmSupportedType,
+        protocol_version: Ver32,
+    ) -> bool {
+        let mut versions = [0; 4];
+        let count = self.get_protocol_versions(pldm_type, &mut versions);
```
swenson, line 134: suggestion: we could just directly check `self.capabilities.find(|cap| cap.pldm_type == pldm_type && cap.protocol_version == protocol_version).is_some()`

runtime/apps/pldm/pldm-lib/src/control_context.rs
```diff
@@ -0,0 +1,803 @@
...
+            .iter()
+            .any(|cap| cap.pldm_type == pldm_type)
+    }
+
+    pub fn is_supported_version(
+        &self,
+        pldm_type: PldmSupportedType,
+        protocol_version: Ver32,
+    ) -> bool {
+        let mut versions = [0; 4];
```
swenson, line 133: is there a maximum number of versions or capabilities we could use as a `const` for these arrays, and use that `const` everywhere?

runtime/apps/pldm/pldm-lib/src/daemon.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    pub async fn start(&mut self) -> Result<(), PldmServiceError> {
+        if self.running.load(Ordering::SeqCst) {
+            return Err(PldmServiceError::StartError);
+        }
+
+        self.running.store(true, Ordering::SeqCst);
+        let mut msg_buffer = [0; MAX_MCTP_PLDM_MSG_SIZE];
+        while self.running.load(Ordering::SeqCst) {
+            let _ = self.cmd_interface.handle_msg(&mut msg_buffer).await;
```
swenson, line 55: Maybe we should add a TODO to add a timeout for this? Setting the `running` to `false` won't interrupt the `.await`, after all.

swenson (commented):
Looking good so far. A few comments.

emulator/app/src/tests/pldm_request_response_test.rs
swenson: If that is the case, then it will never return `Err`?

mlvisaya: that's a good point, yes, in this case it will not return Err.  You're right, I think instead of assert_eq on line 77, need to return Err, and exit(non-zero-code) on line 90

helloxiling: > that's a good point, yes, in this case it will not return Err. You're right, I think instead of assert_eq on line 77, need to return Err, and exit(non-zero-code) on line 90

@mlvisaya, I will address this in next patch.

runtime/apps/pldm/pldm-lib/src/daemon.rs
helloxiling: Thanks for catching this. I tried to leverage `embassy_time` to implement timeout logic, but running into some issues. I put a TODO for now. Will add that in next PRs.

emulator/app/src/tests/pldm_request_response_test.rs
```diff
@@ -1,90 +1,207 @@
...
+/// The emulator sends out different PLDM requests and expects a corresponding response for those requests.
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
+use crate::mctp_transport::MctpPldmSocket;
+use pldm_common::codec::PldmCodec;
+use pldm_common::message::control::*;
+use pldm_common::protocol::base::*;
+use pldm_common::protocol::firmware_update::*;
+use pldm_ua::transport::{PldmSocket, PldmTransportError};
```
mlvisaya, line 13:
```suggestion
use std::process::exit;
use std::sync::atomic::{AtomicBool, Ordering};
use std::sync::Arc;

use crate::mctp_transport::MctpPldmSocket;
use pldm_common::codec::PldmCodec;
use pldm_common::message::control::*;
use pldm_common::protocol::base::*;
use pldm_common::protocol::firmware_update::*;
use pldm_ua::transport::PldmSocket;
```

emulator/app/src/tests/pldm_request_response_test.rs
```diff
@@ -1,90 +1,207 @@
...
+        for message_pair in &self.test_messages {
+            self.socket.send(&message_pair.request)?;
+            let rx_pkt = self.socket.receive(None)?;
+            assert_eq!(
+                rx_pkt.payload.data[..rx_pkt.payload.len],
+                message_pair.response
+            );
+        }
+        Ok(())
+    }
```
mlvisaya, line 83:
```suggestion
    pub fn test_send_receive(&mut self) -> Result<(), ()> {
        self.socket.connect().map_err(|_| ())?;

        for message_pair in &self.test_messages {
            self.socket.send(&message_pair.request).map_err(|_| ())?;
            let rx_pkt = self.socket.receive(None).map_err(|_| ())?;
            if rx_pkt.payload.data[..rx_pkt.payload.len] != message_pair.response {
                return Err(());
            }
        }
        Ok(())
    }
```

emulator/app/src/tests/pldm_request_response_test.rs
mlvisaya, line 90:
```suggestion
                println!("Failed");
                exit(-1);
```

mlvisaya (commented):
added suggestion to exit the test with a non-zero code on failures

swenson (approved):
LGTM

PR #128, opened 2025-03-07
[FW Update Agent] Implement PLDM Firmware Update Inventory Commands

mlvisaya:
1. Add Update Agent State machine template
2. Implement the following commands: QueryDeviceIdentifier
GetFirmwareParameters
RequestUpdate
3. Change FirmwareParameters struct, support more than 1 component entry since Caliptra Firmware Package will include Caliptra FMC, SoC Manifest, MCU RT and other SoC images
4. Add Update Agent SM events to Daemon
5. Store discovery_sm and update_sm in the Daemon so that unit tests can get the states of the SMs
6. Add unit tests

### DIFF
--- a/common/pldm/src/protocol/firmware_update.rs
+++ b/common/pldm/src/protocol/firmware_update.rs
@@ -12,6 +12,7 @@ pub const PLDM_FWUP_BASELINE_TRANSFER_SIZE: usize = 32;
 pub const PLDM_FWUP_MAX_PADDING_SIZE: usize = PLDM_FWUP_BASELINE_TRANSFER_SIZE;
 pub const PLDM_FWUP_IMAGE_SET_VER_STR_MAX_LEN: usize = 255;
 pub const DESCRIPTOR_DATA_MAX_LEN: usize = 64; // Arbitrary limit for static storage
+pub const MAX_COMPONENT_COUNT: usize = 8; // Arbitrary limit, change as needed
 
 #[repr(u8)]
 pub enum FwUpdateCmd {
@@ -270,6 +271,16 @@ pub struct Descriptor {
     pub descriptor_data: [u8; DESCRIPTOR_DATA_MAX_LEN],
 }
 
+impl Default for Descriptor {
+    fn default() -> Self {
+        Descriptor {
+            descriptor_type: 0,
+            descriptor_length: 0,
+            descriptor_data: [0; DESCRIPTOR_DATA_MAX_LEN],
+        }
+    }
+}
+
 impl Descriptor {
     pub fn new_empty() -> Self {
         Descriptor {
@@ -360,7 +371,7 @@ impl PldmCodec for Descriptor {
 }
 
 bitfield! {
-    #[derive(Clone, Copy, FromBytes, IntoBytes, Immutable, PartialEq, Eq)]
+    #[derive(Clone, Copy, FromBytes, IntoBytes, Immutable, PartialEq, Eq, Default)]
     pub struct FirmwareDeviceCapability(u32);
     impl Debug;
     pub u32, reserved, _: 31, 10;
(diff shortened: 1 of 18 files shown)


common/pldm/src/protocol/firmware_update.rs
```diff
@@ -481,7 +492,7 @@ impl<'a> PldmFirmwareVersion<'a> {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, FromBytes, IntoBytes, Immutable)]
+#[derive(Debug, Clone, PartialEq, Eq, FromBytes, IntoBytes, Immutable, Copy)]
```
swenson, line 495: This is a really big type to implement the `Copy` trait. It would be better if we could pass references to it when needed, or use `.clone()` when necessary.

common/pldm/src/protocol/firmware_update.rs
```diff
@@ -499,14 +510,38 @@ pub struct ComponentParameterEntryFixed {
     pub capabilities_during_update: FirmwareDeviceCapability,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Copy)]
```
swenson, line 513: Ditto here about `Copy`.

emulator/bmc/pldm-ua/tests/test_get_fw_parameters.rs
```diff
@@ -0,0 +1,517 @@
+// Licensed under the Apache-2.0 license
+
+#[cfg(test)]
+mod tests {}
```
swenson, line 4: Not sure why this is here, and not wrapping the entire file? Or possibly we can just remove it since this is a test folder.

emulator/bmc/pldm-ua/tests/test_request_update.rs
```diff
@@ -0,0 +1,296 @@
+// Licensed under the Apache-2.0 license
+
+#[cfg(test)]
+mod tests {}
```
swenson, line 4: Ditto.

emulator/bmc/pldm-ua/tests/test_get_fw_parameters.rs
```diff
@@ -0,0 +1,517 @@
...
+use pldm_fw_pkg::FirmwareManifest;
+use pldm_ua::events::PldmEvents;
+use simple_logger::SimpleLogger;
+
+use pldm_common::codec::PldmCodec;
+use pldm_ua::daemon::{Options, PldmDaemon};
+use pldm_ua::transport::{PldmSocket, PldmTransport};
+use pldm_ua::{discovery_sm, update_sm};
+
+struct TestSetup<
```
swenson, line 32: It seems like some of this code is duplicated with the other tests. We should move it to a common place.

common/pldm/src/protocol/firmware_update.rs
mlvisaya: that makes sense. Initially, I used this because when instantiating an array of this struct, I need to do : 

let mut comp_param_table = [ComponentParameterEntry::default(); MAX_COMPONENT_COUNT]; 

which needs Copy trait. 
But I found out just now I can use the core::array::from_fn instead to initialize the array without using Copy trait, so I will replace it with this.

emulator/bmc/pldm-ua/tests/test_get_fw_parameters.rs
mlvisaya: my mistake, this is not needed. i'll remove it.

emulator/bmc/pldm-ua/tests/test_request_update.rs
mlvisaya: i'll remove it.

emulator/bmc/pldm-ua/tests/test_get_fw_parameters.rs
mlvisaya: that's a good a suggestion. i'll create a struct to place all of this common utility setup.

swenson (approved):
LGTM

mlvisaya:
Ty!

PR #129, opened 2025-03-10
Add support for NEGOTIATE_ALGORITHMS command handling

parvathib:
* Add support for NEGOTIATE_ALGORITHMS command handling in SPDM responder.
* Fix the bug in kernel layout to align the code size to multiple of 4 bytes
* Fix bug in the MCTP_UTIL to reset sec_eid and dest_eid on new request.


runtime/apps/spdm/spdm-lib/src/protocol/version.rs
```diff
@@ -1,6 +1,9 @@
 // Licensed under the Apache-2.0 license
 
-use crate::error::SpdmError;
+use crate::error::{SpdmError, SpdmResult};
+
+const MAX_NUM_SUPPORTED_SPDM_VERSIONS: usize = 4;
+const MAX_SUPORTED_VERSION: SpdmVersion = SpdmVersion::V13;
```
swenson, line 6: typo
```suggestion
const MAX_SUPPORTED_VERSION: SpdmVersion = SpdmVersion::V13;
```

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
```diff
@@ -0,0 +1,650 @@
...
+}
+
+impl Prioritize<MeasurementSpecificationType> for MeasurementSpecification {
+    fn prioritize(
+        self,
+        peer: Self,
+        priority_table: Option<&[MeasurementSpecificationType]>,
+    ) -> Self {
+        let common = self.0 & peer.0;
+        if let Some(priority_table) = priority_table {
```
swenson, line 41: This code looks basically identical for all of the different types.

Could you do one `impl` for `u8`, or maybe create a `derive` macro? (Here is an example of a `derive` macro: https://github.com/chipsalliance/caliptra-sw/blob/main/cfi/derive/src/lib.rs#L101)

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
```diff
@@ -0,0 +1,650 @@
...
+                let priority_spec: MeasurementSpecification = priority.into();
+                if common & priority_spec.0 != 0 {
+                    return MeasurementSpecification(priority_spec.0);
+                }
+            }
+        } else {
+            // If priority_table is None, we assume the default behavior
+            // of returning the first common measurement specification.
+            if common != 0 {
+                return MeasurementSpecification(common & (!common + 1));
```
swenson, line 52: I think this will return the smallest set bit in `common`, but wouldn't we want the highest bit (which seems to be the "best")? I think this will be `u8::BITS - common.leading_zeros()`.

Otherwise, I think it is clearer and faster to use `common.trailing_zeros()` (since we have the `Zbb` instruction set available, this is a single instruction).

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
```diff
@@ -0,0 +1,650 @@
...
+    pub dhe_group: DheNamedGroup,
+    pub aead_cipher_suite: AeadCipherSuite,
+    pub req_base_asym_algo: ReqBaseAsymAlg,
+    pub key_schedule: KeySchedule,
+}
+
+impl DeviceAlgorithms {
+    pub fn num_alg_struct_tables(&self) -> usize {
+        let mut num = 0;
+        if self.dhe_group.0.count_ones() > 0 {
```
swenson, line 603: `self.dhe_group.0 != 0` would be faster

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
```diff
@@ -0,0 +1,650 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum KeyScheduleType {
+    // SPDM Key Schedule
+    SpdmKeySchedule,
+}
+
+#[derive(Debug, Clone, Copy, Default)]
```
swenson, line 586: This is a bit of a large `struct` to implement `Copy`. I'd rather we pass references or use `.clone()` when necessary instead of accidentally making copies.

This will also help to reduce stack usage.

swenson (approved):
LGTM; a few small comments, but approved after they are addressed. (Ideally, I'd like to see `Copy` removed from the big struct and any other large structs, and see if we can reduce the required stack size.)

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
parvathib: changed :)

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
parvathib: removed Copy

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
parvathib: I did a bit of rewrite and have 3 default impl for u8, u16 and u32 that cover all the structures. I didn't have to do procedural macros for this.

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
parvathib: the first common bit is the first lsb bit set. I've changed this to use `trailing_zeros()` and it looks much cleaner now. thanks for the suggestion!

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
```diff
@@ -3,8 +3,84 @@
...
+where
+    T: Copy + Into<u8>,
+{
+    fn prioritize(&self, peer: &Self, priority_table: Option<&[T]>) -> Self {
+        let common = self & peer;
+        if let Some(priority_table) = priority_table {
+            for &priority in priority_table {
+                let priority_spec: u8 = priority.into();
+                if common & priority_spec != 0 {
+                    return common & (!common + 1);
```
swenson, line 24: I think this (and the ones below) should be?

```suggestion
                    return 1 << (common & priority_spec).trailing_zeros();
```

runtime/apps/spdm/spdm-lib/src/protocol/algorithms.rs
```diff
@@ -3,8 +3,84 @@
...
+            }
+        } else {
+            // If priority_table is None, we assume the default behavior
+            // of returning the first common algorithm.
+            if common != 0 {
+                return 1 << common.trailing_zeros();
+            }
+        }
+        0
+    }
```
swenson, line 59: I think these could be slightly simplified.
```suggestion
        let common = self & peer;
        if common == 0 {
            0
        } else if let Some(priority_table) = priority_table {
            for &priority in priority_table {
                let priority_spec: u16 = priority.into();
                if common & priority_spec != 0 {
                    return common & (!common + 1);
                }
            }
        } else {
            // If priority_table is None, we assume the default behavior
            // of returning the first common algorithm.
            1 << common.trailing_zeros()
        }
    }
```

swenson (approved):
LGTM; made some minor comments

runtime/apps/spdm/spdm-app/src/main.rs
```diff
@@ -67,46 +79,37 @@ pub(crate) async fn async_main<S: Syscalls>() {
...
 
     let local_capabilities = DeviceCapabilities {
         ct_exponent: CALIPTRA_SPDM_CT_EXPONENT,
-        flags: capability_flags,
+        flags: device_capability_flags(),
         data_transfer_size: max_mctp_spdm_msg_size,
         max_spdm_msg_size: max_mctp_spdm_msg_size,
     };
 
+    let local_algorithms = LocalDeviceAlgorithms {
```
mhatrevi, line 93: Suggest calling this type something like LocalDeviceConfig as it has other non-algorithm settings such as Measurement Specification, Opaque data type etc.

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
```diff
@@ -0,0 +1,496 @@
...
+use crate::protocol::capabilities::*;
+use crate::protocol::common::SpdmMsgHdr;
+use crate::protocol::SpdmVersion;
+use crate::state::ConnectionState;
+use bitfield::bitfield;
+use libtock_platform::Syscalls;
+use zerocopy::{FromBytes, Immutable, IntoBytes};
+
+// Max request length shall be 128 bytes (SPDM1.3 10.4, Table 10.4)
+const MAX_REQUEST_LENGTH: u16 = 128;
```
mhatrevi, line 17: Suggest: MAX_SPDM_REQUEST_LENGTH

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
```diff
@@ -0,0 +1,496 @@
...
+    ext_asyn_count: u8,
+    ext_hash_count: u8,
+    reserved_2: u8,
+    mel_specification: MelSpecification,
+}
+
+impl NegotiateAlgorithmsReq {
+    fn min_req_len(&self) -> u16 {
+        let total_alg_struct_len =
+            core::mem::size_of::<AlgStructure>() * self.num_alg_struct_tables as usize;
```
mhatrevi, line 41: Add use crate core::mem at the top to cleanup the usage.

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
```diff
@@ -0,0 +1,496 @@
...
+    })?;
+
+    // Reserved fields check
+    if req.param2 != 0 || req.reserved_1 != [0; 12] || req.reserved_2 != 0 {
+        Err(ctx.generate_error_response(req_payload, ErrorCode::InvalidRequest, 0, None))?;
+    }
+
+    // min req length check
+    if req.length > MAX_REQUEST_LENGTH || req.length < req.min_req_len() {
+        return Err(ctx.generate_error_response(req_payload, ErrorCode::InvalidRequest, 0, None));
```
mhatrevi, line 184: Use idiomatic RUST: Err(ctx.generate_error_response(req_payload, ErrorCode::InvalidRequest, 0, None))?;

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
```diff
@@ -0,0 +1,496 @@
...
+    ctx: &mut SpdmContext<'a, S>,
+    spdm_hdr: SpdmMsgHdr,
+    req_payload: &mut MessageBuf<'a>,
+) -> CommandResult<()> {
+    // Validate the state
+    if ctx.state.connection_info.state() != ConnectionState::AfterCapabilities {
+        Err(ctx.generate_error_response(req_payload, ErrorCode::UnexpectedRequest, 0, None))?;
+    }
+
+    // process negotiate algorithms request
```
mhatrevi, line 483: Let's be consistent about the comment capitalization. At some places, the comment starts with a capital letter.

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
parvathib: changed

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
parvathib: done

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
parvathib: Changed here and in a few other places too.

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
```diff
@@ -0,0 +1,496 @@
...
+
+    // min req length check
+    if req.length > MAX_REQUEST_LENGTH || req.length < req.min_req_len() {
+        return Err(ctx.generate_error_response(req_payload, ErrorCode::InvalidRequest, 0, None));
+    }
+
+    // other parameters support check
+    let other_params_support = &req.other_param_support;
+    if other_params_support.reserved1() != 0 || other_params_support.reserved2() != 0 {
+        return Err(ctx.generate_error_response(req_payload, ErrorCode::InvalidRequest, 0, None));
```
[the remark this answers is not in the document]
parvathib: done

runtime/apps/spdm/spdm-lib/src/commands/algorithms_rsp.rs
parvathib: done

swenson (approved):
LGTM

PR #132, opened 2025-03-12
Split builder as a library in a separate package

swenson:
This seems a bit cleaner and it means that the integration tests don't need extra calls to `cargo xtask`.

We also switch to `anyhow` instead of using `DynError`.


mlvisaya (approved):
lgtm

swenson:
Thanks!

PR #133, opened 2025-03-12
Align App binaries to 4-bytes

mlvisaya:
When the App binaries are not aligned to 4-bytes, tock throws an "App flash length not supported by MPU." error.

To fix this we will pad the App binary until it is aligned to 4-bytes, so that the next App's TBF will also be aligned to 4-bytes

The layout would be:

TBF for App1
App1.bin
TBF for App2
App2.bin
...

TBF is 132 bytes which is 4-byte aligned, so we just need to make sure that the app binaries are 4-bytes aligned.

### DIFF
--- a/builder/src/apps.rs
+++ b/builder/src/apps.rs
@@ -113,7 +113,12 @@ fn app_build_tbf(
     }
 
     // read the flat binary
-    let b = std::fs::read(&app_bin)?;
+    let mut b = std::fs::read(&app_bin)?;
+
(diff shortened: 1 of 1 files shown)


swenson (approved):
LGTM; I think this file is now in `builder/`

helloxiling (approved):
Good catch! Looks good to me.

mlvisaya:
ty!

PR #134, opened 2025-03-13
Fix deadlock on when receiving and sending MCTP messages at the same …

mlvisaya:
…time

The MCTP transport allows users to spawn two threads: one for TX and one for RX. Both threads shared the same context variable, which contained the MCTP utility, the I3C TCP stream, and the first_response variable.

This led to a deadlock scenario where:

* The RX thread locks context first.
* The TX thread also attempts to lock context.
* Since during the first message exchange, RX typically waits for a message before TX sends one, both threads could get stuck.

Solution:
* Removed unnecessary dependencies.
* Separate MCTP utility instances for TX and RX, as they don’t share data.
* Cloned the I3C TCP stream for independent TX and RX access, avoiding a shared lock.
* Introduced a conditional variable to protect first_response: The RX thread waits for first_response using cvar.wait(), releasing its lock.
The TX thread can then set first_response without causing a deadlock. This ensures proper synchronization while allowing concurr

### DIFF
--- a/emulator/app/src/mctp_transport.rs
+++ b/emulator/app/src/mctp_transport.rs
@@ -8,6 +8,7 @@
 use crate::tests::mctp_util::common::MctpUtil;
 use core::time::Duration;
 use emulator_periph::DynamicI3cAddress;
+
 use pldm_common::util::mctp_transport::{MctpCommonHeader, MCTP_PLDM_MSG_TYPE};
 use pldm_ua::transport::{
     EndpointId, Payload, PldmSocket, PldmTransport, PldmTransportError, RxPacket,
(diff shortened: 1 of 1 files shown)


emulator/app/src/mctp_transport.rs
```diff
@@ -84,26 +95,47 @@ impl PldmSocket for MctpPldmSocket {
...
-            };
-            context.first_response = None;
-            return Ok(ret);
+        {
+            let (context_lock, cvar) = &*self.context;
+            let mut context = context_lock.lock().unwrap();
+            if context.state == MctpPldmSocketState::FirstResponse
+                || context.state == MctpPldmSocketState::Idle
+            {
+                if context.first_response.is_none() {
```
swenson, line 104: When using condition variables, you need to check the condition after you wake up. Although it is unlikely here, there are other reasons the thread could have woken up.

```suggestion
                while context.first_response.is_none() {
```

swenson (approved):
LGTM assuming we change the `if` to a `while` around the condition variable.

emulator/app/src/mctp_transport.rs
mlvisaya: thanks! that's a good point.

mlvisaya:
ty!

PR #135, opened 2025-03-13
Fix libapi can not be included as device app dependency

mlvisaya:
Problem:
If libapi is included as a dependency of a tock app, the compiler will throw an exception that libapi is dependent on std lib so it can not be added.

Solution:
Add no_std in the libapi module lib.rs. Need to remove the test_checksum_overflow unit test case, since it will cause a stack overflow when run on a RISC compiler.

### DIFF
--- a/runtime/apps/api/src/checksum.rs
+++ b/runtime/apps/api/src/checksum.rs
@@ -28,13 +28,6 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


helloxiling (approved):
Looks good.  Thanks for addressing it promptly!

mlvisaya:
ty!

PR #136, opened 2025-03-14
[FW Update Agent] Implement PASS_COMPONENT and UPDATE_COMPONENT PLDM Commands

mlvisaya:
Support for PLDM PASS_COMPONENT and UPDATE_COMPONENT commands.

PASS_COMPONENT command passes the metadata of each component that needs to be updated one by one to the FD.

UPDATE_COMPONENT indicates to the FD which component will be updated.

Add unit tests for these 2 commands.

### DIFF
--- a/emulator/bmc/pldm-ua/src/update_sm.rs
+++ b/emulator/bmc/pldm-ua/src/update_sm.rs
@@ -8,9 +8,11 @@ use pldm_common::codec::PldmCodec;
 use pldm_common::message::firmware_update as pldm_packet;
 use pldm_common::protocol::base::{
     InstanceId, PldmBaseCompletionCode, PldmMsgHeader, PldmMsgType, PldmSupportedType,
+    TransferRespFlag,
 };
 use pldm_common::protocol::firmware_update::{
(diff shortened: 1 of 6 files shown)


emulator/bmc/pldm-ua/src/update_sm.rs
```diff
@@ -179,12 +183,160 @@ pub trait StateMachineActions {
...
+            component_idx = 0;
+            pass_component_flag = TransferRespFlag::Start;
+        } else if num_components_passed < num_of_components_to_pass - 1 {
+            component_idx = 0;
+            pass_component_flag = TransferRespFlag::Middle;
+        } else if num_components_passed == num_of_components_to_pass - 1 {
+            component_idx = 0;
+            pass_component_flag = TransferRespFlag::End;
+        } else {
+            // This should never happen
```
swenson, line 215: If it should never happen, I'd say we should `panic!` then, so that if it does happen, we'll fail a test or whatever we're doing.

swenson (approved):
LGTM

emulator/bmc/pldm-ua/src/update_sm.rs
mlvisaya: i'll add panic()!

PR #139, opened 2025-03-17
Run Caliptra Core by default; autobuild its FW

swenson:
Caliptra Core will now always run, since we will start using it soon for mailbox commands.

To make it easier to use, we now automatically build Caliptra ROM, FW, and SoC manifest if they are not passed in.

I factored out the `CaliptraBuilder` struct to make it easier for it to reason about when to automatically build the FW if not provided, when to build the SoC manifest if not provided, etc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2193,10 +2193,19 @@ name = "mcu-builder"
 version = "0.0.0"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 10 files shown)


mlvisaya (approved):
lgtm!

swenson:
Thanks!

PR #140, opened 2025-03-17
[FW Update Agent] Implement RequestFirmwareData and TransferComplete PLDM Commands

mlvisaya:
Support for downloading the PLDM Firmware by implementing the RequestFirmwareData and TransferComplete PLDM commands. Some changes are also needed in the pldm-fw-pkg lib.

1. Changes in PLDM-FW-PKG lib.

(The PLDM-FW-PKG library is responsible for encoding and decoding PLDM FW package files. In PLDM FW update, this lib is used to decode the package file, so that the firmware content can be retrieved by the update agent)

- Currently, when the pldm-fw-pkg library decodes a PLDM package file, it will copy the firmware binaries in the pldm-fw-pkg into separate files, it will also return a FirmwareManifest struct that contains the file paths of the binaries created. In this patch, the firmware binary is stored into separate local buffers within the FirmwareManifest struct, so that the PLDM Update Agent module can retrieved the firmware directly from the buffers instead of reading from files.

- Made it optional to write the decoded FW binaries into files.

- Added a unit test to encode a PLDM Firmware package into a package file and then decoding it, and verify if the contents are the same.

2. Changes in PLDM-UA lib.

(The PLDM Update Agent library implementing the PLDM FW update protocol)

- Added command handling of RequestFirmwareData and TransferComplete in the update state machine

- Added unit test to simulate downloading firmware

### DIFF
--- a/emulator/bmc/pldm-fw-pkg/src/manifest.rs
+++ b/emulator/bmc/pldm-fw-pkg/src/manifest.rs
@@ -1579,7 +1579,7 @@ impl ComponentImageInformation {
 
     fn verify(&self) -> Result<(), String> {
         if let Some(image_location) = &self.image_location {
-            // Verify image_location length is less than 255
+            // Verify file exists in the image location
             if fs::metadata(image_location).is_err() {
                 return Err(format!(
                     "Component image file does not exist: {}",
--- a/emulator/bmc/pldm-fw-pkg/src/manifest.rs
+++ b/emulator/bmc/pldm-fw-pkg/src/manifest.rs
@@ -64,7 +64,7 @@ pub struct DownstreamDeviceIdRecord {
 
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
 pub struct ComponentImageInformation {
-    pub image_location: String,
+    pub image_location: Option<String>,
     pub classification: u16,
     pub identifier: u16,
     pub comparison_stamp: Option<u32>, // Optional uint32 based on ComponentOptions
@@ -77,6 +77,8 @@ pub struct ComponentImageInformation {
     pub offset: u32,  // Offset to the start of the image
     #[serde(skip)]
     pub size: u32,    // Size of the image
+    #[serde(skip)]
+    pub image_data: Option<Vec<u8>>, // Optional image data, to be filled when package is decoded
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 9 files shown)


emulator/bmc/pldm-fw-pkg/src/manifest.rs
```diff
@@ -1552,12 +1578,16 @@ impl ComponentImageInformation {
...
 
     fn verify(&self) -> Result<(), String> {
-        // Verify image_location exists
-        if fs::metadata(&self.image_location).is_err() {
-            return Err(format!(
-                "Component image file does not exist: {}",
-                self.image_location
-            ));
+        if let Some(image_location) = &self.image_location {
+            // Verify image_location length is less than 255
```
swenson, line 1582: I don't see anything checking the length?

emulator/bmc/pldm-fw-pkg/tests/test_decode.rs
```diff
@@ -0,0 +1,202 @@
+// Licensed under the Apache-2.0 license
+
+use chrono::Utc;
+/*++
+
+Licensed under the Apache-2.0 license.
+
+--*/
```
swenson, line 8: Delete?
```suggestion
```

swenson, line 8: Delete?
```suggestion
```

emulator/bmc/pldm-ua/src/update_sm.rs
```diff
@@ -551,18 +555,100 @@ pub trait StateMachineActions {
...
+                return send_request_helper(&ctx.socket, &response);
+            }
+            let mut buffer = [0u8; MAX_TRANSFER_SIZE as usize];
+            let mut to_copy = min(
+                data.len() as i32 - request.offset as i32,
+                request.length as i32,
+            );
+            to_copy = max(to_copy, 0); // Ensure to_copy is not negative
+            let mut to_pad = min(
+                0,
```
swenson, line 593: I don't think this is correct: `to_pad` will always be `0` after the `min` and `max`.

emulator/bmc/pldm-fw-pkg/src/manifest.rs
mlvisaya: oops wrong comment! there's no limitation on image_location length, this logic should just check if the path exists in the file system. I'll change the comment.

emulator/bmc/pldm-ua/src/update_sm.rs
mlvisaya: good catch! the first min, should limit it to the requested length, not 0. so it should be min(request.len,...)

swenson (approved):
LGTM

PR #142, opened 2025-03-19
[FW Update Agent] Support Post Download PLDM Commands

mlvisaya:
Implement the VerifyComplete, ApplyComplete, ActivateFw and CancelUpdateComponent, GetStatus PLDM Commands.

After download is completed, update_sm will wait for VerifyComplete from device, then the ApplyComplete. If there are no more components to update, the update_sm will send an ActivateFw.

If the device has components with 'self-activation' flag set, update_sm will send periodically GetStatus commands until it gets a response that the device has succesfully activated the Firmware.

If there are any failures, in Verify, Apply or Activate stages, the update_sm will send a CancelUpdateComponent and will terminate the update.

A timer module is created to support the timer related actions in the update_sm.

### DIFF
--- a/emulator/bmc/pldm-ua/src/lib.rs
+++ b/emulator/bmc/pldm-ua/src/lib.rs
@@ -4,5 +4,6 @@
 pub mod daemon;
 pub mod discovery_sm;
 pub mod events;
+pub mod timer;
 pub mod transport;
 pub mod update_sm;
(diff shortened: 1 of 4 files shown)


swenson (approved):
LGTM

mlvisaya:
thank you!

PR #143, opened 2025-03-19
Add PLDM FW Update End-to-End Integration test

mlvisaya:
* Add an integration test that verifies the interaction between the
device' PLDM Daemon and the emulator's PLDM Update Agent Daemon.
If the emulator's PLDM Update Agent Daemon received an unexpected PLDM
message from the device, the test will fail.

* Add CancelUpdateComponent as one of the commands that needs to be
supported. This is needed for terminating the PLDM FW Update if an error
occurs.


helloxiling (approved):
Looks good to me.

mlvisaya:
thank you!

PR #145, opened 2025-03-22
Add Caliptra mailbox capsule

swenson:
This implements the syscall driver we have and connects it with the Caliptra mailbox, which is connected in the emulator to the Caliptra emulator.

One problem with the standard Caliptra `SocManager` interface is that it expects the entire message and response to be buffered by the caller. When we are passing data through the kernel capsule, we want to avoid haivng these large buffers.

So, I wrote an adapter layer that uses `impl Iterator<item = u32>`s instead of requiring a full request or response buffer.

I added a simple test that calls `QUOTE_PCRS`.

I found the source for some of alarms not firing as well: we are not always calling `alarm.setup()` and `alarm.set_client()` when creating our `VirtualMuxAlarm`s. I will fix this in a separate PR since we will need to refactor the way we initialize some pieces so that they use static initialization and components, since we'll need to use explicit static lifetimes to call those methods.


runtime/apps/syscall/src/mailbox.rs
```diff
@@ -21,14 +22,37 @@ impl<S: Syscalls> Default for Mailbox<S> {
     }
 }
 
+// Populate the checksum for a mailbox request.
+pub fn populate_checksum(cmd: u32, data: &mut [u8]) {
+    // Calc checksum, use the size override if provided
+    let checksum = caliptra_api::calc_checksum(cmd, data);
+
+    if data.len() < size_of::<MailboxReqHeader>() {
```
mlvisaya, line 30: nit: should we panic or assert here?

runtime/apps/syscall/src/mailbox.rs
```diff
@@ -21,14 +22,37 @@ impl<S: Syscalls> Default for Mailbox<S> {
...
     pub fn new() -> Self {
         Self {
-            syscall: PhantomData,
+            _syscall: PhantomData,
             driver_num: MAILBOX_DRIVER_NUM,
         }
     }
 
+    // Populate the checksum for a mailbox request.
+    pub fn populate_checksum(&self, cmd: u32, data: &mut [u8]) {
```
mlvisaya, line 46: this seems to be a repeated populate_checksum unction on line 25. should we delegate it there?

runtime/apps/api/src/mailbox.rs
```diff
@@ -17,6 +17,10 @@ impl<S: Syscalls> Default for Mailbox<S> {
     }
 }
 
+// TODO: this is redudant with the caliptra_api mailbox, though we don't
```
mlvisaya, line 20:
```suggestion
// TODO: this is redundant with the caliptra_api mailbox, though we don't
```

mlvisaya (approved):
lgtm (with some nits)

runtime/apps/syscall/src/mailbox.rs
swenson: Yeah, I meant to do that. Will fix.

runtime/apps/syscall/src/mailbox.rs
swenson: I'll return a result.

swenson:
Thanks!

PR #147, opened 2025-03-26
Device PLDM FW Update Initiator/Responder Tasks

mlvisaya:
* In the Device PLDM Daemon, create the initiator and responder tasks.
There will be a separate executor for each task which needs to be polled
in order for the tasks to progress.

* Patch the MCTP driver to allow receiving both requests and responses.
Previously, the driver doesn't allow MCTP receive_response and
receive_request at the same time. This was because the driver can only
expect 1 type of message at a time. The request and response share the
same app buffer and the upcall subscribe are also the same for both. To
allow both request and response, the buffers were separated and also
assign separate upcall subscribe numbers for each.

* Fix a bug in MCTP transport where the response is being sent with the
tag owner bit set. For responses, the tag owner should be cleared and
should have the same tag value as the request. The tag owner bit is bit
7 of the first byte of the MCTP header packet and the tag value is the
last 3 bits of the first byte. Before sending the response, make sure
the tag is masked with 0x7 so that the owner tag is cleared.


swenson (approved):
LGTM; I like the use of two executors. Clever!

mlvisaya:
Thanks!

runtime/capsules/src/mctp/driver.rs
```diff
@@ -334,11 +398,33 @@ impl<'a> SyscallDriver for MCTPDriver<'a> {
                 let result = self
                     .apps
                     .enter(process_id, |app, kernel_data| {
-                        if app.pending_tx.is_some() {
+                        let is_request;
```
parvathib, line 401: should this be declared as mutable?

runtime/apps/pldm/pldm-lib/src/transport.rs
```diff
@@ -62,7 +64,7 @@ impl<S: Syscalls> MctpTransport<S> {
         rsp.fill(0);
         let (rsp_len, _msg_info) = if let Some(tag) = self.cur_req_ctx {
             self.mctp
-                .receive_response(rsp, tag)
+                .receive_response(rsp, tag & MCTP_TAG_MASK)
```
parvathib, line 67: is this masking necessary? You should be able to use the tag returned from `send_request` directly. If that's not the case, the tag returned is wrong and would be a bug that needs to be fixed

runtime/capsules/src/mctp/driver.rs
```diff
@@ -334,11 +398,33 @@ impl<'a> SyscallDriver for MCTPDriver<'a> {
...
+                        if command_num == 3 && app.pending_tx_request.is_none() {
+                            app.pending_tx_request = Some(OpContext {
```
parvathib, line 403: It makes sense to separate the request and response contexts for RX messages in the driver layer. The MUX layer will assemble whichever message arrives first and invoke the callback to the driver layer. The driver can then check if either an RX request or response is pending based on the context (Tag, EID) match and invoke the corresponding upcall.

However, this is a bit tricky for TX. We only have one kernel buffer in the driver layer. So, if a transaction is in progress, a new request might either overwrite the message buffer or fail to copy the message to the buffer because it is locked in the MUX layer due to an on-going TX.

IMO, we should only allow one TX at a time. Either reject new requests here or just have one context like before for both request and response.

parvathib (commented):
RX changes look okay. Have a few concerns about TX. Let's address them together.

runtime/capsules/src/mctp/driver.rs
```diff
@@ -235,8 +255,23 @@ impl<'a> MCTPDriver<'a> {
...
+        };
+
+        if !op_ctx.matches(msg_tag, src_eid) {
+            return false;
+        }
+
+        true
+    }
+
+    fn rx_pending_response(&self, app: &mut App, msg_tag: u8, src_eid: u8) -> bool {
```
parvathib, line 273: nit: Could rename this to `pending_rx_response`

runtime/capsules/src/mctp/driver.rs
mlvisaya: I thought at first that it needs to be mutable, but cargo clippy suggests that if the variable is assigned only once through an if block then it doesn't need to be mutable, and will be assigned value when the if statement is evaluated.

runtime/apps/pldm/pldm-lib/src/transport.rs
mlvisaya: In my test, send_request returns a tag of 0x8 (TO=1, TAG=0), when i do a receive_response on 0x8, and the driver receives MCTP message, it complained of "No Pending RX operation". The MCTP message has a tag of 0x0 (TO=0, TAG=0) since it's a response.

runtime/capsules/src/mctp/driver.rs
mlvisaya: Gotcha. I could revert the changes for the TX, and just add an error for now if there's a pending tx.

parvathib (approved):
LGTM

PR #148, opened 2025-03-27
Remove unused libtock libraries

swenson:


runtime/apps/libtock/Cargo.toml
```diff
@@ -13,33 +13,15 @@ version = "0.1.0"
...
-libtock_temperature = { path = "apis/sensors/temperature" }
-
-embedded-hal = { version = "1.0", optional = true }
+embedded-hal = { workspace = true, optional = true }
+libtock_alarm.workspace = true
+libtock_console.workspace = true
+libtock_debug_panic.workspace = true
+libtock_low_level_debug.workspace = true
+libtock_platform.workspace = true
+libtock_rng.workspace = true
```
helloxiling, line 25: Do we need to keep `libtock_rng`?

swenson: We might? I wasn't sure. Caliptra will have an RNG, so we might want to write driver for it.

helloxiling: > We might? I wasn't sure. Caliptra will have an RNG, so we might want to write driver for it.

Ok, we can keep it for now. Will refactor it out if not being used later.

helloxiling (approved):

PR #149, opened 2025-03-28
Add support for SPDM GET_DIGESTS command

helloxiling:


runtime/apps/spdm/spdm-lib/src/hash_op.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// # Returns
+    ///
+    /// * `Ok(())` if the finalization is successful.
+    /// * `Err(HashEngineError)` if an error occurs during finalization.
+    fn finish(&mut self, digest: &mut SpdmDigest) -> Result<(), HashEngineError>;
+}
+
+// A concrete implementation of the `HashEngine` trait utilizing the `sha2` crate.
+// This implementation is intended for development and testing purposes only.
+// Future iterations should refactor this to integrate with a cryptographic service provider via a mailbox interface.
```
swenson, line 78: For Sha384 and Sha512, we'll be able to use Caliptra's SHA mailbox commands. But sadly it won't support SHA256. For small messages (few KB), we could potentially add a new mailbox command just for SHA256 though, but the hardware does not support restore / resume operations for SHA256.

swenson (approved):
LGTM

runtime/apps/spdm/spdm-lib/src/cert_mgr.rs
```diff
@@ -0,0 +1,370 @@
...
+        Ok(())
+    }
+}
+
+impl AsRef<[u8]> for SpdmCertChainData {
+    fn as_ref(&self) -> &[u8] {
+        &self.data[..self.length as usize]
+    }
+}
+// Represents the buffer for the SPDM certificate chain base format as defined in SPDM Specification Table 28.
```
parvathib, line 83: Could you specify the SPDM version number this Table is referred to?

runtime/apps/spdm/spdm-lib/src/commands/digests_rsp.rs
```diff
@@ -0,0 +1,318 @@
...
+    }
+}
+
+impl Codec for GetDigestsResp {
+    fn encode(&self, buffer: &mut MessageBuf) -> CodecResult<usize> {
+        let mut len = self.common.encode(buffer)?;
+        let slot_cnt = self.common.slot_mask.count_ones() as usize;
+        for digest in self.digests.iter().take(slot_cnt) {
+            len += digest.encode(buffer)?;
+        }
```
parvathib, line 128: There are other fields missing in SPDM 1.3 response structure (Refer to Table 35 in Spdm 1.3.2). We don't need to set these fields, and they could be absent in response as per the specification. But we could define them for future use.

parvathib (approved):
LGTM

runtime/apps/spdm/spdm-lib/src/commands/digests_rsp.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+}
+
+pub(crate) fn handle_digests<'a, S: Syscalls>(
+    ctx: &mut SpdmContext<'a, S>,
+    spdm_hdr: SpdmMsgHdr,
+    req_payload: &mut MessageBuf<'a>,
+) -> CommandResult<()> {
+    // Validate the state
+    if ctx.state.connection_info.state() < ConnectionState::AfterNegotiateAlgorithms {
```
mhatrevi, line 137: Suggest: In a later PR, instead of calling generate_error_response at all palces, use a closure: Example: 

    let result: CaliptraResult<()> = (|| {
        // Perform some operations
        if some_condition {
            return Err(CaliptraError::new("An error occurred"));
        }
        // More operations
        Ok(())
    })();

    match result {
        Ok(_) => println!("Operation succeeded"),
        Err(e) => println!("Operation failed: {}", e),
    }

mhatrevi (approved):

PR #150, opened 2025-03-31
New SystemRDL 2.0 parser in winnow

swenson:
Uses the [winnow](https://crates.io/crates/winnow) parser-combinator library to write a new SystemRDL parser.

The new RDL changes to our registers, in particular MCI, don't show up with the existing parser. The existing parser is a bit of a hack, and skips over a lot of the pieces of the spec that we don't support.

The generator also makes a lot of assumptions about how the RDL files are structured, and those those assumptions don't seem to hold for the new RDL as well.

So, I started a (mostly) from-scratch rewrite of the entire parser using `winnow`, which allows us to implement a parser for the entire SystemRDL 2.0 grammar. The grammar is virtually identical to what appears in the [SystemRDL 2.0
specification](https://www.accellera.org/images/downloads/standards/systemrdl/SystemRDL_2.0_Jan2018.pdf), other than changes to make it parseable in an LL-style that `winnow` supports (notably, I removed some left recursion manually).

This keeps the existing lexer code and expands it to support all of the keywords in the spec.

I have already started working on a new generator using this library, but I will check that in a separate PR.

(`ast.rs` and `parser.rs` are the new files, the rest are copied with some slight modification.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2223,6 +2223,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "mcu-registers-systemrdl-new"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "same-file",
+ "winnow 0.7.4",
+]
+
+[[package]]
 name = "mcu-rom-common"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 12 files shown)


helloxiling (approved):
Looks good to me.

swenson:
Thanks!

PR #152, opened 2025-04-01
Downgrade cmake to 3.x so that SPDM validator can run

swenson:
The SPDM validator's file were built with an old version of CMake (<= 3.5). CMake (in 4.0.0) just dropped support for those build scripts.

So, we download and install the last version of CMake 3.x that supports the SPVM validator build scripts as part of our CI pipeline.


parvathib (approved):
LGTM

swenson:
Thanks!

PR #153, opened 2025-04-04
Update Image Loading Doc

mlvisaya:
* Use Image Loading API as the interface for streaming boot instead of another streaming boot API

* Add details on the PLDM handshake for streaming boot

* Allow user to select where to load the images from (i.e. from flash or from Pldm) rather than getting it from the handoff table provided by the MCU ROM

* Add the concept of Image Component and Location. An Image component (e.g. firmware) can be written to multiple locations idenfied by the load_address. In this case, only 1 Image Component needs to be stored in the flash or the PLDM FW Package. The SOC Manifest should contain a mapping of the location to the component.

* Add a mailbox command to get the Image Info for a component


helloxiling (approved):
Looks good to me.

mlvisaya:
Thank you!

PR #155, opened 2025-04-08
Fix intermittent failures in MCTP message protocol tests.

parvathib:
The PR addresses the following to fix intermittent failures in MCTP message protocol tests
* Adds retry mechanism for SPDM commands
* Ensure main thread waits for tests to complete
* Make the prints in SPDM validator integration tests uniform


swenson (approved):
LGTM, but the tests still failed due to timeouts it looks like.

PR #165, opened 2025-04-15
Add Caliptra crypto API support for SHA

parvathib:
This PR includes the following updates:
1. Adds support for the Caliptra Crypto API for SHA.
2. Refactors the GET_DIGESTS and GET_CERTS SPDM commands to use the crypto API for digest calculations.
3. Adds an argument to the run_test! macro to enable example_app for testing. The example app and other production apps (spdm, pldm) run exclusively from each other. The test must now specify which to run.
4. Increases the test timeout and retry count for crypto API latency.


builder/src/apps.rs
```diff
@@ -49,16 +50,31 @@ pub const BASE_PERMISSIONS: &[(u32, u32)] = &[
 ];
 
 // creates a single flat binary with all the apps built with TBF headers
-pub fn apps_build_flat_tbf(start: usize, ram_start: usize, features: &[&str]) -> Result<Vec<u8>> {
+pub fn apps_build_flat_tbf(
```
swenson, line 53: Comment updated to reflect that it is now just the example app or all of the non-example apps?

swenson (approved):
LGTM

builder/src/apps.rs
parvathib: updated.

PR #169, opened 2025-04-23
Restructure the MCU Runtime firmware folders

parvathib:
Restructures the runtime firmware components without changing the contents and losing the git history.


swenson (approved):
LGTM

I'm surprised that so few changes had to be made to move all the files!

PR #172, opened 2025-04-25
Make Transport genertic and remove thiserror dependency in SPDM

parvathib:
This PR 
1. Makes MCTP transport a trait implementation
2. Removes thiserror crate and uses enums to define errors in SPDM


swenson (approved):
LGTM

PR #176, opened 2025-04-30
Caliptra subsystem external command spec (draft)

helloxiling:


helloxiling:
The draft spec for the external mailbox command is available at: https://github.com/chipsalliance/caliptra-mcu-sw/pull/189. I will create a separate document to describe the customized messages that leverage these mailbox commands.

PR #179, opened 2025-05-02
Remove unnecessary MCU mailbox API

mlvisaya:
Remove the redundant MCU mailbox API. Use the imported caliptra-sw's mailbox API instead since it offers the latest mailbox commands and it is not necessary to duplicate efforts in the MCU.


helloxiling (approved):

mlvisaya:
thanks!

PR #183, opened 2025-05-07
Additional Integration test cases for flash and streaming boot

mlvisaya:
* Add test cases to simulate invalid configurations

- incorrect image id
- unauthorized fw data
- incorrect load address
- incorrect PLDM descriptor
- incorrect PLDM component ID
- corrupted PLDM package
- PLDM package has lower version than device's

* Have the PLDM UA emulator exit with an error code when it detects an
error. This allows the integ test to check for errors.

* Fixed an issue in the image loader PLDM handle_component ops where it
returns with an Err() instead with an Ok(response_code).

* Fixed the issue that sometimes emulator exits early
This was caused by the MCU emulator throwing a Memory Load Access Fault
when handling external interrupt. In some cases, the mode is still in
User mode before accessing the interrupt vector table causing an access fault.
To fix the issue, the mode is elevated to Machine mode first before accessing
the interrupt vector table when handling external interrupts.

### DIFF
--- a/emulator/cpu/src/cpu.rs
+++ b/emulator/cpu/src/cpu.rs
@@ -788,6 +788,21 @@ impl<TBus: Bus> Cpu<TBus> {
             const NON_RAM_NMI: u32 = 0xF000_1002;
             return self.handle_nmi(NON_RAM_NMI, 0);
         }
+        // Elevate to M Mode before reading the vector table
+        let mut status = match self.read_csr_machine(Csr::MSTATUS) {
+            Ok(value) => RvMStatus(value),
+            Err(_) => return StepAction::Fatal,
+        };
+        match self.priv_mode {
+            RvPrivMode::U => {
+                self.priv_mode = RvPrivMode::M;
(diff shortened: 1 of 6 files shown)


helloxiling (approved):
Looks good to me.  
Nice work to root cause and fix the issue in cpu emulation.

emulator/cpu/src/cpu.rs
```diff
@@ -788,6 +788,21 @@ impl<TBus: Bus> Cpu<TBus> {
             const NON_RAM_NMI: u32 = 0xF000_1002;
             return self.handle_nmi(NON_RAM_NMI, 0);
         }
+        // Elevate to M Mode before reading the vector table
+        let mut status = match self.read_csr_machine(Csr::MSTATUS) {
```
swenson, line 792: Great catch!

swenson (approved):
LGTM

parvathib (approved):

mlvisaya:
thanks all!

PR #186, opened 2025-05-08
registers: Support generating emulator peripherals for array registers

swenson:
This computes the index into the (flattened) register array and passes it as an argument to the bus peripheral trait.

This doesn't affect any current peripherals but will be necessary for the watchdog timers in the MCI peripheral.


parvathib (approved):
Looks pretty straightforward.

swenson:
Thanks!

PR #188, opened 2025-05-09
update subscribe calls and move apps to platform specific folder

parvathib:
The PR:
1. Updates the subscribe calls to return TockSubscribe future to be able to cancel the future in case of error.
2. Move apps to the platform specific folder


swenson (approved):
LGTM

PR #190, opened 2025-05-15
Fix PLDM encoding of GetFirmwareParameters command

mlvisaya:
GetFirmwareParameter command's active version string has a dynamic length and should be encoded/decoded manually. Currently, it's inside the fixed parameters struct which is encoded/decoded by zerocopy. As a fix, this string is moved out of the fixed parameters struct.

Also fixed the xtask command to decode the PLDM FW package, it was calling the incorrect function to decode the FW package. It was calling to decode a manifest file, but the function is for decoding the firmware package


helloxiling (approved):
Good catch! Looks good to me.

mlvisaya:
thanks

PR #191, opened 2025-05-16
emulator: Add offsets to the root buses

swenson:
So that the emulator can be more flexible about addresses things are mounted at.


mlvisaya (approved):
lgtm

swenson:
Thanks. There's some flaws with this approach, but it's a good start. I have some fixes in the next PR.

PR #197, opened 2025-05-19
rom/runtime: Add flexible memory map

swenson:
That is passed in at build time for the ROM and runtime.

I specifically did not include flexible addresses for emulator-only peripherals, like our fake flash controller.


helloxiling (approved):
Looks good to me.

swenson:
Thanks!

PR #201, opened 2025-05-23
Remove misplaced test file

mlvisaya:
pldm_fw_update_test.rs is duplicated in
platforms/emulator/runtime/src/tests directory but it's already in emulator/app/src/tests which is the correct location.

Delete the one in platforms/emulator/runtime/src/tests


swenson (approved):
LGTM

PR #209, opened 2025-05-31
hw/model: rom_build now requires the feature argument

swenson:
I merged #202 without rebasing after #208, which added a feature argument to the rom_build.


swenson:
Thanks! Test failures are unrelated.

PR #211, opened 2025-06-02
hw/model: FPGA now mostly boots Caliptra

swenson:
There's a problem with the MCU -> Caliptra mailbox interface that we are debugging, but this gets through most of the boot process.

I updated `caliptra-sw` to point to the latest commit and did some related cleanup (new Rust version so some clippy fixes were necessary).

I also did some more cleanup of the FPGA model to use Tock registers that were automatically generated from RDL and then did some manual tuning.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,19 +102,20 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.6"
+version = "3.0.8"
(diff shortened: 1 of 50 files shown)


swenson:
Test failures are unrelated.

parvathib (approved):

swenson:
Thanks!

PR #214, opened 2025-06-05
Update Recovery Boot Flow Doc

mlvisaya:
Since MCU can not reset Caliptra core in the event of boot failures, the Caliptra core will just indicate the failure to the MCU. The MCU will then switch the active partition to the previous working one and set the FW_ERROR_FATAL bit. This bit will be an indication to the platform to reset the full SOC.

The MCU will no longer initiate reset and will leave it up to the integrator to determine the best time to do so.


docs/src/image_loading.md
```diff
@@ -319,16 +319,15 @@ The method of selecting which partition to boot from is system-specific. For exa
 ###  Caliptra Boot Flow
 This section describes the actions taken by the Caliptra ROM and Caliptra FMC/Runtime (RT) during boot, particularly in handling boot failures.
 
-If a corrupted or unauthorized Caliptra FMC+RT image is loaded, the Caliptra ROM asserts an error signal to the MCU, which initiates a recovery mechanism to provide the Caliptra core with a valid firmware image.
+If a corrupted or unauthorized Caliptra FMC+RT image is loaded, the Caliptra ROM asserts an error signal to the MCU, which initiates a recovery mechanism to provide the Caliptra core with a valid firmware image on the next SOC reset.
```
mhatrevi, line 322: Reword this to "Caliptra ROM sets the _CPTRA_FW_ERROR_FATAL_ register. MCU then initiates a recovery mechanism to provide the Caliptra core with a valid firmware image on the next SOC reset."

mhatrevi (approved):

docs/src/image_loading.md
mlvisaya: will do.

PR #216, opened 2025-06-06
Add support for GET_MEASUREMENTS

parvathib:
This PR includes the following changes:
1. Adds support for the GET_MEASUREMENTS command. The PCR Quote from Caliptra, with ECC signature, is provided as a free-form manifest measurement block.
2. Adds MCTP retries to address dropped commands if the application has not set the syscalls before the command arrived.

Note: This also adds support for the CHUNK_GET command for transmitting large responses, although it has not been tested yet.
The streaming boot tests are temporarily commented out until the SHA accelerator data size limit issue is resolved.


swenson (approved):
LGTM

PR #217, opened 2025-06-09
Avoid unnecessary rebuilds; cache Caliptra ROM/FW.

swenson:
We avoid unnecessary rebuilds of the kernel and applications if the offsets and sizes have not changed or if we can guess the sizes correctly the first time. We cache the sizes in the `target` directory to make our guesses better.

If the Caliptra is specified as a git dependency, then we cache the ROM and runtime, since they are fixed. (If Caliptra is a local dependency, we do not cache the ROM and runtime.)

This should reduce the build and test time by ~50%, if not more.

This also ignores the SPDM tests when the validator is not installed, so that tests won't fail locally.


swenson:
For me, locally:

Before:
```
cargo t  1157.56s user 132.81s system 102% cpu 20:54.28 total
```

After:
```
cargo t  963.44s user 117.58s system 103% cpu 17:29.37 total
```

Speedup: ~17% wall clock time speedup, ~16% decreased CPU usage

helloxiling (approved):
Looks good to me.

PR #218, opened 2025-06-10
Load images from Active Partition

mlvisaya:
- Define the struct for the Partition Table
- Add a new partition in the primary flash for the partition table
- Add arguments to the emulator to pass contents of the primary and secondary flash.
- Add arguments flash_image creator scripts to include the partition table
- In the image-loader app, query the flash partition table to find out the active partition, the image-loader app will then load the image from the active partition. It will update the boot status to Boot Successful if loading succeeds.
- Due to lack of space in the flash and RAM, some cleanups were made to reduce App size. The Flash and PLDM ImageLoader APIs have been separated since, ideally, only 1 type is supported at a time. In this way, if either one is not used, then the corresponding code will not be linked and will help reduce the size of the App


platforms/emulator/config/src/flash.rs
```diff
@@ -1,32 +1,51 @@
+use core::mem::offset_of;
```
helloxiling, line 1: Please place it under the license header.

PR #213, opened 2025-06-05
Backport caliptra-sw CPU to MCU

Elizafox:
This pull request brings minor enhancements from caliptra-sw to MCU.

This also minimises the differences between the two CPU implementations.

PR #219, opened 2025-06-11
Use DCCM for PIC vector table

swenson:
This is required by the RTL. (We had incorrectly been using SRAM, which is not currently allowed by the RTL.)

Also includes #213


emulator/app/src/main.rs
```diff
@@ -21,16 +21,17 @@ mod mctp_transport;
...
 use caliptra_emu_periph::CaliptraRootBus as CaliptraMainRootBus;
 use clap::{ArgAction, Parser};
 use clap_num::maybe_hex;
 use crossterm::event::{Event, KeyCode, KeyEvent};
 use emulator_bmc::Bmc;
-use emulator_bus::{Bus, BusConverter, Clock, Timer};
 use emulator_caliptra::{start_caliptra, StartCaliptraArgs};
-use emulator_consts::{RAM_OFFSET, ROM_SIZE};
-use emulator_cpu::{Cpu, Pic, RvInstr, StepAction};
+use emulator_consts::DEFAULT_CPU_ARGS;
```
parvathib, line 33: nit: any reason why we are not combining these declarations (see some in other places too)

PR #221, opened 2025-06-11
Combine SPDM app and IMAGE loader App to one

mlvisaya:
- Combine SPDM app and IMAGE loader App to one into 1 user app
- The user app uses 1 executor that spawns the spdm and image loader tasks


mlvisaya:
thanks!

PR #226, opened 2025-06-13
Enable SOC BOOT Integration Tests

mlvisaya:
The SOC boot integration tests were previously disabled since the compiled user app with the soc boot features enabled will not fit RAM.

To be able to fit the App, the DATA_RAM_SIZE is decreased by 8KB.

With the Image loading task running the integ tests with the SPDM task, an issue with the mailbox syscall was encountered:

Error copying from mailbox: MailboxRespTypeTooSmall

It was found that there is a race condition between the Image Loading task and the SPDM Task.

Image Loading Task executes a mailbox command. After mailbox driver calls the upcall for the response of this command, the SPDM task immediately enqueues a command to the mailbox. However, the Image Loading Task will call:

        S::unallow_ro(self.driver_num, mailbox_ro_buffer::INPUT);
        S::unallow_rw(self.driver_num, mailbox_rw_buffer::RESPONSE);

before it exits the syscall. This will ungrant the kernel access to the rw_buffer. When the mailbox driver receives the response for the SPDM command, it will try to read an invalid rw_buffer with a length of 0. The driver will then throw the MailboxRespTypeTooSmall error since it can't copy the response to the buffer.

To fix this issue, an mutex is added to make sure only 1 task can execute a mailbox command at a time.

### DIFF
--- a/runtime/userspace/syscall/src/mailbox.rs
+++ b/runtime/userspace/syscall/src/mailbox.rs
@@ -95,9 +95,6 @@ impl<S: Syscalls> Mailbox<S> {
         })?
         .await;
 
-        S::unallow_ro(self.driver_num, mailbox_ro_buffer::INPUT);
-        S::unallow_rw(self.driver_num, mailbox_rw_buffer::RESPONSE);
-
         match result {
             Ok((bytes, error_code, _)) => {
                 if error_code != 0 {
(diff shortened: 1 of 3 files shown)


runtime/userspace/syscall/src/mailbox.rs
```diff
@@ -85,8 +85,6 @@ impl<S: Syscalls> Mailbox<S> {
             {
                 Ok(()) => Ok(TockSubscribe::subscribe_finish(sub)),
                 Err(err) => {
-                    S::unallow_ro(self.driver_num, mailbox_ro_buffer::INPUT);
```
parvathib, line 88: This should not be removed. It is necessary in case the command returns immediately due to an error. The buffers need to be unallowed in that case because there will be no upcall.

mlvisaya: ok. i'll put this back.

swenson (approved):
LGTM

mlvisaya:
thank you

Issue #210, opened 2025-06-02
Automatic `mrac` calculation from McuMemoryMap


swenson:
We need a function that is attached to the `McuMemoryMap` that takes the memory map and computes the `mrac` register value (a 32-bit value containing what 256 MB regions are cachable or have side effects, like MMIO).

[Example memory map](https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/platforms/fpga/config/src/lib.rs#L8)

[VeeR mrac docs](https://chipsalliance.github.io/Cores-VeeR-EL2/html/main/docs_rendered/html/memory-map.html#region-access-control-register-mrac)

[Example mrac setting code](https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/rom/dev/src/start.S#L54-L78)

I would imagine a solution for this would involve sorting the offsets/lengths, splitting them if they cross a region (unlikely), and then AND/ORing together properties, with a default of don't cache + side effects possible, to compute the value.

The value should be somehow injected into either the platform-specific ROM code or the `rom/src/riscv.rs` code -- perhaps as a `extern` constant, injected into the linker script as an address, autogenerated into a `.rs` or `.s` file, or a combination.

swenson:
Done in #228

PR #228, opened 2025-06-15
Implement MRAC register initialization for memory region caching (#210)

raghuncstate:
This commit implements automatic MRAC (Region Access Control) register initialization in the Caliptra MCU ROM startup sequence. The MRAC CSR (0x7c0) controls memory access properties for 16 regions of 256MB each, with 2-bit encoding per region for [side_effect, cacheable] properties.

Key changes:
* Added MemoryRegionType struct with side_effect/cacheable booleans
* Updated McuMemoryMap to use embedded MemoryRegionType properties
* Implemented compute_mrac() method with automatic region classification
* Enhanced ROM linker template to inject computed MRAC_VALUE
* Modified startup assembly (start.s) to initialize MRAC register
* Added bounds checking and conflict detection for memory regions

Platform-specific MRAC values generated:
* Emulator: 0xaaa9a5aa (accommodates different memory layout)
* FPGA: 0xaa6aa6aa (matches FPGA-specific memory configuration)

This enables proper memory caching behavior and MMIO side-effect handling across different platform configurations while maintaining build-time flexibility.


platforms/emulator/rom/src/start.s
```diff
@@ -26,6 +26,15 @@ _start:
     # Initialize the stack pointer
     la sp, STACK_TOP
 
+    # Initialize MRAC (Region Access Control Register)
+    # MRAC controls cacheability and side effects for 16 memory regions (256MB each)
+    # The value is computed from the memory map at build time
+    # CSR address 0x7c0 = MRAC register
+    # Use lui/addi to load the 32-bit constant properly
+    lui     t0, %hi(MRAC_VALUE)
```
raghuncstate, line 34: i think LA does the same thing but wasnt sure. Can replace with LA if equivalent.

swenson: Either is fine. I usually use `li` or `la` because I never remember the `%hi` syntax :)

swenson (approved):
LGTM. Thanks! I appreciate the thorough tests.

PR #230, opened 2025-06-19
Support building runtime firmware with stack in DCCM

swenson:
And a larger DCCM.

This also cleans up some manual fixes in the autogenerated root bus.


helloxiling (approved):
Looks good to me.

PR #235, opened 2025-06-24
I3C and build fixes necessary for FPGA running full recovery mode

swenson:
This does some cleanup in the ROM and I3C code to add the necessary initialization for recovery, and adds in some additional parameters (`McuStraps`) that can be passed into the ROM and runtime so the ROM can support more customization (like setting AXI users and watchdog timeouts).

This brings in all of the changes necessary for the FPGA to run the full recovery flow as simply as:

```
cargo xtask fpga-install-kernel-modules && cargo t -p mcu-hw-model --features fpga_realtime -- --test model_fpga_realtime::test::test_new_unbooted --exact --nocapture
```

Apologies for the mess in `hw/model` -- I'm still cleaning that up.


rom/src/i3c.rs
```diff
@@ -0,0 +1,169 @@
...
+    registers: StaticRef<i3c::regs::I3c>,
+}
+
+impl I3c {
+    pub const fn new(registers: StaticRef<i3c::regs::I3c>) -> Self {
+        I3c { registers }
+    }
+
+    /// Run the initialization steps for the primary and secondary controller.
+    pub fn configure(&mut self, addr: u8, recovery_enabled: bool) {
```
helloxiling, line 23: It seems this configuration only takes effect on actual hardware, as our emulated I3C peripheral does not fully implement the `pub trait I3cPeripheral` defined in `registers/generated-emulator/src/i3c.rs`. Is there any side effect while running MCU ROM on emulator?

swenson: Nope, it should be safe on the emulator, since most of these writes will be ignored. The `I3cPeriperhal` trait by default will ignore all writes and return `0` for any reads.

PR #237, opened 2025-06-24
Calculate PMP Regions from McuMemoryMap

### LINKED ISSUE #236: Calculate PMP entries from McuMemoryMap
we need to set the PMP regions in the Tock kernel a little better. Currently we hardcode these regions, and have to manually specify if they are "napot" (naturally power-of-two aligned) or just a range. We generally prefer napot because they use only a single PMP register instead of two.
 
https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/platforms/fpga/runtime/src/board.rs#L263-L297
 
What we'd like is to have most of that mess automatically calculated from the same memory map struct. The kernel can do the computation, so it doesn't have to be done at build time, and it may make sense to hardcode a few entries.
caliptra-mcu-sw/platforms/fpga/runtime/src/board.rs at main · chipsalliance/caliptra-mcu-sw
Caliptra MCU Software. Contribute to chipsalliance/caliptra-mcu-sw development by creating an account on GitHub.
 
it's a little more complex because we might need to make a judgement call on whether to use napot or ranges, and modify our PMP code to be more flexible about regions

raghuncstate:
Fix #236

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2578,6 +2578,7 @@ dependencies = [
  "mcu-components",
  "mcu-config",
  "mcu-config-emulator",
+ "mcu-platforms-common",
  "mcu-tock-veer",
  "registers-generated",
  "riscv 0.1.0",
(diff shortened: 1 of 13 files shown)


swenson (approved):
LGTM. This is great work!

The test failure is unrelated and should be fixed in #239

PR #241, opened 2025-06-25
Set wdt timer in rom and disable before kernel main loop

ArthurHeymans:


rom/src/riscv.rs
```diff
@@ -190,6 +214,8 @@ pub fn rom_start() {
     soc.registers.cptra_wdt_cfg[0].set(100_000_000);
     soc.registers.cptra_wdt_cfg[1].set(100_000_000);
 
+    soc.configure_wdt(WDT1_TIMEOUT_CYCLES, WDT2_TIMEOUT_CYCLES);
```
swenson, line 217: You can probably delete the two lines above this.

platforms/fpga/runtime/src/board.rs
```diff
@@ -592,6 +596,13 @@ pub unsafe fn main() {
     if let Some(exit) = exit {
         crate::io::exit_emulator(exit);
     }
+
+    // Disable WDT1 before running the loop
+    let soc: StaticRef<soc::regs::Soc> =
```
swenson, line 601: I think we don't need to modify Caliptra's own watchdog timers here -- it should take care of disabling them itself.

swenson (commented):
This is good, but we also need to set our own watchdog timers (and these are the ones we should disable in the runtime before the main loop, not Caliptra's):

https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.mci_top.mci_reg.WDT_TIMER1_EN

ArthurHeymans:
> This is good, but we also need to set our own watchdog timers (and these are the ones we should disable in the runtime before the main loop, not Caliptra's):
> 
> https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.mci_top.mci_reg.WDT_TIMER1_EN

Thanks for clarifying there are multiple WDT :-)

PR #242, opened 2025-06-25
make DOE transport transmit to use u8 slice and receive to use u32 slice

parvathib:
DOE data object receive path operates on a u32 slice, as this may be required by the hardware interface implementing DoeTransport. 
The transmit path uses a u8 slice, since it directly handles the application-provided buffer passed to the driver. This buffer may need to be 4-byte aligned, depending on the underlying implementation. 

It is expected that the upper layer (DOE capsule) ensures the transmit buffer length is 4-byte aligned, although this check may also be performed by the DoeTransport instance.


swenson (approved):
LGTM

PR #246, opened 2025-06-27
Add DOE Capsule

parvathib:
This PR introduces the DOE capsule framework.
The capsule still needs to be instantiated and integrated into the platform.


runtime/kernel/capsules/src/doe/driver.rs
```diff
@@ -0,0 +1,352 @@
+// Licensed under the Apache-2.0 license
+use crate::doe::protocol::*;
```
helloxiling, line 2: nit: leave one line space from license header.

parvathib: I'll add in the follow up PR. Thanks for reviewing.

PR #247, opened 2025-06-28
Initial draft of in-field fuse programming spec

swenson:
This contains an outline of how we will provide APIs and tools to define the vendor-specific fuses and how to program them via our command APIs.

### DIFF
--- a/docs/src/external_mailbox_cmds.md
+++ b/docs/src/external_mailbox_cmds.md
@@ -28,6 +28,9 @@ These commands support a wide range of functionalities, including querying devic
     - Facilitate secure debugging in production environments
     - Ensure controlled access to debugging features
 
+- **In-Field Fuse Provisioning**
+    - See [fuses spec](fuses.md) for details.
+
 ## Mailbox Commands List
 
 | **Name**                          | **Command Code** | **Description**                                                                                     |
@@ -65,6 +68,9 @@ These commands support a wide range of functionalities, including querying devic
 | MC_MLDSA_SIGN                     | 0x4D4C_4D53 ("MMLS") | Requests to sign a SHA-384 digest with the DPE leaf certificate using MLDSA.                       |
 | MC_PRODUCTION_DEBUG_UNLOCK_REQ    | 0x4D44_5552 ("MDUR") | Requests debug unlock in a production environment.                                                 |
 | MC_PRODUCTION_DEBUG_UNLOCK_TOKEN  | 0x4D44_5554 ("MDUT") | Sends the debug unlock token.                                                                      |
+| MC_FUSE_READ                      | 0x4946_5052 ("IFPR") | See [fuses spec](fuses.md) for details |
+| MC_FUSE_WRITE                      | 0x4946_5057 ("IFPW") | See [fuses spec](fuses.md) for details |
+| MC_FUSE_LOCK_PARTITION             | 0x4946_504B ("IFPK") | See [fuses spec](fuses.md) for details |
 
 ## Command Format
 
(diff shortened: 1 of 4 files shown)


docs/src/fuses.md
```diff
@@ -0,0 +1,151 @@
...
+
+The commands are:
+
+* Write
+* Read
+* Lock partition
+
+### Authorization
+
+Only authorized SoC users shall call these commands. This can be authorized through the mechanism itself (i.e., a mailbox that only a trusted SoC component has access to) or cryptographically.
```
helloxiling, line 78: Does it mean cryptographic authorization isn't needed if those commands are coming from mailbox?

docs/src/external_mailbox_cmds.md
```diff
@@ -65,6 +68,9 @@ These commands support a wide range of functionalities, including querying devic
 | MC_MLDSA_SIGN                     | 0x4D4C_4D53 ("MMLS") | Requests to sign a SHA-384 digest with the DPE leaf certificate using MLDSA.                       |
 | MC_PRODUCTION_DEBUG_UNLOCK_REQ    | 0x4D44_5552 ("MDUR") | Requests debug unlock in a production environment.                                                 |
 | MC_PRODUCTION_DEBUG_UNLOCK_TOKEN  | 0x4D44_5554 ("MDUT") | Sends the debug unlock token.                                                                      |
+| MC_FUSE_READ                      | 0x4946_5052 ("IFPR") | See [fuses spec](fuses.md) for details |
```
helloxiling, line 71: Are we exposing a generic READ/WRITE/LOCK mailbox API for fuses, or are we aiming for more fine-grained, operation-specific commands? Just trying to understand the intended level of abstraction.

docs/src/fuses.md
swenson: If the mailbox is over a "trusted channel", then that is generally be okay from a FIPS perspective. Otherwise, there has to be identity-based authorization for FIPS 140-3 Level 3.

Ultimately it is up to the integrator and their auditor, but we can provide an authorization mechanism.

docs/src/fuses.md
```diff
@@ -0,0 +1,151 @@
...
+
+Some example cryptographic mechanisms that could be used:
+
+* A certificate provided to MCU during boot, which will be used to validate signatures on each message.
```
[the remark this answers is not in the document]
swenson: This is just an example. Token-based one is also a good option.

docs/src/external_mailbox_cmds.md
helloxiling: That makes sense. For defining fuse programming commands over the OOB interface (e.g., BMC via MCTP), we may lean toward an operation-specific approach based on practical use cases.

swenson: We could an option to automatically generate commands for certain fields. This is more ergonomic, certainly, but increases code size. It would then be up to the integrator if they wanted to enable and use those commands.

docs/src/fuses.md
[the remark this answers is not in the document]
helloxiling: For our reference implementation, it would be great to stay close to production flows where possible. I’ve worked with the BMC/HKMS team on token-based authorization before. We can connect with them and explore the best approach to bring into our stack.

docs/src/fuses.md
```diff
@@ -0,0 +1,151 @@
...
+We will provide a script, `cargo xtask fuse-autogen`, that can be used to process an `.hjson` file into:
+
+* Firmware definitions and code for the fuses and bits (Rust)
+* Documentation on the fuses and bits (Markdown)
+
+The command will use either a `--platform` flag to indicate the file is `platforms/{platform}/fuse_bits.hjson` or a `--file` to specify the file manually.
+
+## Fuse Provisioning Commands
+
+The fuse provisioning commands can be used over our [generic command](#todo) interface through MCTP, mailboxes, etc., or through commands in firmware manifest (except read operations).
```
helloxiling, line 68: nit:  does generic command refer to `fuse_api_cmd.md`?

swenson: Yes, let me fix that now :)

PR #248, opened 2025-06-28
Remove passive mode from the emulator

swenson:
Caliptra core is our ride or die: we boot together or not at all.

Since nearly all of MCU's functionalities depend on Caliptra, running with passive mode doesn't make sense.

(Passive mode was broken anyway.)

I also did a few small package optimizations and println cleanups


helloxiling (approved):
Looks good to me.

swenson:
Thanks!

PR #253, opened 2025-07-02
Update Flash Format

mlvisaya:
- Use the common flash-image format defined in the common library used for builder, instead of redefining it in the builder.

- Make the flash header format more extensible by adding a pointer to where the payload begins. This allows the header to be extended in the future

- Instead of the payload checksum, have a checksum per image to be added in the TOC. This allows per image update without needing to update the payload checksum.

- Update doc


RaunakGu:
thanks @mlvisaya, So quick to address it,

PR #254, opened 2025-07-03
Set firmware optimization level to 'z'

swenson:
This makes three changes:

* Sets the firmware optimization level to 'z': this is similar to 's' (optimize for space) but removes loop vectorization, which saves more space at the cost of slightly lower performance and slightly increased data RAM usage.
* Use the "test" profile for running the emulator. Previously, we were using `--release`, but after I changed the optimization level to 'z', I noticed the emulator performance dropped. We have a lot of specific level '3' optimizations turned on for the emulator "test" profile, and the "test" profile does not use LTO, so builds should be faster. A few adjustments to timeouts and retries were necessary after this change.
* Small optimization to use the RISC-V `clmul` instruction to calculate the CRC-8 in the MCTP PEC. This removes a loop and replaces it with a few assembly instructions, which will be faster and smaller.

Integration test time before: 1915.19s
Integration test time after: 1616.60s (-15.6%)

Kernel size before: 122212 bytes
Kernel size after: 114516 bytes (-6.3%)

App size before: 104316 bytes
App size after: 102804 bytes (-1.4%)


helloxiling (approved):
Looks good to me.

swenson:
Thanks!

PR #255, opened 2025-07-08
Set the watchdog timer higher for FPGA

swenson:
The FPGA now boots up to loading the user application and starting the SPDM library!

It fails after sending the first SHA call to Caliptra, which is probably related to setting up timers or interrupts. I'm still investigating this.

Also a few other small fixes I noticed when testing:

* Temporarily disable unaligned scalar access. This increases the binary size but is necessary since there is an overlap in the SRAM and the MCI registers (side effects required in MRAC). If side effects == 1, then unaligned access will cause the hardware to throw an exception. We can re-enable this optimization if we can configure the FPGA to have an additional SRAM access port added in a safer region of memory.
* Sorts the PMP regions so that kernel code and SRAM are first. Otherwise the MMIO region ends up coming first, which overlaps with the kernel code and SRAM after conversion to NAPOT MMIO regions. (This could also be fixed by modifying the PMP code to allow TOR regions for MMIO.)
* Set the MCU NMI vector (which the watchdog will jump to)
* Set the OTP / fuse AXI user (missed this when setting other AXI users)
* Fix the FPGA use of Caliptra builder
* Fix FPGA memory addresses in its board file
* Remove some noisy printlns and no longer correct comments
* Remove unused `kernel_layout.ld` (FPGA uses the emulator one).


mlvisaya (approved):
LGTM

swenson:
Thanks!

PR #258, opened 2025-07-09
[SPDM] Update MCU SPDM stack to support message handling through DOE and MCTP transports

parvathib:
This PR includes the following updates:
1. Add support for handling messages from DOE transport.
2. Implement two tasks to listen to SPDM messages from both MCTP and DOE transports simultaneously.
3. Refactor the Certificate store to be shared between MCTP-SPDM and DOE-SPDM responder tasks and extended it for slot management scenarios.


platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/cert_chain/leaf.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
+use embassy_sync::mutex::Mutex;
+use libapi_caliptra::certificate::CertContext;
+use spdm_lib::cert_store::{CertStoreError, CertStoreResult};
+use spdm_lib::protocol::{AsymAlgo, ECC_P384_SIGNATURE_SIZE, SHA384_HASH_SIZE};
+
+const DPE_LEAF_CERT_SIZE: usize = 2048; // Size of the DPE leaf certificate buffer.
+
+pub const DPE_LEAF_CERT_LABEL: [u8; SHA384_HASH_SIZE] = [
```
swenson, line 11: What is this label? Do we need to add a TODO or a note that this will need to be customized?

swenson (approved):
LGTM, one question

platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/cert_chain/leaf.rs
parvathib: We need to provide a label to generate a DPE key (leaf key) and certify it, which also signs the transcript hashes. That’s why we use the same label for certifying the key and for signing, so the generated keys match up. 
I’m still figuring out who sets up this label and where it’s configured. I’ll work on those details over the next few weeks as we iron out the OCP profile stuff.

mhatrevi (approved):

PR #260, opened 2025-07-09
Wait for MCU runtime to start before starting tests

swenson:
The timeout still applies, but this won't start using retries until at least MCU runtime has started running.

This should hopefully make tests a little less flaky.


parvathib (approved):
LGTM. SPDM validator tool works too.

PR #261, opened 2025-07-10
Firmware Update

mlvisaya:
Initial implementation of firmware update.

1. Created a new task that spawns a PLDM service that downloads the firmware update image to a staging area.
The staging area can be implemented by a trait and in this implementation the image is downloaded to an external SRAM.
For flash-based system it will downloaded to the inactive partition of flash. This will be on the next set of changes.

2. Currently this patch only supports update of Caliptra and Auth Manifest.

After the flash image is downloaded, Caliptra is updated using the Caliptra_fw_load command, while the Auth Manifest is updated using the set_auth_manifest command

3. Due to the size of the Caliptra image and the auth manifest, it needs to be sent in chunks through the mailbox. Therefore an API is added to the mailbox to allow sending mailbox request payload in chunks.


runtime/userspace/syscall/src/dma.rs
```diff
@@ -154,3 +154,9 @@ mod dma_ro_buffer {
 mod dma_subscribe {
     pub const XFER_DONE: u32 = 0;
 }
+
+pub fn mcu_sram_to_axi_address(addr: u32) -> u64 {
+    const MCU_SRAM_HI_OFFSET: u64 = 0x1000_0000;
```
swenson, line 159: Not in this PR necessarily, but we don't want to hardcode these addresses in a generic `runtime/` component -- this should be passed in from the platform-specific runtime.

runtime/userspace/api/caliptra-api/src/firmware_update/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        let (image_offset, image_len) = self
+            .get_image_toc(
+                flash_header.image_count as usize,
+                image_headers_offset,
+                CALIPTRA_FMC_RT_IDENTIFIER,
+            )
+            .await
+            .map_err(|_| ErrorCode::Fail)?;
+        self.update_caliptra(image_offset, image_len).await?;
+        self.wait_caliptra_rt_execution().await?;
```
swenson, line 83: Maybe not in this PR, but we'll want to probably restart the user apps when updating, since Caliptra is going to reboot.

Caliptra updates mean that certs, cryptographic mailbox keys, DOT keys, etc. will be invalid and need to be recreated.

cc @parvathib

runtime/userspace/api/caliptra-api/src/firmware_update/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        let mut req = MailboxReqHeader { chksum: 0 };
+        let req_data = req.as_mut_bytes();
+        self.mailbox
+            .populate_checksum(CommandId::FW_INFO.into(), req_data)
+            .unwrap();
+
+        let response_buffer = &mut [0u8; core::mem::size_of::<FwInfoResp>()];
+
+        // Wait indefinitely until Caliptra RT is ready
+        // Todo: Implement a timeout mechanism
```
swenson, line 199: Our watchdog timer perhaps. Though if we time out, we'll probably want to signal the SoC to restart everything so we can do the recover flow.

swenson (approved):
LGTM

swenson (approved):

mlvisaya:
Thanks, all. I will merge this for now, and continue addressing feedback on the next PR.

PR #262, opened 2025-07-14
Add support for flash-based logging capsule

helloxiling:


builder/src/runtime.rs
```diff
@@ -466,7 +483,10 @@ MEMORY
     prog (rx) : ORIGIN = $APPS_OFFSET, LENGTH = $APPS_SIZE
     ram (rwx) : ORIGIN = $DATA_RAM_OFFSET, LENGTH = $DATA_RAM_SIZE
     dccm (rw) : ORIGIN = $DCCM_OFFSET, LENGTH = $DCCM_SIZE
+    flash (r) : ORIGIN = $FLASH_OFFSET, LENGTH = $FLASH_SIZE
 }
 
+PAGE_SIZE = 256;
```
swenson, line 489: This might be good to pass in as a parameter to the builder, or part of the config for the platform.

platforms/common/src/lib.rs
```diff
@@ -4,3 +4,21 @@
 
 #[cfg(target_arch = "riscv32")]
 pub mod pmp_config;
+
+#[macro_export]
+macro_rules! read_volatile_at {
+    ($slice:expr, $index:expr) => {{
+        let ptr = unsafe { $slice.as_ptr().add($index) };
```
swenson, line 11: Ditto here about at least adding `assert!` to do a bounds check.

platforms/common/src/lib.rs
```diff
@@ -4,3 +4,21 @@
...
+        unsafe { core::ptr::read_volatile(ptr) }
+    }};
+}
+
+#[macro_export]
+macro_rules! read_volatile_slice {
+    ($slice:expr, $index:expr, $buf:expr) => {{
+        for (i, b) in $buf.iter_mut().enumerate() {
```
swenson, line 19: I think at the very least we should add `assert!` to do the bounds check, since we are bypassing the normal slice bounds checks. Otherwise, this macro is still unsafe.

swenson (commented):
Could of small comments, otherwise looks good

PR #267, opened 2025-07-17
Add a feature flag for firmware for 2.1 HW.

swenson:
This is an additive flag to turn on 2.1 features, particularly for ROM (I3C AXI recovery bypass) and eventually for SPDM (ML-KEM, external mu support for ML-DSA).

This adds support for switching versions when using `cargo xtask runtime`, which will ensure that the firmware is built with the `hw-2-1` flag if running the emulator for 2.1.


swenson:
Thanks!

PR #275, opened 2025-07-23
Set emulator exiter

swenson:
This is used by some tests, such as the flash ROM tests, to exit.

I don't know how tests were ever working without this :)


parvathib (approved):
LGTM

PR #284, opened 2025-07-24
Make SPDM DOE tests deterministic

swenson:
These now sleep a deterministic amount of time (measured by CPU cycles), so won't be dependent on the speed of the host computer.

This also means we can trim down the sleep times, which reduces how long this test takes by 50% (down from 2.5 hours to 1.2 hours or so).

Next we should do the same for all of the emulator tests that use `sleep`.


parvathib (approved):
LGTM

swenson:
Thanks!

PR #287, opened 2025-07-25
Implement PLDM Update Agent Retries

mlvisaya:
When Update Agent sends a packet, it will schedule a timer and wait for device to respond. If the timer expires, the packet will be resent.

Update agent will retry up to 3 times every 3 seconds. This is done on both discovery and firmware update state machines.

### DIFF
--- a/emulator/bmc/pldm-ua/src/daemon.rs
+++ b/emulator/bmc/pldm-ua/src/daemon.rs
@@ -175,6 +175,10 @@ impl<
                         }
                     }
                     PldmEvents::Stop => {
+                        let discovery_sm = &mut *discovery_sm.lock().unwrap();
+                        discovery_sm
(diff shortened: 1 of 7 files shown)


swenson (approved):
LGTM, but we should try to make these tests deterministic similar to #284 (once I get that working fully)

mlvisaya:
Thanks Chris, I'll update this PR once #284 is merged.

mlvisaya:
I found that the sleep_emulator ticks could not be used directly to replace the timers used in the PLDM update agent, because sleep cannot be canceled but PLDM needs to cancel timer when it receives a response. This needs some extra changes to use the emulator ticks sleep. For now, I'll merge this PR to unblock other PRs that are failing due to the MCTP: no pending rx operation failures.

PR #288, opened 2025-07-27
Use Reset Reasons in ROM

### LINKED ISSUE #115: Warm reset in ROM
From HW team:

> MCI's boot sequencer will be reset to IDLE and will go through LCC init and FC init. MCI has a "Warm reset" status bit for MCU. MCU ROM should use this indication to know MCU SRAM has not been cleared and it can jump to the exiting image after it has configured and locked content within MCI.

raghuncstate:
Fix #115


romtime/src/mci.rs
```diff
@@ -57,4 +81,46 @@ impl Mci {
...
+    /// Read the reset reason register value
+    pub fn reset_reason(&self) -> u32 {
+        self.registers.mci_reg_reset_reason.get()
+    }
+
+    /// Get the reset reason as an enum
+    pub fn reset_reason_enum(&self) -> McuResetReason {
+        let reason = self.reset_reason();
+
+        let warm_reset = (reason & WARM_RESET_BIT) != 0;
```
swenson, line 94: You should be able to use `.read()` to read these directly from the Tock register: https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/registers/generated-firmware/src/mci.rs#L123

raghuncstate: Done.

swenson (approved):
LGTM

PR #289, opened 2025-07-27
Relocate flash drivers/capsules under platforms/emulator and bug fix for error handling in flash capsule 

helloxiling:


swenson (approved):
LGTM

PR #290, opened 2025-07-28
runtime: Set timers correctly and use mcycle

swenson:
The timer counters reset to 0 when the timer reaches the bounds, so they can't pull double duty as cycle counters as well. Instead, we need to use `mcycle` and `mcycleh` CSRs to check the current time.

This also corrects the clock for the FPGA and sets up a pattern for integrators to specify the CPU clock speed.

I commented out the field entropy programming logic until after the relevant PR is merged in caliptra-sw (this was an accidental inclusion on my part -- I should not have merged the previous PR until the caliptra-sw PR was merged).


Cargo.toml
mlvisaya, line 1: Lgtm!

mlvisaya (approved):

PR #292, opened 2025-07-30
Add ROM boot status

swenson:
So that tests and integrators can tell where the ROM has gotten to.

I based this on Caliptra core: https://github.com/chipsalliance/caliptra-sw/blob/main-2.x/common/src/boot_status.rs


swenson:
Thanks!

PR #293, opened 2025-07-30
MCU and SOC FW Update

mlvisaya:
1. Added new params to builder to allow specifying image params for MCU (e.g. staging address, loading address, exec bit and image_id)

2. Increased DATA_RAM_SIZE from 112 KB to 115KB

3. Added support for MCI interrupts

4. Bumped PLDM transfer size from 180 to 500 bytes

5. Added MCI registers for MCU reset and interrupt

6. Added MCI capsule, syscall and interrupt handler

7. Added logic for MCU Update in firmware_update module After downloading full flash image:
- MCU image is copied to the DMA accessible staging area
- MCU sends ACTIVATE_FIRMWARE command
- Caliptra clears FW_EXEC_CTRL[2]
- MCI interrupts MCU, MCU checks interrupt reason and sees reset_request
- MCU request MCI to do reset
- MCI resets MCU
- MCU boots to ROM and checks RESET_REASON
- MCU ROM runs Hitless Update Flow
- Caliptra loads MCU image from staging to MCU SRAM
- MCU boots with new RT image
- On boot, image loading module loads SOC images from staging area
- If everything boots up ok, MCU marks flash partition as active

8. Added example to make flash as staging area rather than the external test sram

9. Moved the firmware-update task to image_loading since image_loading and firmware_update can not happen at the same time. Only after image_loading finishes, the firmware update task is run.

10. Added Hitless Update flow in ROM


runtime/userspace/api/caliptra-api/src/image_loading/mod.rs
```diff
@@ -20,6 +20,7 @@ use libsyscall_caliptra::mailbox::MailboxError;
 use libsyscall_caliptra::{dma::AXIAddr, mailbox::Mailbox};
 use libtock_platform::ErrorCode;
 use libtockasync::TockExecutor;
+use mcu_config_emulator::dma::caliptra_axi_addr_to_dma_addr;
```
swenson, line 23: Not in this PR, but we should pass this in as a map or a function to the runtime, rather than having a dependency on the runtime to the emulator config.

runtime/kernel/drivers/i3c/src/core.rs
```diff
@@ -16,7 +16,7 @@ use registers_generated::i3c::regs::I3c;
 use tock_registers::{register_bitfields, LocalRegisterCopy};
 
 pub const MDB_PENDING_READ_MCTP: u8 = 0xae;
-pub const MAX_READ_WRITE_SIZE: usize = 250;
+pub const MAX_READ_WRITE_SIZE: usize = 512;
```
swenson, line 19: I think this number is determined by hardware: https://github.com/chipsalliance/i3c-core/tree/main/src/rdl#queue_size-register

The maximum size is 2^(5+1) * 4 = 256. I think for safety we should limit it to 252.

romtime/src/mci.rs
```diff
@@ -128,4 +128,30 @@ impl Mci {
...
+    }
+
+    pub fn read_wdt_timer1_en(&self) -> u32 {
+        self.registers.mci_reg_wdt_timer1_en.get()
+    }
+    pub fn write_wdt_timer1_en(&self, value: u32) {
+        self.registers.mci_reg_wdt_timer1_en.set(value);
+    }
+
+    pub fn handle_interrupt(&self) {
```
swenson, line 147: It might be good to add a comment to this function (or possibly rename the function) to indicate that this will clear the interrupt and request a reset.

runtime/userspace/api/caliptra-api/src/image_loading/mod.rs
mlvisaya: yeah, i think that's a good idea. i'll see if I can pass a function or a map.

runtime/kernel/drivers/i3c/src/core.rs
mlvisaya: i see. looks like there's no way to increase this size. i'll find another way to speed up packet transfer. i'll revert this change.

romtime/src/mci.rs
mlvisaya: btw, I separated the MCI changes to this PR: https://github.com/chipsalliance/caliptra-mcu-sw/pull/297. I'll address this feedback there. This is to unblock other changes that are dependent on MCI interrupt.

swenson (approved):
LGTM

mlvisaya:
Thanks!

PR #294, opened 2025-07-30
Increase User RAM size

mlvisaya:
https://github.com/chipsalliance/caliptra-mcu-sw/pull/266 is failing due to User app not fitting RAM.

Increase RAM size to 76KB.
DATA_RAM_SIZE also needs to be increased.


swenson (approved):
LGTM

PR #297, opened 2025-07-31
Support for handling MCI interrupts, and MCI app interface

mlvisaya:
* Added support for MCI interrupts
* Added MCI registers for MCU reset and interrupt
* Added MCI capsule, syscall and interrupt handler


runtime/kernel/capsules/src/mci.rs
```diff
@@ -0,0 +1,113 @@
+// Licensed under the Apache-2.0 license
+
+//! This provides the MCI capsule that calls the underlying MCI driver
+
+use kernel::grant::{AllowRoCount, AllowRwCount, Grant, UpcallCount};
+use kernel::syscall::{CommandReturn, SyscallDriver};
+use kernel::{ErrorCode, ProcessId};
+
+/// The driver number for Caliptra mailbox commands.
```
helloxiling, line 9: nit:  this is the driver number for MCI regs, right?

helloxiling (approved):
Looks good to me.

runtime/kernel/capsules/src/mci.rs
mlvisaya: oh yeah, i'll update this comment

PR #307, opened 2025-08-02
Add a first test similar to Caliptra integration tests.

zhalvorsen:
This adds a simple test to make sure the Caliptra registers are accessible from the MCU hardware model.


registers/generated-emulator/src/mci.rs
```diff
@@ -48,7 +48,7 @@ pub trait MciPeripheral {
         &mut self,
     ) -> caliptra_emu_bus::ReadWriteRegister<u32, registers_generated::mci::bits::HwRevId::Register>
     {
-        caliptra_emu_bus::ReadWriteRegister::new(0)
+        caliptra_emu_bus::ReadWriteRegister::new(0x1000)
```
zhalvorsen, line 51: This is a generated file. Is there a better mechanism to update this value?

swenson: The better way is to overwrite this method in the peripheral that implements this trait.

swenson (approved):
LGTM

PR #308, opened 2025-08-02
fuses: Use lifecycle partition to manage security state

swenson:
And write Vendor PK hash to fuses in FPGA tests instead of hard-coding in the ROM.

This adds in the ability to pre-provision the OTP memory with the correct bits for the various lifecycle states, tokens, and vendor PK hash, and to compute the cSHAK128 hashes and PRESENT digests where appropriate.

The values in the tests were pulled from the OTP memory of the FPGA to ensure we are doing the calculations correctly.

Still TODO: update the emulator for the lifecycle controller and OTP to handle these values as well.


clundin25 (approved):

swenson:
Thanks!

PR #309, opened 2025-08-04
[SPDM] Add FINISH command support with Handshake in the clear mode

parvathib:
Adds the FINISH implementation with Handshake in clear mode.
I'll add a follow-up PR for encrypted and authenticated handshake and application phases.

### DIFF
--- a/runtime/userspace/api/spdm-lib/src/commands/finish_rsp.rs
+++ b/runtime/userspace/api/spdm-lib/src/commands/finish_rsp.rs
@@ -2,20 +2,19 @@
 
 #![allow(dead_code)]
 
-use crate::codec::{encode_u8_slice, Codec, CommonCodec, MessageBuf};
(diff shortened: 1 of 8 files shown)


runtime/userspace/api/spdm-lib/src/commands/finish_rsp.rs
```diff
@@ -53,13 +50,42 @@ impl FinishRspBase {
...
+        .session_mgr
+        .session_info_mut(session_id)
+        .map_err(|e| (false, CommandError::Session(e)))?;
+
+    let computed_hmac = session_info
+        .compute_hmac(SessionKeyType::RequestFinishedKey, &th1_transcript_hash)
+        .await
+        .map_err(|e| (false, CommandError::Session(e)))?;
+
+    if computed_hmac != *requester_verify_data {
```
swenson, line 79: For hashes and HMACs we should use a constant-time comparison function, like the `constant_time_eq` crate.

swenson (approved):
LGTM, one comment

PR #314, opened 2025-08-04
FPGA: Add new_unbooted test case

clundin25:
This is a basic test case that boots the FPGA and checks the MCI boot
status.

This change also removes "fpga_realtime" as an enabled dependency in the xtask crate.

Since the dependency features of a workspace are resolved via a union,
this enabled "fpga_realtime" for the rest of the workspace.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,6 +2,7 @@
 
 [alias]
 xtask = "run --package xtask --"
+xtask-fpga = "run --package xtask --features=fpga_realtime --"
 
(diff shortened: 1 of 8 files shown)


clundin25:
@swenson This adds a test harness for the FPGA.

Currently tests are allowlisted, so we need to opt them in when they are ready.

Future work:

- Improve caching of binaries.
- Improve test harness to boot FPGA further.

swenson (approved):
LGTM

I like the extra cargo alias and using feature flags for the FPGA stuff.

PR #317, opened 2025-08-05
Set PQC type from fuses

swenson:
This is necessary to run Caliptra tests.

### DIFF
--- a/emulator/app/src/emulator.rs
+++ b/emulator/app/src/emulator.rs
@@ -63,6 +63,17 @@ pub type ExternalWriteCallback = Box<
     ) -> bool,
 >;
 
+fn parse_vendor_pqc_type(s: &str) -> Result<FwVerificationPqcKeyType, String> {
+    match s.to_lowercase().trim() {
+        "mldsa" => Ok(FwVerificationPqcKeyType::MLDSA),
+        "lms" => Ok(FwVerificationPqcKeyType::LMS),
+        _ => Err(format!(
+            "Invalid vendor PQC type: {}. Supported types are 'mldsa' and 'lms'.",
+            s
+        )),
+    }
+}
+
 #[derive(Parser, Debug, Clone)]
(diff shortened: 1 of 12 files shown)


emulator/app/src/emulator.rs
```diff
@@ -121,6 +122,10 @@ pub struct EmulatorArgs {
     #[arg(long)]
     pub owner_pk_hash: Option<String>,
 
+    /// 1 = MLDSA, 3 = LMS
+    #[arg(long, default_value_t = 3)]
```
mlvisaya, line 126: nit: maybe we can pass the "MLDSA" or "LMS" string instead of integer 1 and 3 so that we don't have to remember the enum mapping.

i.e. instead of "--vendor_pqc_type 1", we can pass "--vendor_pqc_type LMS"

this can be changed to:
    /// PQC key type: "MLDSA" or "LMS"
    #[arg(long, value_parser = parse_vendor_pqc_type, default_value = "LMS")]
    pub vendor_pqc_type: u8,

and then adding a custom parser:

/// Parse the vendor PQC key type from a string ("LMS" or "MLDSA") to a u8 value.
fn parse_vendor_pqc_type(s: &str) -> Result<u8, String> {
    match s.to_ascii_uppercase().as_str() {
        "LMS" => Ok(FwVerificationPqcKeyType::LMS as u8),
        "MLDSA" => Ok(FwVerificationPqcKeyType::MLDSA as u8),
        _ => Err(format!("Invalid PQC key type: {}", s)),
    }
}

mlvisaya (approved):
LGTM. just a minor suggestion but it's all good without changing it.

clundin25 (approved):

emulator/app/src/emulator.rs
swenson: Good call. I'll do that.

swenson:
Thanks!

swenson:
Successfully tested on the FPGA as well.

PR #319, opened 2025-08-06
Cache FirmwareBinaries::from_env

clundin25:
This adds a static cache to `FirmwareBinaries::from_env`. This makes the
function performant to call multiple times.

I believe there is more cleanup / refactoring to the builder that needs to be done, but this should keep the FPGA tests from spending a lot of time de-compressing the zip file.


swenson (approved):
LGTM

PR #320, opened 2025-08-06
Correct VeeR timer calculation

swenson:
We were setting the timer bound to be the delta from the reference time instead of the current time, so if there was a large timer set previously, then the bound would be set to some potentially very high time.

This was indeed happening due to a bug in the UART timer, which is also fixed here. (It is not clear that we need the UART timer, but I left it in now and set it to a large, fixed value.)

This fixes some weird issues in tests where SHA commands (for example) would randomly take 10+ seconds to execute.


mlvisaya (approved):
LGTM

parvathib (approved):
LGTM

swenson:
Thanks!

PR #321, opened 2025-08-06
Add an MCU manager to the HW model.

zhalvorsen:
This will be used by integration tests to support accessing registers as if from the external SoC. This also makes it easy to make tests that work with both the emulator and the FPGA.

This can been tested locally and on the FPGA, tests will be added in a follow on PR.


swenson (approved):
LGTM

PR #322, opened 2025-08-07
[CRYPTO] Add support for AES_256_GCM cipher

parvathib:
Adds crypto API support for AES_256_GCM crypto on MCU
Also adds convenience functions to encrypt and decrypt data in one-shot


swenson (approved):
LGTM

PR #323, opened 2025-08-07
Don't use alarm at all in uart

swenson:
This was causing the tock$ prompt to keep being printed.

This alarm doesn't seem to be needed at all for us, so I just removed it altogether.


swenson:
Thanks!

swenson:
Closed in favor of #327

PR #324, opened 2025-08-07
Don't read secret fuses and don't read fuses with incorrect addresses

swenson:
The fuses with incorrect addresses are not used yet, so we comment it out until we update the fuse RTL for 2.0 final in a few days.


zhalvorsen (approved):

zhalvorsen (approved):

swenson:
Thanks!

PR #326, opened 2025-08-08
Firmware Update Image Verification

mlvisaya:
Verify the firmware package after downloading the firmware update package from the PLDM update agent and before activating the images.

Steps:

1. Download the firmware package to MCU
2. Parse the components of the package
3. Verify the SOC manifest through mailbox command (without setting it in Caliptra)
4. Verify the downloaded MCU image by getting sha384 of image and comparing it with downloaded Manifest's SHA.
4. Verify the downloaded SOC images by getting sha384 of images and comparing it with downloaded Manifest's SHAs.
5. Verify and Activate Caliptra fw using FIRMWARE_LOAD mbox command.
6. Activate MCU FW using Hitless Update Reset
7. After boot, set the downloaded manifest to Caliptra using SET_AUTH_MANIFEST mbox command
8. Use image loading to load the rest of the SOC images from the staging partition.

This change relies on the new VERIFY_MANIFEST mailbox command that allows verifying the SOC manifest without activating it in Caliptra.

Also fixed a bug in the ROM for FW Hitless Update flow, the flow didn't wait for Caliptra to set the EXEC_CTRL[2] before jumping to MCU RT.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,7 +268,7 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 [[package]]
 name = "caliptra-api"
 version = "0.1.0"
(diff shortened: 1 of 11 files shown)


mlvisaya:
Thanks!

PR #327, opened 2025-08-08
Eliminate recursion in flash tests

swenson:
The flash tests were written with recursive calls to `run()`, which
calls `kernel_loop()`, which calls `alarm()`, which calls operations,
which in turn call `run()`, etc.

This would overflow the stack.

I'm not sure why this wouldn't happen before the prompt fix (also
included); probably something to do with how the alarms were scheduled
if there was another alarm to be processed?

Includes #323


parvathib (approved):
LGTM

PR #328, opened 2025-08-08
[MCTP] Add buffering to the MCTP capsule to hold incoming messages until the app is ready.

parvathib:
Add buffering to the MCTP capsule to hold incoming messages until the app is ready.
Also make the delays in MCTP tests more deterministic.
MCTP-Caliptra and MCTP-Secure SPDM capsule instances are currently commented out to avoid exceeding memory limits.
 I'll optimize and update the memory map and enable (at least MCTP-Caliptra) in future PRs.


swenson (approved):
LGTM

PR #339, opened 2025-08-13
Add support for MCU mailbox driver

helloxiling:
This PR refines the existing mailbox HIL to align with MCU mailbox protocol, and adds the MCU mailbox driver with in-kernel testing.


helloxiling:
Thanks!

PR #341, opened 2025-08-15
Add emulator SOC requester MCU Mailbox integration test

mlvisaya:
Add a transport module that emulates a SOC MCU mailbox requester in the emulator.

Add integration test "test-mcu-mbox-soc-requester-loopback" where the emulator would send MCU mailbox messages to the device and device loopsback the request back to emulator. The emulator would check if the messages are received and contains the exact data.

Allow MCU Mailbox registers to be shared through several threads through Arc<Mutex>. This is because the integration test is running on a different thread with the MCU emulator thread, and they both access the MCU Mailbox registers


emulator/app/mcu-mbox/src/mcu_mailbox_transport.rs
```diff
@@ -0,0 +1,157 @@
+// Licensed under the Apache-2.0 license
+
+use emulator_periph::{MCU_MAILBOX0_SRAM_SIZE, McuMailbox0Internal};
+use registers_generated::mci::bits::MboxExecute;
+use tock_registers::interfaces::Readable;
+
+#[derive(Clone)]
+pub struct McuMailboxTransport {
+    mbox: McuMailbox0Internal,
```
helloxiling, line 9: We'd better use `McuMailbox0External`, that represents the view of MCU mailbox from SoC side.

platforms/emulator/runtime/src/tests/mcu_mbox_test.rs
```diff
@@ -218,6 +225,57 @@ impl<'a> EmulatedMbxSender<'a> {
     pub fn finish(&self) {
         self.regs.mcu_mbox0_csr_mbox_execute.set(0);
     }
+
+    pub fn send_response(&self, req_payload: &[u32]) {
```
helloxiling, line 229: The original test setup is to `send_response` within `handle_deferred_call`. We'd like to exercise the code path of MCU mbox driver `send_response` directly.  `poll_and_check_response` is the function to check test result from sender perspective.  It is a loop back test in kernel mode.

Is the new test case `test_mcu_mbox_soc_requester_loopback() ` working in a similar way? If yes, I would recommend reusing the test function.

emulator/app/mcu-mbox/src/mcu_mailbox_transport.rs
```diff
@@ -0,0 +1,157 @@
...
+        // Sender writes data to MBOX_SRAM
+        for (index, chunk) in payload.chunks(4).enumerate() {
+            let val = u32::from_le_bytes(chunk.try_into().unwrap_or([0; 4]));
+            self.mbox
+                .regs
+                .lock()
+                .unwrap()
+                .write_mcu_mbox0_csr_mbox_sram(val, index);
+        }
+        // Sender writes data length in words to MBOX_DLEN
```
helloxiling, line 37: Please update the comment. "// Sender writes data length in bytes to MBOX_DLEN".  The code is correct.

helloxiling (approved):
Looks good to me.

mlvisaya:
thanks for the review!

PR #347, opened 2025-08-19
Fix integration tests

parvathib:
correct the test_exit position to keep the app running


mlvisaya (approved):
LGTM. thanks for fixing it.

swenson (approved):
LGTM

PR #366, opened 2025-08-23
fpga: Reset AXI before and after test

swenson:
This makes the FPGA significantly more stable and should remove the need to rebooting the FPGA between runs occasionally.

### DIFF
--- a/hw/model/src/model_fpga_realtime.rs
+++ b/hw/model/src/model_fpga_realtime.rs
@@ -141,6 +141,12 @@ impl ModelFpgaRealtime {
                 .modify(Control::BootfsmBrkpoint::CLEAR);
         }
     }
+
+    fn axi_reset(&mut self) {
+        self.wrapper.regs().control.modify(Control::AxiReset.val(1));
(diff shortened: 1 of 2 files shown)


zhalvorsen (approved):

hw/model/src/model_fpga_realtime.rs
```diff
@@ -141,6 +141,12 @@ impl ModelFpgaRealtime {
                 .modify(Control::BootfsmBrkpoint::CLEAR);
         }
     }
+
+    fn axi_reset(&mut self) {
+        self.wrapper.regs().control.modify(Control::AxiReset.val(1));
+        self.wrapper.regs().control.modify(Control::AxiReset.val(0));
```
jlmahowa-amd, line 147: This bit auto-clears.
It has to because while the AXI reset is asserted the FPGA wrapper registers are not accessible, so there wouldn't be a way to clear it. I put a delay loop in the wrapper to assert for 16 clocks, and the reset block asserts the reset to the interconnects for 16 clocks.

PR #377, opened 2025-08-26
Pass SoC Manifest SVN fuse values to Caliptra

mlvisaya:
- MCU ROM retrieves SoC Manifest SVN and Max SVN values and passes it to Caliptra using SoC interface.
- Add logic in MCU ROM to monitor Caliptra for Fatal Errors while waiting for FW to be ready. Caliptra will signal a fatal error if the SVN check fails. In this way, MCU will propagate the fatal error to SoC.
- Added SoC Manifest SVN to the OTP emulator peripheral
- Add arguments to the emulator runner to pass SoC Manifest SVN
- Add integration test to verify SoC Manifest Verification

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,7 +268,7 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 [[package]]
 name = "caliptra-api"
 version = "0.1.0"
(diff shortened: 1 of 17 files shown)


swenson (approved):
LGTM; nicely done on integrating into OTP

mlvisaya:
Thanks! I also included the fix for the failing FPGA test by adding the pqc type and the vendor hash in the init and boot params.

PR #389, opened 2025-09-02
Allow Core Test ROM to continue with a GPIO is asserted.

zhalvorsen:
MCU ROM needs to load mutable firmware for many of the tests. This will allow the ROM used for testing Caliptra Core to continue if the test needs mutable firmware.


swenson (approved):
LGTM

PR #390, opened 2025-09-02
Add MCU SVN verification

mlvisaya:
Provide an example implementation of how SVN can be added to an MCU image

- Add a header that contains an SVN to MCU FW image. This package (header and FW binary) will be loaded in the MCU SRAM. MCU ROM will verify the SVN in the image header and jump to the offset where the FW binary is found.
- Use Vendor Hash Prod fuses to implement MCU SVNs
- Verify MCU SVN in ROM, user must implement the ImageVerifier trait that implements the SVN verification. An example implementation is provided in mcu_imager_verifier.rs


helloxiling (commented):
Overall, this PR is good. We also need to include SVN verification in runtime firmware update flow. That could be done in separate PRs.

helloxiling (approved):
Looks good to me.

PR #393, opened 2025-09-03
Support arbitrary-length payloads in the MCU mailbox userland API 

helloxiling:


mlvisaya (approved):
lgtm! i think making it serializable by letting app finish the transaction and buffering the message ensures the handling order of the mbox transactions.

helloxiling:
Thanks!

PR #401, opened 2025-09-04
Set AXI user to Core test default.

zhalvorsen:
Several core tests assume the AXI user is `0x1`. This sets the default in MCU FPGA user to that value. This makes several more tests pass.

At some point this should be communicated to the MCU ROM in a configurable way.


swenson (approved):
LGTM

PR #406, opened 2025-09-04
Fix FPGA HW Model being unable to run recovery flow.

zhalvorsen:
The default lifecycle state was being set to prod which later would later cause it to fail the Caliptra firmware integrity checks because the built image uses ML-DSA by default, but the hardware model uses LMS by default.


swenson (approved):
LGTM

PR #408, opened 2025-09-04
Add caliptra-sw argument to `xtask fpga built-test`.

zhalvorsen:
This optional argument makes it easier to test local caliptra-sw changes on the FPGA using xtask.

### DIFF
--- a/xtask/src/fpga.rs
+++ b/xtask/src/fpga.rs
@@ -70,6 +70,9 @@ pub(crate) enum Fpga {
         /// When copy test binaries to `target_host`
         #[arg(long)]
         target_host: Option<String>,
+        /// Local caliptra-sw path. Used in conjunction with the Cargo.toml change.
+        #[arg(long)]
+        caliptra_sw: Option<PathBuf>,
     },
     /// Run FPGA tests
     Test {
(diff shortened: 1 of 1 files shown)


xtask/src/fpga.rs
```diff
@@ -70,6 +70,9 @@ pub(crate) enum Fpga {
         /// When copy test binaries to `target_host`
         #[arg(long)]
         target_host: Option<String>,
+        /// Local caliptra-sw path. Used in conjunction with the Cargo.toml change.
+        #[arg(long)]
+        caliptra_sw: Option<PathBuf>,
```
clundin25, line 75: nit:

```suggestion
        caliptra_sw_path: Option<PathBuf>,
```

xtask/src/fpga.rs
```diff
@@ -231,18 +234,41 @@ pub(crate) fn fpga_entry(args: &Fpga) -> Result<()> {
...
+            ]);
+
+            // Add optional path to the caliptra-sw directory
+            if let Some(caliptra_sw) = caliptra_sw {
+                let basename = caliptra_sw.file_name().unwrap().to_str().unwrap();
+                cmd.arg(&format!(
+                    "-v{}:/{basename}",
+                    std::fs::canonicalize(&caliptra_sw)?.display()
+                ));
+            }
```
clundin25, line 265: If my local caliptra-sw is at `/home/clundin/code/sample` and I set `caliptra-sw` to `/home/clundin/code/sample`, will basename be `/home/clundin/code/sample`?

zhalvorsen: No it would be `sample`. I mostly tested it with a relative path.

My path in the Cargo.toml was `../2.x`. I used `--caliptra-sw ../2.x` when running the command. So the argument became `-v/usr/local/google/home/zhalvorsen/caliptra/2.x:/2.x`.

clundin25: Hmm maybe we should add a note to the doc string that it has to be exact match for the path in cargo.toml

PR #424, opened 2025-09-11
[ROM] Reserve bits in boot flow for milestones.

zhalvorsen:
This splits the MCI FW boot flow register into two pieces:

1. 16 checkpoint status bits (same as previous behavior)
2. 16 milestones bitflags

This is intended to make it easier to wait for certain milestones in the boot flow. With the FPGA we don't have strong control over timing so it could be easy to miss a checkpoint. These bitflags are a union of all the milestones seen ensuring they can't be missed.

The checkpoints are still very helpful for more fine grained boot flow debugging.


clundin25 (approved):

swenson (approved):
LGTM

PR #432, opened 2025-09-16
Replace Emulated DMA with AMD DMA driver

mlvisaya:
There is a need for an actual DMA implementation that can be supported in hardware.

For this change, the support for AMD DMA controller is added (see https://docs.amd.com/r/en-US/pg034-axi-cdma).

Only the simple transfer mode is supported.

The emulated DMA driver code is kept, but the MCU runtime is now using the AMD DMA controller. The emulator peripheral for AMD DMA is also created.


platforms/emulator/runtime/src/components/dma.rs
```diff
@@ -9,14 +9,14 @@ use kernel::component::Component;
 use kernel::create_capability;
 
 pub struct DmaComponent {
-    driver: &'static dma_driver::dma_ctrl::EmulatedDmaCtrl<'static>,
+    driver: &'static dma_driver::axicdma::AxiCDMA<'static>,
```
swenson, line 12: Can we delete the old driver?

swenson (approved):
LGTM

platforms/emulator/runtime/src/components/dma.rs
swenson: I saw the comment in the body of the PR that you mentioned you left the old one in. It's okay for now either way.

mlvisaya:
thanks!

PR #436, opened 2025-09-17
[test] add raw unlock LC transition FPGA test

timothytrippel:
This adds an E2E test that performs a raw unlock lifecycle transition via the LCC JTAG TAP. This partially addresses #233.

### DIFF
--- a/hw/model/src/lcc.rs
+++ b/hw/model/src/lcc.rs
@@ -1,0 +1,152 @@
+// Licensed under the Apache-2.0 license
+//
+// Derived from OpenTitan project with original copyright:
+//
+// Copyright lowRISC contributors (OpenTitan project).
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::iter;
+use std::time::Duration;
+
+use anyhow::{bail, Context, Result};
+use thiserror::Error;
+
+use caliptra_hw_model::lcc::{LcCtrlReg, LcCtrlStatus, LcCtrlTransitionCmd};
+use caliptra_hw_model::openocd::openocd_jtag_tap::{JtagTap, OpenOcdJtagTap};
+use mcu_rom_common::{Lifecycle, LifecycleControllerState};
+use poll_common::poll_until;
+
+/// Errors related to the LCC operations below.
+#[derive(Error, Debug)]
+pub enum LccUtilError {
+    #[error("LCC is not initialized; STATUS: {0}")]
+    NotInitialized(u32),
+    #[error("LCC transition mutex is already claimed.")]
+    MutexAlreadyClaimed,
+    #[error("Failed to claim the LCC transition mutex (value = 0x{0:x}).")]
+    FailedToClaimMutex(u32),
+    #[error("Bad post transition state (state = {0}).")]
+    BadPostTransitionState(String),
+    #[error("Functionality unimplemented.")]
+    Unimplemented,
+}
+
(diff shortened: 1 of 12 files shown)


tests/integration/src/test_lc_transitions.rs
```diff
@@ -0,0 +1,64 @@
...
+            .jtag_tap_connect(&jtag_params, JtagTap::LccTap)
+            .expect("Failed to connect to the LCC JTAG TAP.");
+
+        // Read the LC state.
+        let mut lc_state = read_lc_state(&mut *tap).expect("Unable to read LC state.");
+        println!("Initial LC state: {}", lc_state);
+        assert_eq!(lc_state, LifecycleControllerState::Raw);
+
+        // Perform the raw unlock LC transition operation.
+        const RAW_UNLOCK_TOKEN: [u32; 4] = [0xef1fadea, 0xadfc9693, 0x421748a2, 0xf12a5911];
```
timothytrippel, line 48: @ekarabu

I pulled this token from the test here: https://github.com/chipsalliance/caliptra-ss/blob/3952ef9a458f90d976cb606bf974d7398f4e3613/src/integration/test_suites/libs/lc_ctrl/lc_ctrl.c#L112; where is the origin of this constant?

I see what looked like the raw unlock token (preimage) constant here: https://github.com/chipsalliance/caliptra-rtl/blob/4a1f346bb318da3c5ce3cce5ca408be6e2b32de8/src/lc_ctrl/rtl/lc_ctrl_state_pkg.sv#L424

and the what looked like the hashed token (post image) here: https://github.com/chipsalliance/caliptra-rtl/blob/4a1f346bb318da3c5ce3cce5ca408be6e2b32de8/src/lc_ctrl/rtl/lc_ctrl_state_pkg.sv#L429

But the token that works is the one in the test above. I would like to link the RTL source to this line via a comment so that developers know to update this if the token changes.

hw/model/src/lcc.rs
```diff
@@ -0,0 +1,152 @@
...
+    // Check the LCC is initialized and ready.
+    let lcc_status = tap.read_lc_ctrl_reg(&LcCtrlReg::Status)?;
+    if LcCtrlStatus::from_bits_truncate(lcc_status)
+        != LcCtrlStatus::INITIALIZED | LcCtrlStatus::READY
+    {
+        bail!(LccUtilError::NotInitialized(lcc_status));
+    }
+
+    // Attempt to claim the LC transition mutex if it has not been claimed yet.
+    const MULTI_TRUE: u32 = 0x96;
```
timothytrippel, line 76: @swenson @zhalvorsen : are these mubi values defined somewhere globally in one of the repos like on OT: https://cs.opensource.google/opentitan/opentitan/+/master:sw/device/lib/base/multibits.h;l=6?q=multi&sq=&ss=opentitan%2Fopentitan ?

zhalvorsen: Not that I'm aware of

tests/integration/src/lib.rs
```diff
@@ -4,6 +4,8 @@ mod i3c_socket;
 mod test_firmware_update;
 #[cfg(feature = "fpga_realtime")]
 mod test_jtag_taps;
+#[cfg(feature = "fpga_realtime")]
+mod test_lc_transitions;
```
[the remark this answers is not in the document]
timothytrippel: Sure thing. Let me submit a follow up with said cleanup if that is OK?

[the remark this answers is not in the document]
zhalvorsen: Sounds good to me

PR #437, opened 2025-09-18
Onboard mctp-capsule-loopback test to FPGA CI

swenson:
The `FwId` and `REGISTERED_FW` abstraction doesn't quite work for specifying the different runtime binaries since they are in a different crate for FPGA. There's additional complexity in that a different runtime would also require a different SoC manifest to authorize it. So I added a slightly different mechanism for building and packaging them in the the `all-fw.zip` bundle.

The downside is that the FPGA test is flaky due to it halting when going through the full recovery flow to load MCU runtime.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -96,7 +96,9 @@ jobs:
       - name: Run all tests
         run: |
           # Build ROM and runtime binaries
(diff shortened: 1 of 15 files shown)


clundin25 (approved):

swenson:
Thanks!

PR #457, opened 2025-09-25
Remove process console by default to save 20 KB of code space

swenson:
`cargo xtask runtime` will still have the process console though.

The FPGA doesn't need a process console at all since it does not have keyboard input.

This also does some cleanup of the MCTP capsule to save code space:

* Using `u32` instead of `[u8; 4]` for MCTPHeader saves about 592 bytes of code space
* Using `u32` instead of [u8; 3]` for MCTPCtrlMsgHdr saves about 440 bytes of code space
* Using `u32::to_le_bytes()`, etc., instead of `zerocopy` saves about 100 bytes.


helloxiling (approved):
Looks good to me.

parvathib (approved):
LGTM

swenson:
Thanks!

PR #463, opened 2025-09-26
[xtask]: FPGA Workfow docs and usage improvements

clundin25:
- Add checks that both the FPGA and HOST have the correct dependencies installed.
- Store configuration in RAM so we can tell if FPGA needs to be bootstrapped.

### DIFF
--- a/docs/src/fpga.md
+++ b/docs/src/fpga.md
@@ -8,8 +8,11 @@ The machine that is used for development and cross compilation should have:
 
 - Rust
 - Docker
+  - Pull the FPGA build image: `$ docker pull ghcr.io/chipsalliance/caliptra-build-image:latest` 
 - rsync
 - git
+
+**NOTE:** Setup rootless docker. Here are the steps [here](https://docs.docker.com/engine/security/rootless/). 
 
 ### FPGA System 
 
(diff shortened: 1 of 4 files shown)


xtask/src/fpga/utils.rs
```diff
@@ -9,6 +9,50 @@ use std::{
...
+        (
+            "docker --version",
+            "'docker' not found on PATH. Please install docker.",
+        ),
+        (
+            "rsync --version",
+            "'rsync' not found on PATH. Please install rsync.",
+        ),
+        (
+            "cargo nextest --version",
```
zhalvorsen, line 24: I'm not sure how path resolution works with cargo installed programs.  Do we want `cargo nextest` here or `cargo-nextest` or are they equivalent?

zhalvorsen (approved):

docs/src/fpga.md
```diff
@@ -8,9 +8,12 @@ The machine that is used for development and cross compilation should have:
 
 - Rust
 - Docker
+  - Pull the FPGA build image: `$ docker pull ghcr.io/chipsalliance/caliptra-build-image:latest` 
 - rsync
 - git
 
+**NOTE:** Setup rootless docker. Here are the steps [here](https://docs.docker.com/engine/security/rootless/).
```
timothytrippel, line 15: Alternatively, in the future we could consider using podman. (Just noting here; not a blocker for this PR).

PR #467, opened 2025-09-26
[ROM] Reserve bits in boot flow for milestones. (#424)

clundin25:
This splits the MCI FW boot flow register into two pieces:

1. 16 checkpoint status bits (same as previous behavior)
2. 16 milestones bitflags

This is intended to make it easier to wait for certain milestones in the boot flow. With the FPGA we don't have strong control over timing so it could be easy to miss a checkpoint. These bitflags are a union of all the milestones seen ensuring they can't be missed.

The checkpoints are still very helpful for more fine grained boot flow debugging.

* Split at 16:16 instead of 8:24 to give move milestone bits.

This limits to 64k checkpoint options and 16 milestones.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2832,6 +2832,7 @@ name = "mcu-rom-common"
 version = "0.1.0"
 dependencies = [
  "bitfield",
+ "bitflags 2.9.4",
  "caliptra-api",
  "flash-image",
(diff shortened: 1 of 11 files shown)


clundin25:
Missed this commit in last cherry-pick

PR #470, opened 2025-09-29
Enable PLDM fw update e2e test in FPGA

mlvisaya:
- Test PLDM FW update protocol in FPGA by enabling the pldm_fw_update_e2e test

- Add MCI and MBOX SRAM drivers in FPGA board.rs since Image Loading App depends on it.

- Fix a bug in pldm update e2e test application, where the application doesn't exit since it's waiting for a SIGNAL() that hasn't been triggered. Moved the SIGNAL() into the pldm_fdops and trigger it on the last stage of PLDM which is activation.

- Update sleep_emulator_ticks while test is stepping.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,9 +78,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.11"
+version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "862ed96ca487e809f1c8e5a8447f6ee2cf102f846893800b20cebdf541fc6bbd"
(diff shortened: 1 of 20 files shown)


helloxiling:
The PR looks good to me!  
Just a quick question about the bug fix "_Fix a bug in pldm update e2e test application, where the application doesn't exit since it's waiting for a SIGNAL() that hasn't been triggered. Moved the SIGNAL() into the pldm_fdops and trigger it on the last stage of PLDM which is activation._":  why didn't we run into issues in the emulator test?

helloxiling (approved):
Looks good to me.

PR #475, opened 2025-10-01
DMA patch for new FPGA bitstream

mlvisaya:
- Upon testing the AXI DMA in FPGA, it was found that interrupts were not supported. As such, the AXI DMA driver is updated to support polling the DMA status registers by scheduling alarms.

- It was also found that the current DMA implementation doesn't support MCU SRAM access. This will be added on the next FPGA release. For now, a fake DMA driver is created which does manual SW transfer and emulates DMA. This is the nodma.rs driver.

- The DMA base peripheral address has been updated.

- Added the new FPGA Staging SRAM support in FW.

- Updated the emulator DMA addresses to be the same as the FPGA to simplify code.


mlvisaya:
Thanks!

PR #479, opened 2025-10-02
[test] add E2E FPGA manuf debug unlock test

timothytrippel:
This partially addresses #233 by adding an E2E FPGA test case that perfoms a manuf debug unlock flow via the Caliptra Core TAP.

### DIFF
--- a/rom/src/rom.rs
+++ b/rom/src/rom.rs
@@ -236,7 +236,21 @@ impl Soc {
         self.registers
             .fuse_soc_stepping_id
             .write(soc::bits::FuseSocSteppingId::SocSteppingId.val(soc_stepping_id));
-        // TODO: debug unlock / rma token?
+
+        if fuses.cptra_ss_manuf_debug_unlock_token().len()
+            != self.registers.fuse_manuf_dbg_unlock_token.len() * 4
+        {
+            romtime::println!("[mcu-fuse-write] SS manuf debug unlock token length mismatch");
+            fatal_error(1);
+        }
+        for i in 0..fuses.cptra_ss_manuf_debug_unlock_token().len() / 4 {
+            let word = u32::from_le_bytes(
+                fuses.cptra_ss_manuf_debug_unlock_token()[i * 4..i * 4 + 4]
+                    .try_into()
+                    .unwrap(),
+            );
+            self.registers.fuse_manuf_dbg_unlock_token[i].set(word);
+        }
     }
 
     pub fn fuse_write_done(&self) {
--- a/tests/integration/src/jtag/test_lc_transitions.rs
+++ b/tests/integration/src/jtag/test_lc_transitions.rs
@@ -3,33 +3,52 @@
 #[cfg(test)]
 mod test {
     use std::path::PathBuf;
+    use std::thread;
+    use std::time::Duration;
 
+    use caliptra_api::soc_mgr::SocManager;
+    use caliptra_hw_model::jtag::CaliptraCoreReg;
     use caliptra_hw_model::lcc::LcCtrlStatus;
     use caliptra_hw_model::openocd::openocd_jtag_tap::{JtagParams, JtagTap};
-    use caliptra_hw_model::{Fuses, DEFAULT_LIFECYCLE_RAW_TOKEN};
+    use caliptra_hw_model::HwModel;
+    use caliptra_hw_model::{
+        Fuses, DEFAULT_LIFECYCLE_RAW_TOKEN, DEFAULT_MANUF_DEBUG_UNLOCK_RAW_TOKEN,
+    };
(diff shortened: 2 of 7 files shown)


tests/integration/src/jtag/test_lc_transitions.rs
```diff
@@ -170,4 +199,134 @@ mod test {
...
+            Some(LifecycleControllerState::Dev),
+            /*debug_intent=*/ true,
+            /*bootfsm_break=*/ true,
+            /*enable_mcu_uart_log=*/ true,
+        );
+
+        println!("Give the FPGA time to boot...");
+        let mut flow_status = model.base.soc_ifc().cptra_flow_status().read();
+        println!("Flow status: 0x{:08x?}", u32::from(flow_status));
+        for _ in 0..5_000 {
```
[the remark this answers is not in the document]
timothytrippel: fixed.

tests/integration/src/jtag/test_lc_transitions.rs
```diff
@@ -170,4 +199,134 @@ mod test {
...
+        }
+        println!("Token programming complete.");
+
+        // Executing the manuf debug unlock operation and wait for completion.
+        println!("Executing the manuf debug unlock operation ...");
+        tap.write_reg(&CaliptraCoreReg::MboxExecute, 0x1)
+            .expect("Unable to write to MboxExecute register.");
+
+        // Wait for debug unlock operation to complete.
+        while let Ok(ss_debug_manuf_response) =
```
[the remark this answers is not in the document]
timothytrippel: fixed.

tests/integration/src/jtag/test_lc_transitions.rs
```diff
@@ -170,4 +199,134 @@ mod test {
...
+        assert_eq!(ss_debug_manuf_request, 1);
+        println!("Request sent.");
+
+        // Continue Caliptra Core boot.
+        tap.write_reg(&CaliptraCoreReg::BootfsmGo, 0x1)
+            .expect("Unable to write BootfsmGo.");
+
+        // Acquire the Caliptra Core mailbox lock.
+        println!("Attempting to acquire Caliptra Core mailbox lock ...");
+        while let Ok(mbox_lock) = tap.read_reg(&CaliptraCoreReg::MboxLock) {
```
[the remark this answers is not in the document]
timothytrippel: If the OpenOCD read register operation fails this could fail. I fixed this to also check that failures never occur.

rom/src/rom.rs
```diff
@@ -236,7 +236,21 @@ impl Soc {
         self.registers
             .fuse_soc_stepping_id
             .write(soc::bits::FuseSocSteppingId::SocSteppingId.val(soc_stepping_id));
-        // TODO: debug unlock / rma token?
+
+        if fuses.cptra_ss_manuf_debug_unlock_token().len()
+            != self.registers.fuse_manuf_dbg_unlock_token.len() * 4
```
zhalvorsen, line 241: nit: This looks fine, but I've seen comparisons like this go wrong in the past. Can we use `size_of_val` instead?


```suggestion
        if size_of_val(&fuses.cptra_ss_manuf_debug_unlock_token())
            != size_of_val(&self.registers.fuse_manuf_dbg_unlock_token)
```

rom/src/rom.rs
```diff
@@ -236,7 +236,21 @@ impl Soc {
...
+            fatal_error(1);
+        }
+        for i in 0..fuses.cptra_ss_manuf_debug_unlock_token().len() / 4 {
+            let word = u32::from_le_bytes(
+                fuses.cptra_ss_manuf_debug_unlock_token()[i * 4..i * 4 + 4]
+                    .try_into()
+                    .unwrap(),
+            );
+            self.registers.fuse_manuf_dbg_unlock_token[i].set(word);
+        }
```
zhalvorsen, line 253: nit: What do you think of using an iterator here instead?


```suggestion
        for (reg, fuses) in core::iter::zip(self.registers.fuse_manuf_dbg_unlock_token, fuses.cptra_ss_manuf_debug_unlock_token().chunks_exact(4) {
            let word = u32::from_le_bytes(fuses.try_into().unwrap());
            reg.set(word);
        }
```

rom/src/rom.rs
timothytrippel: good point. fixed.

rom/src/rom.rs
timothytrippel: This causes an illegal move because self.registers.fuse_manuf_dbg_unlock_token does not implement the Copy trait. So I think we are best using slices here.

zhalvorsen: Okay, this works for me!

zhalvorsen (approved):

PR #481, opened 2025-10-02
[Evidence] Add commands to get device state

parvathib:
Start adding commands to retrieve attester device's state.


swenson (approved):
LGTM

PR #482, opened 2025-10-02
Fix Misaligned exception on noDMA driver

mlvisaya:
The fake DMA driver that does manual SW copy of data from source to destination address can throw Misaligned load Exception if the source or destination is not 4-byte aligned. This is because the driver transfers in 4-byte chunks.

Transfer in bytes instead of words since is not assured that application will pass 4-byte aligned source or destination address.


swenson (approved):
LGTM

mlvisaya:
thanks!

PR #483, opened 2025-10-03
xtask: Add optional path to `caliptra-sw`

zhalvorsen:
This adds the optional path to add caliptra-sw to build_test for subsystem. This would be used when building tests with a local reference to caliptra-sw.

This will also delete the test binary archive before building, if it exists. Sometimes the docker build can fail silently, this will force the rsync to fail in those cases.

### DIFF
--- a/xtask/src/fpga/configurations.rs
+++ b/xtask/src/fpga/configurations.rs
@@ -90,6 +90,9 @@ impl<'a> ActionHandler<'a> for CommandExecutor {
     }
 
     fn build_test(&self, args: &'a BuildTestArgs<'a>) -> Result<()> {
+        // Delete the file if it exists. Sometimes the docker build fails silently. This will force
+        // the rsync to fail in those cases.
+        let _ = std::fs::remove_file("caliptra-test-binaries.tar.zst");
         match self {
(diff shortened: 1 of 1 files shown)


xtask/src/fpga/configurations.rs
```diff
@@ -153,8 +153,11 @@ impl<'a> ActionHandler<'a> for Subsystem {
         Ok(())
     }
 
-    fn build_test(&self, _args: &'a BuildTestArgs<'a>) -> Result<()> {
-        let mut base_cmd = build_base_docker_command(None::<String>)?;
+    fn build_test(&self, args: &'a BuildTestArgs<'a>) -> Result<()> {
+        // Delete the file if it exists. Sometimes the docker build fails silently. This will force
+        // the rsync to fail in those cases.
+        let _ = std::fs::remove_file("caliptra-test-binaries.tar.zst");
```
clundin25, line 159: Should this be on line 92? Then all the `build_test` types will delete the test binaries.

zhalvorsen: That's a good idea! I moved it to the higher level function.

PR #496, opened 2025-10-08
Firmware Update fixes and add HW model test for FW update and streami…

mlvisaya:
…ng boot

- Add common FW update integ tests for FW update and streaming boot.

- MCI interrupts were not enabled therefore MCU won't do the reset when activating MCU FW

- When handling MCI interrupts, MCI interrupt handler used the wrong register to clear the interrupt. corrected it.

- Fixed the logic in MCU ROM for Fw Hitless update. MCU ROM should enable the MCU Reset Interrupt and check if FW is available through the NotifCptraMcuResetReqSts interrupt bit. If it is set, it should clear it again and wait for Caliptra to clear FW_EXEC_CTRL[2] bit by checking the interrupt again.


mlvisaya:
will merge this once the dependent PR is in: https://github.com/chipsalliance/caliptra-sw/pull/2606

PR #513, opened 2025-10-20
errors: Create a central definition for fatal errors

### LINKED ISSUE #502: Unify Fatal Error Definitions
Currently `fatal_error` takes a u32 with magic numbers at the call sites. We should change this function to take an enum. This would make debugging which error was called significantly easier. It will also make it harder to have multiple call sites with the same error code used.

zhalvorsen:
Fixes #502

Creates unique error codes with the same strategy as Caliptra to define ROM fatal errors.

### DIFF
--- a/hw/model/test-fw/hitless_update_flow.rs
+++ b/hw/model/test-fw/hitless_update_flow.rs
@@ -6,9 +6,9 @@
 #![no_main]
 #![no_std]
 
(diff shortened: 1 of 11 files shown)


clundin25 (approved):
Awesome!

I think two things that could be valuable follow ups:

* Prefix ROM error codes with `ROM_` so it's easier to tell what stage the error is from.
* Make `fatal_error_raw` a private function and require every error path to have a unique error code.

PR #518, opened 2025-10-27
Verify Caliptra FW before activating

mlvisaya:
During FW update, verify caliptra FW with the other firmware components before activating it.
This uses the new FIRMWARE_VERIFY mailbox command to verify the caliptra firmware bundle.


parvathib (approved):
Looks great overall! Just a small suggestion—using an enum instead of a boolean for verify/load in `verify_or_load_caliptra` might make things a bit clearer and easier to read.

swenson (approved):
LGTM; I also like the idea of using `enum` instead of `bool`. It's a pattern I am trying to use more as well.

mlvisaya:
good suggestion. i'll make that param an enum instead of a bool

PR #524, opened 2025-10-29
Add download partition for fw update

mlvisaya:
Add a new flash partition for downloading PLDM fw update image. After the image in the download partition is verified, the image will be copied to the inactive A or B partition.
During the hitless update reboot, MCU will boot from the inactive partition with the valid image


mlvisaya:
> Looks good to me. Just a small suggestion: we usually name it as "staging partition" instead of "download partition".

ok, i can rename it to staging

PR #527, opened 2025-10-31
RISC-V compliance tests: pin sail to 0.19.1

swenson:
sail was recently updated to 0.20, which breaks the RISC-V sail module.

This pins us to use the known good sail.

Verified that this passes the compliance tests again: https://github.com/chipsalliance/caliptra-mcu-sw/actions/runs/18980302255


mlvisaya (approved):
lgtm!

PR #543, opened 2025-11-04
Switch to Tock 2.2 and move off nightly Rust

### LINKED ISSUE #430: Move to Tock 2.2+ and stable Rust
It looks like Tock now works with stable Rust, so we should move to that.

swenson:
Tock 2.2 added support for building on stable Rust. With a little bit of small tweaks and updates to Embassy, it looks like we can move off nightly Rust and use a proper released version of Tock.

Fixes #430

### DIFF
--- a/builder/src/runtime.rs
+++ b/builder/src/runtime.rs
@@ -184,23 +184,11 @@ pub fn runtime_build_no_apps_uncached(
         OBJCOPY_FLAGS
     );
 
-    // Add flags since we are compiling on nightly.
-    //
-    // - `-Z build-std=core,compiler_builtins`: Build the std library from source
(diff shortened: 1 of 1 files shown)


builder/src/runtime.rs
```diff
@@ -199,8 +199,8 @@ pub fn runtime_build_no_apps_uncached(
         "--verbose".into(),
         format!("--target={}", TARGET),
         format!("--package {}", bin),
-        "-Z build-std=core,compiler_builtins".into(),
-        "-Z build-std-features=core/optimize_for_size".into(),
+        // "-Z build-std=core,compiler_builtins".into(),
```
mlvisaya: nit: do these optimization options no longer work on the non-nightly version?

PR #562, opened 2025-11-06
docs: Add DOT spec start

swenson:
This adds the diagrams and the basic spec for DOT.

### DIFF
--- a/docs/src/dot.md
+++ b/docs/src/dot.md
@@ -92,8 +92,6 @@ TBD
 - Disable DOT while maintaining ownership (DOT_DISABLE)
 - Unlock and release ownership (DOT_UNLOCK)
 
-**MCU (Microcontroller Unit)**: Microcontroller component within Caliptra that executes ROM, FMC, and RT firmware.
-
 **ODD STATE**: Locked/Disabled state (fuse value % 2 == 1) where ownership is cryptographically bound to the silicon via DOT_BLOB.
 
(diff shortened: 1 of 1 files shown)


jameszhang-nvidia (approved):
LGTM, we should approve this PR so we can update OCP spec to point here as the caliptra example instead of having it in the actual OCP doc.

docs/src/dot.md
```diff
@@ -0,0 +1,1429 @@
...
+
+## Initialization Flow
+
+### Boot Sequence (DOT ROM Startup)
+
+<a id="dot-1-init"></a>
+
+```plantuml
+@startuml
+title DOT ROM Startup and DOT State Initialization
```
parvathib: Is this supposed to show a sequence diagram? It doesn't seem to render correctly in markdown.

docs/src/dot.md
```diff
@@ -0,0 +1,1429 @@
...
+```
+
+---
+
+## Security Considerations
+
+### Security Properties
+
+1. **Silicon Binding**
+   - DOT_ROOT_KEY is unique per device
```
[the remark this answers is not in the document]
jameszhang-nvidia: We should say this is integrator dependent.  DOT_ROOT_KEY for 2.0/2.1 will NOT reside in Caliptra Core so it will be something Caliptra MCU passes into Caliptra Core.  We should add that into the documentation. 
- DOT_ROOT_KEY is unique per device, for Caliptra 2.0/2.1, it is up to the integrator to decide how to form this per device unique key.

mhatrevi (approved):

docs/src/dot.md
swenson: I hope :)

swenson:
Diagrams did render correctly in mdbook: https://chipsalliance.github.io/caliptra-mcu-sw/dot.html

PR #563, opened 2025-11-07
feat: add caliptra-util-host library

mlvisaya:
This commit adds a complete caliptra-util-host library providing:

* Rust API for Caliptra MCU device communication
* C binding layer with comprehensive test suite
* Transport abstraction supporting mailbox protocol
* Session management with high-level command APIs
* Modular architecture: command-types, transport, session, cbinding

Key Changes:
- Added mailbox transport with automatic command translation
- Implemented checksum verification system for external mailbox protocol
- Fixed response checksum verification to use cmd=0 (not external_cmd_code)
- Request checksum calculation still uses external_cmd_code as required
- Added comprehensive test suite including MockMailbox for integration tests
- Provided C binding layer with CMockMailboxDriver for C/C++ integration

Library Structure:
- /command-types: Internal command definitions (DeviceInfo, Crypto, etc.)
- /transport: Transport layer with mailbox protocol implementation
- /session: High-level session management and command APIs
- /commands: Command packing/unpacking utilities
- /cbinding: C API bindings with opaque struct design
- /osal: OS abstraction layer for cross-platform support
- /tests: Rust integration test suite with MockMailbox
- /apps/mailbox: Reference client/server applications

Testing:
- All Rust integration tests: 5/5 passing
- All C binding tests: passing
- Transport layer tests: 4/4 passing
- Zero compilation warnings across workspace

To test with MCU integration, run:
cargo test --package tests-integration --lib --
test::test_caliptra_util_host_validator


caliptra-util-host/build.sh
```diff
@@ -0,0 +1,109 @@
...
+echo ""
+echo "Artifacts generated:"
+echo "  - Rust libraries: target/release/"
+echo "  - C bindings: caliptra-util-host-cbinding/target/release/"
+echo "  - C header: caliptra-util-host-cbinding/caliptra_util_host.h"
+echo "  - Examples: examples/"
+echo "  - Documentation: target/doc/"
+echo ""
+echo "To run examples:"
+echo "  cargo run --example basic_usage -- --help"
```
swenson: I think we should move this all into `xtask` -- if we require they have `cargo` installed, then we should continue to minimize the number of shell scripts.

caliptra-util-host/osal/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct PlatformInfo {
+    pub os_name: &'static str,
+    pub arch: &'static str,
+    pub has_std: bool,
+    pub has_alloc: bool,
+}
+
+pub fn platform_info() -> PlatformInfo {
+    PlatformInfo {
+        os_name: std::env::consts::OS,
```
swenson: Will these lines work in `no_std`?

caliptra-util-host/cbinding/tests/Makefile
```diff
@@ -0,0 +1,80 @@
+# Makefile for Caliptra C bindings tests
+
+# Compiler settings
+CC = gcc
```
swenson: I think it's usually best to use `?=` for these so they can be overridden

caliptra-util-host/build.sh
mlvisaya: that make sense. i will put all these scripts and cargo commands into an xtask.

mlvisaya:
> Possibly in a future PR, it would be good to include a way to run this on the FPGA.

Yes, i will make the hw-model based version of the integration test so we can run it in the FPGA.

mlvisaya:
Addressed the comments above in the new commit:

- Add a separate xtask command to run scripts for this library
- Link the docs to the the MCU docs/summary.md
- Create binaries for the mailbox applications
- Use ../../../target/caliptra-util-host for consistent builds
- Update READMEs
- - run cargo fmt/clippy with the latest cargo version

mlvisaya:
thanks for the review!

PR #570, opened 2025-11-11
fpga: Enable JTAG tests in GHA

swenson:
Fixes a syntax error that was causing tests to be skipped.


timothytrippel:
Ah thanks @swenson 

Looks like the reason the Raw unlock test is failing is the token is wrong: the lc_ctrl.STATUS register indicates wrong token (0b1000001 --> https://opentitan.org/book/hw/ip/lc_ctrl/doc/registers.html#status). The raw unlock token https://github.com/chipsalliance/caliptra-mcu-sw/blob/95634ef90661cd677c62654e790bc172cc857256/tests/integration/src/jtag/test_lc_transitions.rs#L62 seems to be wrong.

swenson:
Thanks!

PR #582, opened 2025-11-14
feat(rom): populate revocation fuses

ArthurHeymans:
- Uncomment vendor revocations production partition reading in OTP
- Load ECC, LMS, and MLDSA revocation values from fuses into CSRs
- Populate owner public key hash registers from fuses
- Add SOC IFC overrides application for core tests

### DIFF
--- a/rom/src/rom.rs
+++ b/rom/src/rom.rs
@@ -251,14 +251,8 @@ impl Soc {
         // TODO: vendor-specific fuses when those are supported
         // Load Owner ECC/LMS/MLDSA revocation CSRs.
         // ECC Revocation.
-        let vendor_ecc_revocation = u32::from_le_bytes(
-            fuses
-                .cptra_core_ecc_revocation_0()
-                .try_into()
-                .unwrap_or_else(|_| {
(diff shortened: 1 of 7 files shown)


rom/src/rom.rs
```diff
@@ -249,9 +249,58 @@ impl Soc {
...
-        // TODO: load ECC Revocation CSRs.
-        // TODO: load LMS Revocation CSRs.
-        // TODO: load MLDSA Revocation CSRs.
+        // Load Owner ECC/LMS/MLDSA revocation CSRs.
+        // ECC Revocation.
+        let vendor_ecc_revocation = u32::from_le_bytes(
+            fuses
+                .cptra_core_ecc_revocation_0()
+                .try_into()
+                .unwrap_or_else(|_| fatal_error(McuError::ROM_SOC_KEY_MANIFEST_PK_HASH_LEN_MISMATCH)),
```
zhalvorsen: Can you make unique error codes for these? It would be really helpful debugging if something goes wrong

swenson: Confirmed, no panic.

zhalvorsen: Lets just use `unwrap()` for these then

mhatrevi (approved):

PR #584, opened 2025-11-14
Add support for OCP EAT in SPDM attestation

parvathib:
Add support for OCP EAT in SPDM attestation.


swenson (approved):
LGTM, a couple of suggestions

PR #596, opened 2025-11-18
MCTP fix: don't send two IBIs in a row

swenson:
The tests require a strict sequence of IBI, private read, IBI, private read, etc.

But due to the way the XI3C controller was handled, we would process all IBIs (which there could be multiple), send those to the test code, and then try a private read and send that to the test code.

This instead queues up IBIs and private reads so that they are received in the correct sequence by the test; this fixes the spurious PEC mismatch errors and empty packets that the tests would occasionally see.


swenson:
We have a different mechanism in the emulator, so I don't think this error happens there?

PR #597, opened 2025-11-19
Setting DAI IDLE bit and direct access cmd reg offsets in generic straps

mhatrevi:


timothytrippel (commented):
can you just pass in the number of OTP partitions so you don't have to burn two  generic straps for this?

swenson:
I manually tested that the FE programming command is workin on the FPGA. We should add an automated test for that though.

PR #602, opened 2025-11-20
[test] fix RMA unlock test

timothytrippel:
This fixes the RMA unlock test. The fix required setting the RMA physical presents detection strap that has recently been exposed in the HW model in https://github.com/chipsalliance/caliptra-sw/pull/2813.

Fixes https://github.com/chipsalliance/caliptra-sw/issues/2755

### DIFF
--- a/rom/src/rom.rs
+++ b/rom/src/rom.rs
@@ -145,9 +145,7 @@ impl Soc {
             OTP_DIRECT_ACCESS_CMD_REG_OFFSET
         );
         self.registers.ss_strap_generic[0].set(OTP_DAI_IDLE_BIT_OFFSET << 16);
-        self.registers
-            .ss_strap_generic[1]
-            .set(OTP_DIRECT_ACCESS_CMD_REG_OFFSET);
+        self.registers.ss_strap_generic[1].set(OTP_DIRECT_ACCESS_CMD_REG_OFFSET);
 
         // PQC Key Type.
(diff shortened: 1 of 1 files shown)


tests/integration/src/jtag/test_lc_transitions.rs
```diff
@@ -168,35 +170,32 @@ mod test {
...
+        .expect("Unable to transition to RMA.");
+        println!("Post transition LC state: {}", lc_state);
 
-        assert_eq!(
-            *status,
-            LcCtrlStatus::TOKEN_ERROR | LcCtrlStatus::INITIALIZED
-        );
+        // Reset and read the LC state again.
+        model.set_subsystem_reset(true);
+        model.set_subsystem_reset(false);
```
zhalvorsen: nit: You should be able to use HwModel.cold_reset now that it is implemented in `caliptra-sw`

zhalvorsen (approved):

timothytrippel:
> I think this is just missing the `rom/src/rom.rs` changes from #597 and the UDS test passes locally for me.

Ahh, I will rebase on #597 then. Thanks.

tests/integration/src/jtag/test_lc_transitions.rs
timothytrippel: ah nice; fixed.

zhalvorsen (approved):

PR #612, opened 2025-11-25
warm_boot: Set cleared registers on warm reset

zhalvorsen:
Warm reset was missing setting some registers that are cleared on warm reset. This caused some issues with Core tests on Subsystem in `caliptra-sw`.

This creates a separate function to set the users to reduce the amount copy-pasted between cold reset and warm reset.

### DIFF
--- a/rom/src/rom.rs
+++ b/rom/src/rom.rs
@@ -564,7 +564,7 @@ pub struct AxiUsers {
     pub dma_user: u32,
 }
 
-impl<'a> From<&McuStraps> for AxiUsers {
+impl From<&McuStraps> for AxiUsers {
     fn from(straps: &McuStraps) -> Self {
         AxiUsers {
             mbox_users: [


rom/src/rom.rs
```diff
@@ -516,3 +555,28 @@ pub fn rom_start(params: RomParameters) {
...
+
+#[derive(Debug, Default)]
+pub struct AxiUsers {
+    pub mbox_users: [Option<u32>; 5],
+    pub fuse_user: u32,
+    pub trng_user: u32,
+    pub dma_user: u32,
+}
+
+impl<'a> From<&'a StaticRef<McuStraps>> for AxiUsers {
```
swenson: It feels a little weird to be taking `&'a` of a `StaticRef`. Why not just `impl From<&'static McuStraps> for AxiUsers`?

PR #619, opened 2025-12-01
Add GitHub workflow for caliptra-util-host validator testing

mlvisaya:
This commit introduces a new CI/CD nightly pipeline for the
caliptra-util-host validator component:

- Add .github/workflows/caliptra-util-host-validator.yml
This will spawn the Caliptra util host validator in another process
separate from the MCU emulator. They will communicate via a UDP socket.
- Create a test configuration TOML file for the validator that contains
the expected values and other validator configuration. The device under
test should pass this configuration file to the validator prior to
testing.
- Add timeout handling (5-second UDP socket timeout) to prevent
validator hanging
- Fix server and validator binary paths in xtask commands

### DIFF
--- a/.github/workflows/caliptra-util-host-validator.yml
+++ b/.github/workflows/caliptra-util-host-validator.yml
@@ -38,13 +38,6 @@ jobs:
           sudo apt-get install -qy build-essential curl gcc-multilib gcc-riscv64-unknown-elf git rustup &&
           rustup toolchain install -c clippy,rust-src,llvm-tools,rustfmt,rustc-dev
 
-      - name: Install cmake 3.x
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/caliptra-util-host-validator.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Build caliptra-util-host
+        run: |
+          pushd caliptra-util-host
+          cargo xtask build
+          popd
+
+      - name: Run Caliptra Util Host tests
+        run: |
+          pushd caliptra-util-host
+          cargo xtask test
```
swenson: I don't think we need to run every test?

PR #623, opened 2025-12-02
rom: Add stub DOT implementation and MCI FW fatal error

swenson:
This adds a simple DOT implementation in the ROM that always fails if the DOT is not empty.

Mostly this is implemented so that we can get all of the other tests and plumbing needed:

* Add `FlashStorage` impl in the ROM to support reading and writing the DOT blob
* Add a small memory in the emulator for the DOT flash
* Adds ability for tests to write to the DOT flash
* Sets the MCI FW fatal error register when a fatal error occurs
* Writes a simple test that a corrupt DOT flash causes an error

There appears to be a possible compiler bug in the `copy_from_slice()` operation, which is noted in the `SimpleFlash` struct: even though the lengths are guaranteed to be equal, the compiler will add a panic check. Even if I add an `assert_eq!` with the two length, the compiler seems to add the panic. Temporarily working around this using `ptr::copy_nonverlapping()`, which is what `copy_from_slice` simplifies to.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3267,6 +3267,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "network-interface"
+version = "1.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4a43439bf756eed340bdf8feba761e2d50c7d47175d87545cd5cbe4a137c4d1"
+dependencies = [
+ "cc",
+ "libc",
+ "thiserror 1.0.69",
+ "winapi",
+]
+
+[[package]]
 name = "nix"
 version = "0.26.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


rom/src/device_ownership_transfer.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Main Device Ownership Transfer flow executed during ROM boot.
+///
+/// This function orchestrates the DOT process, which includes:
+/// 1. Deriving the DOT_EFFECTIVE_KEY from hardware secrets and fuse state
+/// 2. Verifying the DOT blob authenticity using HMAC
+/// 3. Burning DOT fuses if a state transition is pending
+/// 4. Determining the final owner based on fuse state and DOT blob
+///
+/// # Arguments
+/// * `dot_fuses` - DOT fuse data.
```
parvathib: Minor nit: documentation for main fuses is missing here.

swenson:
Thanks!

PR #634, opened 2025-12-03
tests: Check that registers are up-to-date during precheckin

swenson:
This will ensure that we don't accidentally reset caliptra-ss.


timothytrippel (approved):

swenson:
Thanks!

PR #642, opened 2025-12-05
rom: Endianness incorrect for PROT_CAP2 and indirect fifo data

swenson:
I validated that this makes the PROT_CAP2 register emulator behavior match the FPGA behavior (there should be no endian swapping when reading or writing the register).

This requires a change in the AXI recovery bypass mode of the ROM so that the correct bit of the PROT_CAP2 register is read for the device_status field. It's a little bit confusing as the PROT_CAP2 register corresponds to bytes 8, 9, 10, and 11 of the PROT_CAP recovery block. Bytes 10 to 11 are the agent capabilities, of which bit 4 is the device status. The OCP recovery spec states that bytes 10 and 11 should be considered to be a little endian 16-bit word, so byte 10, bit 4 should be device status. When translated to a 32-bit little endian dword, this means bit 20 of our u32, not bit 4.

I'm still working on validating the indirect fifo data register behavior on FPGA, but I don't see any reason why it should be big endian when everything else is little endian in i3c-core.

### DIFF
--- a/emulator/periph/src/i3c.rs
+++ b/emulator/periph/src/i3c.rs
@@ -686,10 +686,9 @@ impl I3cPeripheral for I3c {
             registers_generated::i3c::bits::ProtCap2::Register,
         >,
     ) {
-        // convert the value in the other endianess
-        let x = u32::from_be_bytes(val.reg.get().to_le_bytes());
-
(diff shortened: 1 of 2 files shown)


mlvisaya:
thanks for fixing this. LGTM

swenson:
Thanks!

PR #644, opened 2025-12-05
[SPDM] Add chunking support for measurements response

parvathib:
This PR adds chunking support for the SPDM MEASUREMENTS block, validated against the CCC `spdm-requester-emu` tool.
The certificate chain continues to use implicit chunking; explicit support will be added in a future update.
The SPDM responder buffer size has been reduced from 2048 to 1024 bytes for chunking to effectively take place. This will now be the default mode.
The `SPDM-Responder-Validator` tool currently lacks support for chunked commands when executing positive flow tests. To accommodate this, a sufficiently large buffer (2048 bytes) is configured under the `large-buffer` feature flag.


swenson (approved):
LGTM

PR #647, opened 2025-12-05
build: remove build cache for runtime (#645)

### LINKED ISSUE #645: Remove build cache
The tracking issue of the removal of build cache.

pjhades:
This patch removes the build cache of runtime and always uses platform default values in builds.

Closes #645


swenson (approved):
LGTM

Awesome. I think these are more trouble than they are worth.

PR #650, opened 2025-12-06
tests: Mark failing tests as ignore for now

swenson:
And add feature flags that are missing

With this, all (non-ignored) tests pass again at least so we can re-enable tests as blocking and fix these tests in future PRs.


clundin25 (approved):

clundin25:
@swenson I have made all three checks required

PR #652, opened 2025-12-09
Write additional soc_ifc settings for caliptra-tests

ArthurHeymans:
This sets initial value of UDS_SEED and FIELD_ENTROPY.

### DIFF
--- a/rom/src/rom.rs
+++ b/rom/src/rom.rs
@@ -464,6 +464,43 @@ impl Soc {
                 .set(mcu_sram_addr as u32);
         }
 
+        // We use non secret production fuses to have caliptra tests pass some initial fuse values
+        if cfg!(feature = "core_test") {
+            // UDS Seed from vendor_non_secret_prod_partition (first 64 bytes)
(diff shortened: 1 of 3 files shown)


rom/src/rom.rs
```diff
@@ -464,6 +464,43 @@ impl Soc {
...
+            romtime::println!("[mcu-fuse-write] Writing field entropy");
+            if size_of_val(&fuses.vendor_non_secret_prod_partition[64..96])
+                != size_of_val(&self.registers.fuse_field_entropy)
+            {
+                fatal_error(McuError::ROM_SOC_FIELD_ENTROPY_LEN_MISMATCH);
+            }
+            for i in 0..self.registers.fuse_field_entropy.len() {
+                let word = u32::from_le_bytes(
+                    fuses.vendor_non_secret_prod_partition[64 + i * 4..64 + i * 4 + 4]
+                        .try_into()
```
clundin25: Nit: I think it would be easier to read an iterator, you should be able to use `chunks_exact`

clundin25 (approved):

PR #653, opened 2025-12-09
rom: rom trap wrapper

ternst-nv:
Wrap the rom exception handler in assembly which sets a custom stack pointer.  This allows the exception handler to rely on the stack within rust code, regardless of the state of the previously running code (which could have encountered stack overflow or corruption).  

Verification:  Ran a case which previously output less helpful information in the exception handler due to stack overflow and verified it included a useful mepc, and stack pointer.

exception-wrapper


swenson (approved):
LGTM

Thanks!

PR #655, opened 2025-12-10
[xtask-fpga]: Add support for package filters when building tests

clundin25:
This reduces the number of artifacts being built and copied to the FPGA.

Note: This requires the latest docker image for `caliptra-build-image`.

### DIFF
--- a/xtask/src/fpga/configurations.rs
+++ b/xtask/src/fpga/configurations.rs
@@ -183,10 +183,15 @@ impl<'a> ActionHandler<'a> for Subsystem {
         Ok(())
     }
 
-    fn build_test(&self, _args: &'a BuildTestArgs<'a>) -> Result<()> {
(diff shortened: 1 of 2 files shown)


xtask/src/fpga/configurations.rs
```diff
@@ -183,10 +183,15 @@ impl<'a> ActionHandler<'a> for Subsystem {
...
     }
 
-    fn build_test(&self, _args: &'a BuildTestArgs<'a>) -> Result<()> {
+    fn build_test(&self, args: &'a BuildTestArgs<'a>) -> Result<()> {
         let mut base_cmd = build_base_docker_command()?;
+        let package_filter_set = if let Some(package_filter) = args.package_filter {
+            format!("-E '{package_filter}'")
+        } else {
+            String::new()
+        };
```
zhalvorsen: nit: You could simplify this and the others to 


```suggestion
        let package_filter_set = args.package_filter.unwrap_or_default();
```

PR #658, opened 2025-12-10
Add support for imaginary flash controller on FPGA

helloxiling:


helloxiling:
@mlvisaya, I updated this PR after your review. The flash controller integration test was onboard to FPGA CI. It works well.  Could you please help review the latest changes before I merge? Thanks!

PR #662, opened 2025-12-10
Switch to nexttest for running emulator test

mlvisaya:
For some reason, the test_soc_boot is not being run by CI. Currently CI uses 'cargo test' to run
the unit and integration tests.
Switching to 'cargo nexttest run' was able to solve this problem and run all tests.
The nexxtest for emulator is using a different profile since the tests are slower than the FPGA so the timeouts are different.

It was also found that ROM isn't compiling properly due to the ROM code size is above 32KB. To fix this,
the ROM_SIZE was increased to 64 KB.

### DIFF
--- a/xtask/src/test.rs
+++ b/xtask/src/test.rs
@@ -42,7 +42,7 @@ fn cargo_test() -> Result<()> {
             "--exclude",
             "syscalls_tests",
             "--test-threads=1",
-            "--profile=nightly",
+            "--profile=nightly-emulator",
         ])
         .status()?;
 
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -9,3 +9,13 @@ slow-timeout = { period = "30s", terminate-after = 6 }
 path = "/tmp/junit.xml"
 store-success-output = true
 store-failure-output = true
+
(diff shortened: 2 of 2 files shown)


xtask/src/test.rs
```diff
@@ -13,15 +13,43 @@ pub(crate) fn test() -> Result<()> {
...
 fn cargo_test() -> Result<()> {
-    println!("Running: cargo test");
-    let status = Command::new("cargo")
+    // Run all tests with nextest for proper sequencing, excluding ROM packages that don't have tests
+    println!("Running: cargo nextest run");
+    let nextest_status = Command::new("cargo")
         .current_dir(&*PROJECT_ROOT)
-        .args(["test", "--workspace"])
+        .args([
+            "nextest",
```
mlvisaya: note: the reason why we put --exclude <package> here is because we need to exclude packages that doesn't have #[test], otherwise nexttest fails to generate list. I also removed the libtock unit tests because it contains code that we're not using.

swenson (approved):
LGTM

mlvisaya:
Thank you!

PR #663, opened 2025-12-10
[xtask-fpga]: Filter caliptra fw binaries and allow specifying the FW…

clundin25:
…ID as a flag

The binaries take a long time to compile and filtering the list down speeds up the `build` command.

### DIFF
--- a/xtask/src/fpga/utils.rs
+++ b/xtask/src/fpga/utils.rs
@@ -299,3 +299,27 @@ pub fn download_bitstream_pdi<P: AsRef<Path>>(
     )?;
     Ok(())
 }
+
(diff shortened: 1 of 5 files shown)


xtask/src/fpga/utils.rs
```diff
@@ -299,3 +299,27 @@ pub fn download_bitstream_pdi<P: AsRef<Path>>(
     )?;
     Ok(())
 }
+
+pub fn build_caliptra_firmware(caliptra_workspace: &PathBuf, fw_id: Option<&str>) -> Result<()> {
+    let fw_dir = PathBuf::from("/tmp/caliptra-test-firmware");
+    run_command(
+        None,
+        "mkdir -p /tmp/caliptra-test-firmware/caliptra-test-firmware",
+    )?;
```
clundin25: Should just port this to Rust...

PR #664, opened 2025-12-10
tooling: Allow alternatives to docker

cfrantz:
Some environments do not permit docker or require exceptions to organization policies to use docker.  Allow use of docker-compatible alternatives such as `podman` via the `DOCKER` environment variable.


clundin25:
@cfrantz I think we can just make `podman` the default and fallback to `docker`. This approach is more flexible but I think we can be more opinionated in the tool.

PR #668, opened 2025-12-11
rom: Add HMAC verification of Device Ownership Transfer blob

swenson:
And start definining some of the pieces of the DOT blob.
    
This adds a helper function in the tests to make it easier to create DOT blobs for testing as well.

### DIFF
--- a/rom/src/device_ownership_transfer.rs
+++ b/rom/src/device_ownership_transfer.rs
@@ -144,6 +144,7 @@ impl DotBlob {
 ///
 /// # Arguments
 /// * `env` - Mutable reference to the ROM environment containing hardware interfaces.
+/// * `main_fuses` - Main fuse data.
 /// * `dot_fuses` - DOT fuse data.
 /// * `blob` - DOT blob loaded from storage.
 /// * `stable_key_type` - The type of stable key to derive to verify the DOT blob with.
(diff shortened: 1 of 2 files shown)


romtime/src/soc_manager.rs
```diff
@@ -227,6 +232,43 @@ impl CaliptraSoC {
             expected_checksum,
         }))
     }
+
+    /// Executes a mailbox request that is represented as a u32 slice and
+    /// writing the response to a u32 slice.
+    /// This is useful for code size to avoid unaligned and byte-level access,
+    /// when possible.
+    pub fn exec_mailbox_req_u32(
```
helloxiling: It's a good optimization. We can port it to runtime mailbox API as well, if we have similar use cases.

rom/src/device_ownership_transfer.rs
```diff
@@ -46,7 +143,7 @@ pub struct DotBlob {
 /// 4. Determining the final owner based on fuse state and DOT blob
 ///
 /// # Arguments
-/// * `main_fuses` - Main fuse data.
+/// * `env` - Mutable reference to the ROM environment containing hardware interfaces.
 /// * `dot_fuses` - DOT fuse data.
```
parvathib: nit: `main_fuses` argument got replaced with `env` I guess :)

parvathib (approved):
LGTM

Issue #508, opened 2025-10-16
Call Caliptra ACTIVATE_FIRMWARE for activating custom SoC images in streaming boot


babraham9:
Caliptra requires MCU to call ACTIVATE_FIRMWARE after calling AUTHORIZE_AND_STASH to set the corresponding SS_GENERIC_FW_EXEC_CTRL bit. This allows the custom SoC core to start executing the streamed firmware.
But, the MCU SDK image_loading API only calls Caliptra AUTHORIZE_AND_STASH and does not call ACTIVATE_FIRMWARE. Suggest calling ACTIVATE_FIRMWARE to allow execution of the custom SoC core.

swenson:
@mlvisaya do we have more to do here?

PR #671, opened 2025-12-11
Set FW_EXEC_CTRL bit of SoC after loading image

mlvisaya:
Fixes: https://github.com/chipsalliance/caliptra-mcu-sw/issues/508

Set the FW_EXEC_CTRL of the the SoC after loading its image. This is done by sending the ACTIVATE_FIRMWARE mailbox command to Caliptra. Since the SoC architecture is platform dependent, there could be additional integrator steps to activate the SoC. Therefore, this is just an example of how the SoC can be activated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5104,6 +5104,7 @@ version = "0.1.0"
 dependencies = [
  "arrayvec",
  "async-trait",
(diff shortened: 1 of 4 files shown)


swenson (approved):
LGTM

PR #673, opened 2025-12-12
build: set app stack size in builder (#590)

### LINKED ISSUE #590: `builder/src/app.rs` and the application's `stack_size` should use the app's data size as the source of truth (including using it for the ld script constants) rather than having to tweak multiple constants when changes are needed

pjhades:
This patch sets the app stack size in builder, along with other app-specific
constants to avoid having to modify multiple places when updates are
required. To achieve that, instead of calling `stack_size!`, it reserves
stack space of the desired size in the linker script.

Closes #590

### DIFF
--- a/builder/src/apps.rs
+++ b/builder/src/apps.rs
@@ -13,11 +13,13 @@ pub const EMULATOR_APPS: &[App] = &[
         name: "example-app",
         permissions: vec![],
         minimum_ram: 48 * 1024,
+        stack_size: 0x7600,
     },
     App {
         name: "user-app",
         permissions: vec![],
         minimum_ram: 116 * 1024,
+        stack_size: 0xae00,
     },
 ];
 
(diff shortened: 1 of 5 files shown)


pjhades:
There might be better ways to handle this. I ended up choosing this way because
- `mcu-builder` depends on `std` but both apps and `libtock_runtime` (where `stack_size!` is defined) are `no_std` so we can't put the constants in the builder and use them in apps.
- Keeping app-specific constants with the apps makes more sense because the apps should define them rather than the builder.

But I had to add a lib.rs to expose the app constants. It would be nice to define the [`App` type](https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/builder/src/apps.rs#L38-L42) somewhere else, in order to avoid repeating `user_app::XYZ` in the builder. But I'm not sure where the ideal place would be.

swenson (approved):
LGTM. It would be nice if we didn't have to tweak the minimum_ram and stack_size, or the total stack size individually, but this is an improvement.

PR #676, opened 2025-12-13
hw model: Allow caliptra mailbox commands to be run from McuHwModel

swenson:
And removes an extra Caliptra root bus that was being created in the emulator model that wasn't hooked up to the rest of the emulator correctly.

We need to add extra methods to access Caliptra's mailbox and set AXI users appropriately (copied from Caliptra HwModel), and pass an additional mailbox/soc accessor out of the `start_caliptra` function.


mlvisaya (approved):
LGTM

swenson:
Thanks!

PR #695, opened 2025-12-30
[syscalls] Add OTP syscalls

zhalvorsen:
These syscalls will be used to read and write to OTP. This required moving a couple libraries from `mcu-rom-common` to `romtime`.


romtime/src/otp.rs
```diff
@@ -378,41 +378,41 @@ impl Otp {
         Ok(fuses)
     }
 
-    pub(crate) fn burn_lifecycle_tokens(&self, tokens: &LifecycleHashedTokens) -> McuResult<()> {
+    pub fn burn_lifecycle_tokens(&self, tokens: &LifecycleHashedTokens) -> McuResult<()> {
```
zhalvorsen: Should we move this function back to the `mcu-rom-common` crate because it was specifically limited to it previously?

swenson: I think this is okay since we might actually want to call this from non-ROM places, like a manufacturing firmware.

runtime/kernel/capsules/src/otp.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+}
+
+/// Provide an interface for userland.
+impl SyscallDriver for Otp {
+    fn command(&self, cmd: usize, arg1: usize, arg2: usize, processid: ProcessId) -> CommandReturn {
+        match cmd as u32 {
+            cmd::OTP_READ => self.read_reg(processid),
+            cmd::OTP_WRITE => self.write_reg(arg1 as u32, processid),
```
swenson: Do we need to use a subscribe / upcall for writing at least? I'm not sure how long it takes to burn this many fuses

PR #698, opened 2025-12-31
rom: Speed up flash boot scenario

swenson:
Rather than dword-at-a-time loading, try to load 256 bytes at a time from flash and write it in a tighter loop.

We also measure and print the number of cycles it takes to load the image and print the loading speed (bytes per 1,000 cycles).

Before: 32,178,550 cycles to load 197968 bytes
After: 1,597,018 cycles

This is a speedup of 20×, which greatly improves test speed.

There are some misc fixes as well:

* Move the `mcycle` code to `romtime`
* The emulated I3C peripheral was not setting the empty bit
* The `recovery.rs` was using the wrong bit to check for FIFO full.
* Simplify the instruction sequence for printing percentages in `recovery.rs` (small performance benefit of a few thousand cycles and probably a code size improvement).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4010,6 +4010,8 @@ dependencies = [
  "caliptra-api",
  "caliptra-registers",
  "registers-generated",
+ "riscv-csr",
+ "rv32i",
  "tock-registers",
  "ureg",
  "zerocopy",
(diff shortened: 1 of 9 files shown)


mlvisaya (approved):
LGTM

PR #700, opened 2025-12-31
pldm: Increase packet size to 196, remove unnecessary async

swenson:
Increases the default packet size from 180 to 196, which should speed up transfers and tests. This should still fit comfortably in an I3C packet size of of 250 bytes. (200 or 204 seemed to cause test failures.)
    
This also removes some unnecessary `async`s where `.await` is never called, and removes `AsyncAlarm` futures where we are just getting the current time. This saves about 3584 bytes in the runtime size.


runtime/userspace/api/caliptra-api/src/firmware_update/pldm_fdops.rs
```diff
@@ -52,7 +52,7 @@ impl UpdateFdOps {
 
 #[async_trait(?Send)]
```
helloxiling: We can remove this marker since all functions are converted to sync.

runtime/userspace/api/pldm-lib/src/firmware_device/fd_ops.rs
helloxiling: It seems that we can remove async for this method as well.

Issue #694, opened 2025-12-29
[tests] fix broken emulator and FPGA tests


timothytrippel:
Several test cases were disabled in #650, #687, and #874. They should be fixed and re-enabled. These tests include:

# Emulator Tests
- [x] test_firmware_update_flash_successful
    - [x] [main](https://github.com/chipsalliance/caliptra-mcu-sw/blob/957411dd01a9987616e2edcab5ab25e6abae5f6d/tests/integration/src/test_firmware_update.rs#L855)
    - [x] main-2.1)
- [x] test_firmware_update_flash_successful_fast
    - [x] [main](https://github.com/chipsalliance/caliptra-mcu-sw/blob/957411dd01a9987616e2edcab5ab25e6abae5f6d/tests/integration/src/test_firmware_update.rs#L865)
    - [x] main-2.1
- [ ] test_mctp_user_loopback
    - [x] [main](https://github.com/chipsalliance/caliptra-mcu-sw/blob/957411dd01a9987616e2edcab5ab25e6abae5f6d/tests/integration/src/lib.rs#L802) (#1009)
    - [ ] [main-2.1](https://github.com/chipsalliance/caliptra-mcu-sw/blob/e13b87a341c99cb82b5ef5ca6dc546b304397245/tests/integration/src/lib.rs#L610)

# FPGA Tests
- [ ] test_firmware_update_streaming_fpga
    - [ ] [main](https://github.com/chipsalliance/caliptra-mcu-sw/blob/957411dd01a9987616e2edcab5ab25e6abae5f6d/.config/nextest.toml#L44)
    - [ ] [main-2.1](https://github.com/chipsalliance/caliptra-mcu-sw/blob/e13b87a341c99cb82b5ef5ca6dc546b304397245/.github/workflows/fpga.yml#L260)
- [ ] test_mctp_capsule_loopback
    - [x] [main](https://github.com/chipsalliance/caliptra-mcu-sw/blob/957411dd01a9987616e2edcab5ab25e6abae5f6d/.config/nextest.toml#L44) (#1009)
    - [ ] [main-2.1](https://github.com/chipsalliance/caliptra-mcu-sw/blob/e13b87a341c99cb82b5ef5ca6dc546b304397245/.github/workflows/fpga.yml#L260)
- [ ] test_mctp
[...]

PR #687, opened 2025-12-23
[cherry-pick] from the `main` branch to the `main-2.1` branch from 12/02--12/17

timothytrippel:
This contains cherry-picks from the `main` branch to the `main-2.1` branch for the following PRs/commits from 12/02--12/17:
- [[xtask-fpga]: Fix boolean logic (](https://github.com/chipsalliance/caliptra-mcu-sw/commit/0c48fa892124de76458c7f275afb65778f93f34d)[#629](https://github.com/chipsalliance/caliptra-mcu-sw/pull/629)[)](https://github.com/chipsalliance/caliptra-mcu-sw/commit/0c48fa892124de76458c7f275afb65778f93f34d)
- [builder: platform specific bin (](https://github.com/chipsalliance/caliptra-mcu-sw/commit/029e1db96a8bcf0d79f94804a6f2421c367c6248)[#640](https://github.com/chipsalliance/caliptra-mcu-sw/pull/640)[)](https://github.com/chipsalliance/caliptra-mcu-sw/commit/029e1db96a8bcf0d79f94804a6f2421c367c6248)
- [rom: Endianness incorrect for PROT_CAP2 and indirect fifo data (](https://github.com/chipsalliance/caliptra-mcu-sw/commit/518cf5748993759ca92bc2158e2bb3cba2d7aeaa)[#642](https://github.com/chipsalliance/caliptra-mcu-sw/pull/642)[)](https://github.com/chipsalliance/caliptra-mcu-sw/commit/518cf5748993759ca92bc2158e2bb3cba2d7aeaa)
- [build: remove build cache for runtime (](https://github.com/chipsalliance/caliptra-mcu-sw/commit/c37b3064193794de4988184ff64343fde6695334)[#647](https://github.com/chipsalliance/caliptra-mcu-sw/pull/647)[)](https://github.com/chipsalliance/caliptra-mcu-sw/commit/c37b3064193794de4988184ff64343fde6695334)
- [rom: rom trap handler (](https://github.com/chipsalliance/caliptra-mcu-sw/commit/fc059b342e227599512cbca49a1455fe0d15c066)[#641](https://github.com/chipsalliance/caliptra-mcu-sw/pull/641)[)](https://github.com/chipsalliance/caliptra-mcu-sw/commit/fc059b342e227599512cbca49a1455fe0d15c066)
[...]

PR #704, opened 2026-01-05
[test] fix `test_soc_boot` tests

timothytrippel:
This partially addresses #694 by fixing the `test_soc_boot` tests that were disabled in #687 to unblock cherry-picks.

@swenson @mlvisaya : it is unclear why this fix is needed on the main-2.1 branch but no the main branch to enable the tests to pass. Any thoughts?

### DIFF
--- a/emulator/periph/src/i3c.rs
+++ b/emulator/periph/src/i3c.rs
@@ -248,7 +248,7 @@ impl I3c {
                         .read_recovery_interface(caliptra_emu_types::RvSize::Word, *start_addr)
                     {
                         Ok(data) => {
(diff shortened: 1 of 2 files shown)


swenson (approved):
The endianness in the AXI DMA was changed in the emulator between 2.0 and 2.1, so this is correct.

https://github.com/chipsalliance/caliptra-sw/blob/main/sw-emulator/lib/periph/src/dma/axi_root_bus.rs#L417

vs.

https://github.com/chipsalliance/caliptra-sw/blob/caliptra-2.0/sw-emulator/lib/periph/src/dma/axi_root_bus.rs#L417

PR #721, opened 2026-01-08
rom: Set ss_config_done_sticky, ss_config_done, prod debug unlock fuses, MCU mailbox users

### LINKED ISSUE #669: ROM check PK_HASH and MBOX_USER after locking
PROD_DEBUG_UNLOCK_PK_HASH_REG in MCI is lockable, but there are no access protections (MCU only writable). This could open us up to some other SW entity manipulating the PK hashes between configuration and locking. 

We need to add guidance that MCU ROM needs to check the PK hash after SS_CONFIG_DONE_STICKY is set to ensure it hasn't been manipulated. 

The same guidance should be given for MBOX[0,1]_VALID_AXI_USER and MBOX[0,1]_AXI_USER_LOCK

swenson:
And verify that they are set correctly.

Also updates the docs so that it is documented that the MCU ROM needs to do these things.

Fixes #669.

### DIFF
--- a/rom/src/rom.rs
+++ b/rom/src/rom.rs
@@ -491,18 +491,21 @@ impl Soc {
     }
 }
 
+/// Number of users supported by the MCU MBOX ACL mechanism.
+pub const MCU_MBOX_USERS: usize = 5;
+
 /// Structure to hold expected values for MCU mailbox AXI user configuration.
 /// Used for verification after SS_CONFIG_DONE_STICKY is set.
 #[derive(Debug, Default, Clone)]
 pub struct McuMboxAxiUserConfig {
     /// Expected values for MBOX0 valid AXI users (None = not configured)
-    pub mbox0_users: [Option<u32>; 5],
+    pub mbox0_users: [Option<u32>; MCU_MBOX_USERS],
     /// Expected values for MBOX1 valid AXI users (None = not configured)
-    pub mbox1_users: [Option<u32>; 5],
+    pub mbox1_users: [Option<u32>; MCU_MBOX_USERS],
     /// Expected lock status for MBOX0 AXI users
-    pub mbox0_locks: [bool; 5],
+    pub mbox0_locks: [bool; MCU_MBOX_USERS],
     /// Expected lock status for MBOX1 AXI users
-    pub mbox1_locks: [bool; 5],
+    pub mbox1_locks: [bool; MCU_MBOX_USERS],
 }
 
 /// Configures MCU mailbox AXI users in MCI and returns the configuration for later verification.
@@ -529,7 +532,7 @@ pub fn configure_mcu_mbox_axi_users(
     }
 
     // Configure MBOX1 AXI users based on straps
-    for (i, user) in straps.mcu_mbox0_axi_users.into_iter().enumerate() {
+    for (i, user) in straps.mcu_mbox1_axi_users.into_iter().enumerate() {
         // skip unconfigured users and avoid impossible panics
         if user != 0 && i < config.mbox1_users.len() && i < config.mbox1_locks.len() {
             romtime::println!(


docs/src/rom.md
```diff
@@ -176,4 +187,55 @@ On any fatal or non-fatal failure, MCU ROM can use the MCI registers `FW_ERROR_F
...
+### MCU Mailbox AXI User Verification
+
+The `MBOX0_VALID_AXI_USER` and `MBOX1_VALID_AXI_USER` registers control which AXI users can access the MCU mailboxes. These registers can be locked using `MBOX0_AXI_USER_LOCK` and `MBOX1_AXI_USER_LOCK`. Similar to the PK hash registers, these values could be manipulated before locking.
+
+**Security Requirement**: MCU ROM must configure the MCU mailbox AXI users, lock them, and verify that both the AXI user values and the lock register values match the expected configuration after both `SS_CONFIG_DONE_STICKY` and `SS_CONFIG_DONE` are set. If verification fails, the ROM must trigger a fatal error and halt.
+
+The verification process:
+1. MCU ROM configures `MBOX[0,1]_VALID_AXI_USER` registers with the expected AXI user values
+2. MCU ROM sets `MBOX[0,1]_AXI_USER_LOCK` to lock each configured slot
+3. MCU ROM sets `SS_CONFIG_DONE_STICKY` and `SS_CONFIG_DONE` to lock the configuration
```
clayton8: SS_CONFIG_DONE* do not change access for the MBOX* registers. Only the MBOX*_LOCK register will restrict access to the MBOX*_USER register.

docs/src/rom.md
```diff
@@ -176,4 +187,55 @@ On any fatal or non-fatal failure, MCU ROM can use the MCI registers `FW_ERROR_F
...
+2. MCU ROM sets `SS_CONFIG_DONE_STICKY` and `SS_CONFIG_DONE` to lock the registers
+3. MCU ROM verifies that both config done registers are actually set
+4. MCU ROM reads back the PK hash register values and compares them against the original fuse values
+5. If any mismatch is detected, MCU ROM reports a fatal error (`ROM_SOC_PK_HASH_VERIFY_FAILED`)
+
+### MCU Mailbox AXI User Verification
+
+The `MBOX0_VALID_AXI_USER` and `MBOX1_VALID_AXI_USER` registers control which AXI users can access the MCU mailboxes. These registers can be locked using `MBOX0_AXI_USER_LOCK` and `MBOX1_AXI_USER_LOCK`. Similar to the PK hash registers, these values could be manipulated before locking.
+
+**Security Requirement**: MCU ROM must configure the MCU mailbox AXI users, lock them, and verify that both the AXI user values and the lock register values match the expected configuration after both `SS_CONFIG_DONE_STICKY` and `SS_CONFIG_DONE` are set. If verification fails, the ROM must trigger a fatal error and halt.
```
clayton8: MCU ROM can verify after the DONE bits are set, but these don't change the access of the MBOX registers. The LOCK registers are what change the MBOX registers to RO. Are you having MCU check after DONE bits are set to help consolidate where we check the MCI registers?

swenson: I'll clarify.

clayton8 (approved):

rom/src/rom.rs
```diff
@@ -490,6 +491,188 @@ impl Soc {
     }
 }
 
+/// Structure to hold expected values for MCU mailbox AXI user configuration.
+/// Used for verification after SS_CONFIG_DONE_STICKY is set.
+#[derive(Debug, Default, Clone)]
+pub struct McuMboxAxiUserConfig {
+    /// Expected values for MBOX0 valid AXI users (None = not configured)
+    pub mbox0_users: [Option<u32>; 5],
```
mlvisaya: nit: maybe better to define a const for the MAX users?

swenson:
@RaunakGu good question.

@clayton8 @ekarabu I don't see any guidance on setting [FC_FIPS_ZEROZATION](https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.mci_top.mci_reg.FC_FIPS_ZEROZATION) in the integration spec or hardware spec -- how should we be setting this in the MCU ROM?

PR #727, opened 2026-01-12
firmware-bundler: LD File Generation

ternst-nv:
Add the firmware-bundler, and for its initial functionality add the ability to generate linker files. These will eventually be used by the build process to place embedded applications at specific locations in the memory space.

This was tested via unit tests, as well as examining the output for for various runs of:
```
cargo xtask experimental generate ...
```


swenson (approved):
LGTM, few small nits

PR #732, opened 2026-01-13
tests: Remove test-i3c-simple,test-i3c-constant-writes feature flags from emulator

swenson:
So that they can be tested in the FPGA as well. This should also improve test time since we reduce the number of emulator compilations we have to do.


mlvisaya (approved):
LGTM

parvathib (approved):

PR #759, opened 2026-01-15
docs: Update cold boot specification with latest information

### LINKED ISSUE #717: ROM flow documentation updates for cold boot
* For AXI bypass boot, can skip I3C basic initialization except for what is needed for the recovery interface
* Should assert caliptra boot go signal (same as warm reset)
* Move stashing to later (after step 4) since Caliptra isn't even up
* Clarify which reset is happening in step 11 and how it is generated
* Go into more detail about what happens during Caliptra boot in step 10

### LINKED ISSUE #744: Spec cleanup and bump version to 0.9

swenson:
Fixes #717
Fixes #744

Includes the documentation changes from #721

### DIFF
--- a/docs/src/mcu.md
+++ b/docs/src/mcu.md
@@ -1,6 +1,6 @@
 # Caliptra Manufacturer Control Unit (MCU) Firmware and SDK
 
-*Spec revision: 0.3*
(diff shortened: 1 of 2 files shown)


swenson:
Thanks!

PR #763, opened 2026-01-15
[xtask-fpga]: Cherry pick from main-2.1

clundin25:
These features make the development loop much faster by allowing unused binaries to be filtered away.


swenson (approved):
LGTM

I pulled it down and it seems to work for me with my petalinux setup.

PR #765, opened 2026-01-15
fix: Update MCU reset sequence for hitless firmware update (Cherry pick to 2.1)

mlvisaya:
This change refactors the MCU reset sequence to properly handle hitless firmware updates. Key changes include:

- Caliptra now triggers MCU interrupt by clearing FW_EXEC_CTRL[2] not through intr_trigger register.
- MCU ROM now jumps directly to MCU Runtime after reboot when reset_reason is HITLESS_UPDATE. Caliptra already does the verification and update before MCI releases reset.
- Ensure MCU stays in reset until Caliptra signals firmware is ready
- Update MCI emulator peripheral so that it can have access SOC_IFC. MCI reset behavior depends on the FW_EXEC_CTRL_BIT which is in the SOC_IFC.

Fixes: https://github.com/chipsalliance/caliptra-ss/issues/995
Depends-on: https://github.com/chipsalliance/caliptra-sw/pull/3128

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -298,7 +298,7 @@ checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
 [[package]]
 name = "caliptra-api"
 version = "0.1.0"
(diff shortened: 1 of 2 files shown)


swenson (approved):
Looks good other than git commit

Issue #705, opened 2026-01-05
Out of Tree Build Proposal


ternst-nv:
# Overview
	
The [caliptra-mcu-sw](https://github.com/chipsalliance/caliptra-mcu-sw) project is an open source IROT reference implementation, which is intended to be integrated with numerous vendors.  As such it would be useful to provide an easily usable build procedure generating binary packages which integrate software components provided by the Chip Alliance with vendor provided code.  Currently a similar build system exists in the project, however it has a number of configuration primitives which are determined as build time dependencies which requires any vendor integrator to make source code changes to the repository to utilize.  As such this proposal focuses on extending the capabilities of said system to allow easy integration by vendors, and as part of its development process should be integrated to support the `emulator` and `fpga` platforms currently provided by the reference application.

# Goals

* Provide a library which can easily integrate with xtask to provide the capability to generate machine loadable binaries as well as support iterative development for individual applications.  

E.g. `cargo xtask compose platform/vendor.toml` 

* All rustc options should be specified by the user via either .cargo/config.toml entries within their own project or command line overrides.  The composer should not manually add any rustc flags.

* The memory utilization should be intuitively specified by a definition file, and the utilization should be verified by the composer.  If a build would overflow budgeted memory use, it should fail.

* The vendor should be able to utilize the tool without having to make any source code changes to the caliptra-mcu-sw project which hosts the tool.

* The composer should be able to complete its work with only a single pass build of each component.  

# Stretch Goals
 
* Provide a tool which can be used as a binary application by an external vendor to create a firmware binary from a collection of rust dependencies.  This binary will only be able to provide the output binary files.

		E.g. composer pack <out of tree>/platform/vendor.toml

# Non-Goals

* Support for arbitrary MCU software systems.  Builds which are not utilizing the caliptra-mcu-sw reference architecture are not guaranteed/expected to work and cannot serve as motivating use cases for feature/bug development

* Overriding standard cargo build semantics.  The composer will by default only run when invoked.  This means it will not by default support `build.rs` semantics where the linker file is automatically provided every time the given application is compiled.  Instead the default behavior will be to compile without a linker file provided.

	Note: Effort will be made to provide APIs within the composer library for the integrator to replicate this functionality themselves if they so desire.

# Architecture

## Configuration

[...]

PR #766, opened 2026-01-16
firmware-bundler: Build Phase (#705)

ternst-nv:
Add the build phase of the firmware bundler tooling.  This phase takes a package name, and the linker script associated with it and compiles it via a `rustc` invocation.  It then takes the produced elf file and renders it as a binary via `objcopy` to enable the bundling step in the future.

As part of this correct a couple small bugs found in the `ld` file generation.

Testing:

Using the provided `simple.toml` manifest, verified that the emulator binaries produced via:

```
cargo xtask experimental build firmware-bundler/examples/simple.toml
```

resemble those produced by doing a runtime-build.

### DIFF
--- a/firmware-bundler/examples/simple.toml
+++ b/firmware-bundler/examples/simple.toml
@@ -5,29 +5,41 @@ tuple = "riscv32imc-unknown-none-elf"
 page_size = 256
 
 [platform.rom]
-offset = 0x0
-size = 0x10000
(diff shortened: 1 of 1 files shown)


swenson (approved):
LGTM; we should also maybe start on adding some documentation to `docs/` or `firmeware-bundler/README.md` on how to use this build system.

PR #770, opened 2026-01-19
fpga: Add flash control registers to wrapper (2.1)

swenson:
This add the flash controller registers from the "fake" flash peripheral to the wrapper registers. This adds both a primary and a secondary controller.

This will allow us to test flash-booting flows using AXI bypass recovery more easily.

The actual implementation will of the "hardware" will be in `ModelFpgaRealtime` because I am a software engineer. :)

This was tested by building the bitstream locally and using the new registers.


jlmahowa-amd (approved):

swenson:
Thanks!

PR #775, opened 2026-01-20
tests: Split up firmware update tests

swenson:
If one of the firmware update tests fails, it was difficult to tell which. Plus, the one test took a very long time.

This splits them up into separate tests so we get better visibility. Potentially, we could also use sharding to speed up tests.


mlvisaya (approved):
LGTM

PR #780, opened 2026-01-21
rom: OTP init set timeout before setting periodic checks

### LINKED ISSUE #691: MCU-ROM: OTP check timeout problem
Q1: MCU-ROM do not set [check_trigger](https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.otp_ctrl.CHECK_TRIGGER), but consistency/integrity still check after setting period. Why?
Q2: OTP starts consistency/integrity check and timeout happens when MCU-ROM set consistency/integrity check period, but [check_timeout](https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.otp_ctrl.CHECK_TIMEOUT) do not set yet. Should `check_timeout` be set first before setting check period?

https://github.com/chipsalliance/caliptra-mcu-sw/blob/main-2.1/rom/src/otp.rs#L63
Image

### LINKED ISSUE #769: [MCU ROM bug?]: OTP init results in OTP timer expiry and OTP error
### Bug (?) overview

I believe that there is a bug in the OTP driver for MCU. What I see is that the MCU ROM is initializing the OTP hardware and I think that during this init sequence the timeout expires instantly after configuration, **causing the ROM to be stuck forever.**

### MCU code

MCU ROM source code I used is from _main-2.1_ branch and is based on git hash _0d79b3fe1cb91acde30b7dbbac72587da26bb19c_.

### OTP driver init steps

The OTP driver init is somewhat as follows, and the problematic part is step 5:

1. Read the OTP [STATUS](https://chipsalliance.github.io/caliptra-ss/v2_1/regs/?p=soc.otp_ctrl.STATUS) register and check that there are no errors
2. Read the OTP **STATUS** register and check that DAI is idle
3. Configure [CONSISTENCY_CHECK_PERIOD](https://chipsalliance.github.io/caliptra-ss/v2_1/regs/?p=soc.otp_ctrl.CONSISTENCY_CHECK_PERIOD) register
4. Configure [INTEGRITY_CHECK_PERIOD](https://chipsalliance.github.io/caliptra-ss/v2_1/regs/?p=soc.otp_ctrl.INTEGRITY_CHECK_PERIOD) register
5. Configure the timeout value of 100 000 for register [CHECK_TIMEOUT](https://chipsalliance.github.io/caliptra-ss/v2_1/regs/?p=soc.otp_ctrl.CHECK_TIMEOUT)
6. Configure the register [CHECK_REGWEN](https://chipsalliance.github.io/caliptra-ss/v2_1/regs/?p=soc.otp_ctrl.CHECK_REGWEN)

### Problem description

However, from what I can see in emulation, looks like after configuring the **CONSISTENCY_CHECK_PERIOD** register, the RTL FSM state in the OTP lsfr timer transitions from “IdleSt” to state “CnstyWaitSt”. It looks to me like if the FSM is in this state when the **CHECK_TIMEOUT** register is configured, the timeout value is not configured to the non-zero value yet, however, the timeout logic immediately detects that a timeout happened.
A bit later in the cold boot flow, during fuses reading, the ROM is polling the OTP **STATUS** register until DAI idle bit is set. That never happens, and instead, the OTP **STATUS** register is filled with all the error bits set. The ROM is stuck polling the DAI idle bit forever.

I did some experimentation and in order to avoid getting into this error state, it’s **enough to place two reads of the OTP status register** in between steps 4 and 5 in the ROM code.

### Suggested solution

My suggested solution is to move step 5 and place it right after step 2. I tested this solution and it looks like it’s working (ROM doesn’t get stuck waiting for DAI idle any longer in read_fuses).

swenson:
This fixes a few issues (some raised in #691) related to OTP init:

* Setting the timeout before setting periodic checks
* Ensuring that the check pending is not set before continuing
* Waiting for idle and no error state at the end
* Add setting the periodic checks and timeouts as ROM parameters

The FPGA now boots with the integrity checks enabled.

Fixes #691.
Fixes #769.

### DIFF
--- a/romtime/src/otp.rs
+++ b/romtime/src/otp.rs
@@ -17,6 +17,7 @@ const OTP_STATUS_ERROR_MASK: u32 = (1 << 22) - 1;
 const OTP_CONSISTENCY_CHECK_PERIOD_MASK: u32 = 0x3ff_ffff;
 const OTP_INTEGRITY_CHECK_PERIOD_MASK: u32 = 0x3ff_ffff;
 const OTP_CHECK_TIMEOUT: u32 = 0x10_0000;
+const OTP_PENDING_CHECK_MAX_ITERATIONS: u32 = 1_000_000;
 
 pub struct Otp {
     registers: StaticRef<otp_ctrl::regs::OtpCtrl>,
@@ -31,12 +32,18 @@ impl Otp {
         self.registers.vendor_pk_hash_volatile_lock.set(1);
     }
 
-    pub fn wait_for_not_pending(&self) {
-        while self
-            .registers
-            .otp_status
-            .is_set(otp_ctrl::bits::OtpStatus::CheckPending)
-        {}
+    pub fn wait_for_not_pending(&self) -> McuResult<()> {
+        for _ in 0..OTP_PENDING_CHECK_MAX_ITERATIONS {
+            if !self
+                .registers
+                .otp_status
+                .is_set(otp_ctrl::bits::OtpStatus::CheckPending)
+            {
+                return Ok(());
+            }
+        }
+        crate::println!("[mcu-rom-otp] OTP pending check exceeded maximum iterations");
+        Err(McuError::ROM_OTP_PENDING_TIMEOUT)
     }
 
     pub fn check_error_and_idle(&self) -> McuResult<()> {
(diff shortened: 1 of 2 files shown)


romtime/src/otp.rs
```diff
@@ -12,36 +12,34 @@ use crate::{LifecycleHashedToken, LifecycleHashedTokens, LC_TOKENS_OFFSET};
...
+        Otp { registers }
     }
 
     pub fn volatile_lock(&self) {
         self.registers.vendor_pk_hash_volatile_lock.set(1);
     }
 
-    pub fn init(&self) -> McuResult<()> {
-        crate::println!("[mcu-rom-otp] Initializing OTP controller...");
+    pub fn wait_for_not_pending(&self) {
```
helloxiling: Is there any possibility that `CheckPending` bit is never cleared due to a hardware issue and this could block OTP initialization indefinitely?  
Would it be worth adding a software timeout?

helloxiling (approved):
Looks good to me!

PR #783, opened 2026-01-21
Use emulator ticks instead of wall clock time for tests

swenson:
This should help with tests being a bit more deterministic

### DIFF
--- a/tests/integration/src/test_fpga_flash_ctrl.rs
+++ b/tests/integration/src/test_fpga_flash_ctrl.rs
@@ -64,7 +64,6 @@ pub mod test {
                     break;
                 }
                 flash_controller.process_flash_ios();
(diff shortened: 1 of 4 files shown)


tests/integration/src/test_fpga_flash_ctrl.rs
```diff
@@ -64,7 +64,8 @@ pub mod test {
                     break;
                 }
                 flash_controller.process_flash_ios();
-                thread::sleep(Duration::from_millis(1));
+                // Sleep for 1,000 ticks (1ms at 1 MHz)
+                sleep_emulator_ticks(1_000);
```
mlvisaya: nit: it maybe helpful to have a milliseconds to tick conversion macro or function.

PR #784, opened 2026-01-21
Add network recovery boot design documentation

mlvisaya:


docs/src/network_boot.md
```diff
@@ -0,0 +1,816 @@
...
+```
+
+### Message Passing Protocol
+
+The protocol uses a LOCK-based mechanism for mutual exclusion, similar to mailbox designs. Since communication is one-way at a time, a single shared buffer serves both directions.
+
+#### Sending a Message
+
+**Sender Side (MCU or Network ROM):**
+1. Acquire lock: Write `1` to `CTRL[LOCK]` and verify it reads back as `1`
```
swenson: This feels too hardware-dependent.

docs/src/network_boot.md
```diff
@@ -0,0 +1,816 @@
...
+
+// Register offsets
+const CTRL_REG: usize = 0x00;
+const MSG_SIZE_REG: usize = 0x04;
+const MSG_OFFSET_REG: usize = 0x08;
+const SEQUENCE_REG: usize = 0x0C;
+
+// Control register bit masks
+const CTRL_LOCK: u32 = 1 << 0;
+const CTRL_DATA_READY: u32 = 1 << 1;
```
swenson: We probably don't need to list all of this code in the docs. Interfaces/traits are probably sufficient.

docs/src/network_boot.md
```diff
@@ -0,0 +1,816 @@
...
+
+```
+┌────────────────────────────────────────────┐
+│ Shared External Memory (MCU ↔ Network ROM) │
+├────────────────────────────────────────────┤
+│ Synchronization Registers (256 bytes)      │ ← Memory-mapped control registers
+├────────────────────────────────────────────┤
+│ Shared Message Buffer (8 KB)               │ ← Bidirectional message buffer
+├────────────────────────────────────────────┤
+│ Reserved for Future Use                    │
```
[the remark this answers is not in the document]
mlvisaya: yes this can be removed

docs/src/network_boot.md
```diff
@@ -0,0 +1,814 @@
+# Network Recovery Boot
+
+This document outlines the design for a lightweight network recovery boot utility for the Caliptra subsystem. The system enables the Caliptra SS to download firmware images over the network through a dedicated Network Boot Coprocessor within a ROM environment, providing a resilient fallback path when flash memory is corrupted.
+
+The network boot coprocessor acts as an intermediary between remote image servers and the Caliptra SS, handling network communications including DHCP configuration, TFTP server discovery, and firmware image downloads. The system supports downloading multiple firmware components including Caliptra FMC+RT images, SoC manifests, and MCU runtime images through a firmware ID-based mapping system.
```
parvathib: Should this list also include support for downloading the initial boot components of the SoC firmware

docs/src/network_boot.md
```diff
@@ -0,0 +1,814 @@
...
+
+        Caliptra <--> RecoveryIF
+        RecoveryIF <--> MCU_ROM
+        Caliptra <--> Mailbox
+        Mailbox <--> MCU_RT
+    end
+
+    MCU_ROM <--> Network_ROM["Network ROM<br/>- DHCP<br/>- TFTP Client<br/>- FW ID Mapping"]
+    MCU_RT <--> Network_ROM
+    Network_ROM <--> Image_Server["Image Server<br/>- Image Store<br/>- DHCP<br/>- TFTP Server<br/>- Config File"]
```
parvathib: Use `DHCP Server` here

docs/src/network_boot.md
```diff
@@ -0,0 +1,814 @@
...
+    MCU->>NET: Initiate boot request
+    NET->>IMG: DHCP Discovery
+    IMG-->>NET: DHCP Offer
+    NET->>IMG: TFTP GET config file (TOC)
+    IMG-->>NET: TOC (FW ID mappings)
+    NET->>NET: Store FW ID to filename mappings
+    NET-->>MCU: Network boot available
+```
+
+- **Early firmware image transfer**
```
parvathib: nit: May be good to add a line break here.

docs/src/network_boot.md
```diff
@@ -0,0 +1,814 @@
...
+        NET->>NET: Store FW ID to filename mappings
+    end
+    NET-->>MCURT: Network boot available
+```
+
+- **Remainder firmware images transfer**
+
+```mermaid
+sequenceDiagram
+    participant CORE as Caliptra RT
```
parvathib: Looks like Caliptra RT is not participating in this flow and can be removed.

docs/src/network_boot.md
```diff
@@ -0,0 +1,816 @@
...
+    }
+    Ok(())
+}
+```
+
+### Configuration File Format (TOC - Table of Contents)
+
+The network boot coprocessor downloads a configuration file (TOC) that maps firmware IDs to filenames and metadata:
+
+```json
```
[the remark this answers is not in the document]
mlvisaya: will follow Aspeed's recommendation

docs/src/network_boot.md
```diff
@@ -0,0 +1,814 @@
...
+    - Option 60: Bootfile Parameters (optional, for additional metadata)
+    - Note: DHCPv6 does not define a separate TFTP server option; use Bootfile URL
+
+## Boot Source Provider Interface
+
+The Boot Source Provider Interface defines a generic contract for boot image providers, enabling support for multiple boot sources (network boot coprocessor, flash device, or other custom implementations). The MCU ROM communicates with any boot source through this unified interface.
+
+### Messaging Protocol
+
+The boot source provider communication uses a simple request-response messaging protocol. The following section defines the message types, packet formats, and field definitions.
```
parvathib: It may help to include a small note explaining the message direction, such as: “MCU ROM sends all requests to the Network ROM, which returns the responses.”

docs/src/network_boot.md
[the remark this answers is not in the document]
mlvisaya: I can rename it to Caliptra Subsystem firmware images

docs/src/network_boot.md
```diff
@@ -0,0 +1,814 @@
...
+**Response Packet (per chunk):**
+```
+Offset  Size  Field              Description
+------  ----  -----              -----------
+0       1     Message Type       0x83 - Image Chunk
+1       1     Status             0x00=Success, non-zero=Error
+2       2     Sequence Number    For ordered delivery
+4       4     Offset             Current byte offset in image
+8       4     Chunk Size         Size of data in this chunk
+12      4     Total Size         Total image size (0 if unknown)
```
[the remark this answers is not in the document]
mlvisaya: I think we can just remove this and can use the Chunk Size to indicate end of transfer. There should be a CHUNK_SIZE parameter that both Network ROM and MCU ROM shares. If the transferred chunk is smaller than this parameter, then it's considered the last chunk.

mlvisaya:
Thanks!

PR #789, opened 2026-01-22
[EAT] update  protected header fields and fix its verification flow.

parvathib:
Make `content-type` a mandatory field in the `COSE_Sign1 Protected header` for all OCP EAT definitions, and update the data structures and OCP Profile verification flow accordingly


swenson (approved):
LGTM

PR #790, opened 2026-01-22
 allow CBOR tag(55799) and CWT tag(61) along with CoseSign1(18) tag, also use from_tagged_slice for CoseSign1 bytes

meilu-git:


ocp-eat-verifier/ocptoken-rs/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 authors = ["Caliptra Contributors"]
 
 [dependencies]
+ciborium = "0.2"
```
parvathib: You don’t need this dependency just for `Value`; use the `Value` re‑export from Coset instead.

ocp-eat-verifier/ocptoken-rs/src/token/evidence.rs
```diff
@@ -87,25 +84,40 @@ impl Evidence {
...
 /* -------------------------------------------------------------------------- */
 
-fn skip_cbor_tags(slice: &[u8]) -> OcpEatResult<Value> {
+fn verify_tags(slice: &[u8]) -> OcpEatResult<CoseSign1> {
     let mut value = Value::from_slice(slice).map_err(OcpEatError::CoseSign1)?;
 
     loop {
         match value {
-            Value::Tag(_, boxed) => value = *boxed,
+            // Verify allowed CBOR tags
```
parvathib: The tags must be validated in the fixed order 55799 → 61 → 18; other orders are not permitted. Also, if this function decodes the CoseSign1, consider giving it a more descriptive name (e.g., parse_tagged_evidence).

parvathib (approved):

swenson (commented):
I think we need to run fmt/clippy on this directory. I suspect it is not being checked by `cargo xtask precheckin`.

PR #809, opened 2026-01-27
Add the support for MCU mailbox ECDSA sign/verify commands

helloxiling:


tests/integration/src/test_mcu_mbox.rs
```diff
@@ -1516,6 +1526,232 @@ pub mod test {
...
+        }
+
+        /// Verify a signature using an ECDSA CMK.
+        fn ecdsa_verify(
+            &mut self,
+            cmk: &Cmk,
+            message: &[u8],
+            signature_r: &[u8; 48],
+            signature_s: &[u8; 48],
+        ) -> Result<bool, ()> {
```
mlvisaya: nit: could simplify return to Result<(),()>, since Ok(false) is not returned.

PR #813, opened 2026-01-28
emulator: Watchdog timer clamp fix

swenson:
This clamps additional timers so that they cannot overflow the underlying emulated CPU timers.


mlvisaya:
LGTM

PR #815, opened 2026-01-28
Resolve firmware update integration test hang

mlvisaya:
- Bring in the fix in caliptra-sw to fix the incorrect set sequence for
FW_EXEC_CTRL: https://github.com/chipsalliance/caliptra-sw/pull/3200
- Update RESET_STATUS immediately after FW_EXEC_CTRL bit is cleared
- Remove redundant timers in MCI emulator
- Instead of relying on timers to schedule reset, use a simple boolean
to indicate that MCU is on reset and reset MCU on the next tick
- Extend Activation timeout to 10 minutes, instead of 5 mins

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -240,31 +240,31 @@ libtock_small_panic = { path = "runtime/userspace/libtock/panic_handlers/small_p
 libtock_unittest = { path = "runtime/userspace/libtock/unittest" }
 
 # caliptra dependencies; keep git revs in sync
-caliptra-api = { git = "https://github.com/chipsalliance/caliptra-sw", rev = "a19319b60825977aa6813364eda6247507192967" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -240,31 +240,31 @@ libtock_small_panic = { path = "runtime/userspace/libtock/panic_handlers/small_p
 libtock_unittest = { path = "runtime/userspace/libtock/unittest" }
 
 # caliptra dependencies; keep git revs in sync
-caliptra-api = { git = "https://github.com/chipsalliance/caliptra-sw", rev = "5cff948de1a5da013c8edcc7efbf4a959387ec87" }
```
swenson: Reminder to update these before merging.

swenson (approved):
LGTM

PR #817, opened 2026-01-28
Temporarily disable FW Update integ test (2.0)

mlvisaya:
FW update integ test has some issues that are currently being investigated. Temporarily disable the FW update integ test so that other fixes can go in and unblock CI.


swenson:
Force merging these so that we can unblock some other PRs.

PR #824, opened 2026-01-30
Feature/firmware bundler two pass

ternst-nv:
Add the ability to do a two-pass build.  This will use the first build to determine the size of the various binaries in regard to both instruction and data memory.  Then that data is used to optimally allocate the memory available in the platform to the individual applications.

This was tested by bundling the emulator binaries and verifying the resulting binary was similar to the one created via the builder infrastructure.

Also add appropriate documentation.

### DIFF
--- a/docs/src/firmware_bundler.md
+++ b/docs/src/firmware_bundler.md
@@ -10,11 +10,38 @@ The firmware-bundler package is a library for unifying and simplifying the proce
 
 The firmware-bundler can either be used directly by platforms within the `caliptra-mcu-sw` repository, or integrated with a Vendor's Out of Tree repository via an `xtask` extension to Cargo.
 
+## Build Patterns
+
+The firmware-bundler supports 2 methods of determining the memory allocation of applications.  The first is a budgeted approach, where each application has a specified instruction and data memory usage within the manifest file.  The other is dynamic sizing, where the size of the memory segments are determined based on the actual memory used by the application.
+
+### Budget
+
+In the budget approach the memory requirements of each application are specified within the manifest, and then the firmware-bundler allocates out portions of the memory hierarchy to match those requirements.  This approach verifies that the application can fit in the specified memory region, and if not the compilation will fail with an appropriate error message.  Since the memory regions are known a priori, the firmware-bundler can use a single pass compilation of each binary.
+
+Pros:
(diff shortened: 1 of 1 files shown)


firmware-bundler/src/lib.rs
```diff
@@ -26,32 +26,110 @@ pub mod build;
...
+    const ALIGNMENT: u64 = 4096;
+
+    // A utility to increase the address to match the alignment requirement.
+    let align = |addr| {
+        if addr % ALIGNMENT != 0 {
+            addr + (ALIGNMENT - (addr % ALIGNMENT))
+        } else {
+            addr
+        }
+    };
```
pjhades: Is this `addr.next_multiple_of(ALIGNMENT)`?

docs/src/firmware_bundler.md
```diff
@@ -10,11 +10,38 @@ The firmware-bundler package is a library for unifying and simplifying the proce
...
+Pros:
+* Requires a single build pass for every application, which can significantly reduce overall build time.
+* Memory requirements are explicit, and are verified as part of the compilation process.
+
+Cons:
+* Manual adjustment is required if applications grow or change sizes in a way that requires different allocation of the overall memory space.
+
+### Dynamic Sizing
+
+In the dynamic sizing approach the memory requirements of individual applications are not defined, but determined dynamically as part of the bundling process.  While this doesn't effect the ROM binary (since it is the only binary deployed to ROM, it always gets the full memory hieararchy), this does have a significant impact on the runtime applications.  In this model the exec and data memory must not be specified for runtime applications, as it will be dynamically sized by the bundler.
```
pjhades: Would it be more user-friendly if we allow `exec_mem` and `data_mem` settings even if `dynamic_sizing = true` is present? If the user somehow switches back and forth between budgeted and dynamic sizing, they'd only need to comment/uncomment `dynamic_sizing = true`, without having to touch the memory settings. I think this may save them some effort especially when they have many apps. Of course the bundler can print a warning saying that, for example,
```
dynamic_sizing is on, ignoring exec_mem and data_mem settings.
```

pjhades (commented):
LGTM, left a few comments.

firmware-bundler/src/lib.rs
ternst-nv: It is indeed, thanks!

PR #832, opened 2026-01-31
fix(pldm-ua): Handle RequestFirmwareData from ReadyXfer state

mlvisaya:
Fix race condition in PLDM firmware update state machine where RequestFirmwareData could arrive before the StartDownload event is processed internally.

When the MCU receives UpdateComponentResponse, it sends the response back and immediately starts sending RequestFirmwareData. However, the UA internally queues a StartDownload event after receiving the response. If the first RequestFirmwareData arrives before StartDownload is processed, it would fail because the state machine was still in ReadyXfer state.

This fix adds a transition to handle RequestFirmwareData from ReadyXfer state, initializing transfer tracking and transitioning to Download state directly.


swenson (approved):
LGTM

PR #835, opened 2026-01-31
docs: Fix Identifier field values to match u32 fmt (cherry-pick from main)

iyernaveenr:
Cherry-pick of commit 47b495cb2634afd942fb7747a822f2961f1a2e0f from `main` to `main-2.1`.

This commit fixes the Identifier field values in documentation to match the u32 format.

Original PR: #798


iyernaveenr:
> @iyernaveenr there's a lot of other stuff in this PR. I think you're targeting the wrong branch.

My bad. You are right. I accidentally set the base branch to `main` instead of `main-2.1`, which caused Git to include all the commits that differ between the two branches. PR #837 fixes this by creating a fresh PR with the correct target branch. It also includes the other commit merged this morning.

Thanks!

PR #836, opened 2026-01-31
working on size history; current only

swenson:
Start on size_history

First stab at size history


mhatrevi:
This is now implemented; closing the PR.

PR #841, opened 2026-02-02
fpga: ci: migrate test filter to nextest profile (#719)

pjhades:
This commit moves the test filter for FPGA CI to the nextest profile, following the same pattern used in [caliptra-sw PR #3062](https://github.com/chipsalliance/caliptra-sw/pull/3062), in order to keep a single source of truth for CI tests.

Close #719

PR #845, opened 2026-02-03
xtask-fpga: migrate default test filter to nextest profile (#720)

### LINKED ISSUE #720: [xtask-fpga]: Use test filter profile as default
We can replace, e.g. https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/xtask/src/fpga/configurations.rs#L211-L222 with nextest profiles, similar to https://github.com/chipsalliance/caliptra-mcu-sw/issues/719.

pjhades:
This commit moves the default test filter for FPGA tests to nextest profile, so that we only maintain a single source of truth.

Please note that I don't have access to FPGA yet so I only tested by verifying the test command.

Close #720

### DIFF
--- a/xtask/src/fpga/configurations.rs
+++ b/xtask/src/fpga/configurations.rs
@@ -25,6 +25,17 @@ pub enum Configuration {
     Core,
 }
 
+impl Configuration {
+    pub fn default_test_profile(&self) -> &str {
+        match self {
+            Self::Subsystem => "nightly-ci",
+            // Test profiles defined in caliptra-sw
+            Self::CoreOnSubsystem => "fpga-subsystem",
+            Self::Core => "fpga-core",
+        }
+    }
+}
+
 pub enum CommandExecutor {
     /// Runs commands for a subsystem FPGA.
     Subsystem(Subsystem),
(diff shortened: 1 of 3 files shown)


.config/nextest.toml
```diff
@@ -19,3 +19,26 @@ slow-timeout = { period = "30m", terminate-after = 6 }
...
+(package(tests-integration) and test(test_firmware_update_streaming))
+"""
+
+[profile.nightly-fpga-core-on-subsystem]
+inherits = "nightly"
+default-filter = "package(caliptra-drivers)"
+
+[profile.nightly-fpga-core]
+inherits = "nightly"
+default-filter = "package(caliptra-drivers)"
```
clundin25: For the core and core-on-subsystem configurations we should use the profile define in `caliptra-sw`.

I think you will get this "for free" by passing `--profile fpga` or `--profile fpga-subsystem` on line 230 in the utils changes.

pjhades: Yeah I just realized that because of the error in CI 😅, will update it! Thanks for the pointer!

clundin25:
> @clundin25 Right I opened #841 earlier. Regarding the default filter it seems the filter in #841 is a superset of the one used here. So I think I'll keep that. ~Let me rebase this PR.~
> 
> Edit: I've updated this PR to refer to refer to the `nightly-ci` profile that I added in #841. That would be the default profile to run for the `Subsystem` configuration. `Core` and `CoreOnSubsystem` will run `fpga-core` and `fpga-subsystem` from caliptra-sw respectively. Filters specified on the CLI will override the default ones and use the `nightly` profile.

Oh whoops, missed that PR :)

PR #850, opened 2026-02-04
use EAT from caliptra 2.0 and remove it from caliptra-mcu-sw

meilu-git:


Cargo.lock
```diff
@@ -737,26 +738,31 @@ dependencies = [
  "zerocopy",
 ]
 
+[[package]]
```
parvathib: this is still appearing in the Cargo.lock and should be removed.

PR #852, opened 2026-02-04
Add Network Coprocessor ROM support

mlvisaya:
- Add a new instance of RISCv5 emulator for Network CPU
- Add network CPU in emulator model.
- Add build infrastructure for network ROM
- Implement network ROM hello world test
- Add network build in xtask and firmware binaries
- Add model-based integration test to verify network CPU boots and prints network CPU ROM banner

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -244,6 +244,7 @@ jobs:
             --exclude example-app \
             --exclude libtock_unittest \
             --exclude syscalls_tests \
+            --exclude network-rom \
             --archive-file /tmp/nextest-archive.tar.zst
           sccache --show-stats
 


helloxiling (commented):
The changes look good to me. 

Just a quick question: it looks like the network ROM integration test isn’t currently wired into CI. Do we plan to add a dedicated workflow for this?

mlvisaya:
> The changes look good to me.
> 
> Just a quick question: it looks like the network ROM integration test isn’t currently wired into CI. Do we plan to add a dedicated workflow for this?

Thanks. good question. The network ROM integration test is actually wired. If you are referring to the "--exclude network-rom" in the [build-test.yml](https://github.com/chipsalliance/caliptra-mcu-sw/pull/852/changes#diff-063ca77e012f959eba648db60e6868875fd1e70e5f5ac081193d848f8d61ef32R247), it is to prevent nextest to treat the network ROM FW as a test. "network-rom" is the actual RISC fw, so it shouldn't be treated as a test. You could also see the other ROM firmware are excluded there like the mcu-rom-emulator and mcu-rom-fpga.

In the CI logs, we can verify that the network ROM integration test ran: https://github.com/chipsalliance/caliptra-mcu-sw/actions/runs/21692961386/job/62559305020?pr=852#step:8:145

helloxiling (approved):
Looks good to me.

timothytrippel:
For future reference, this should probably be an RFC, where the design is posted somewhere in an issue.

PR #859, opened 2026-02-07
caliptra-util-lib: add the support for crypto commands SHA/HMAC/AES 

helloxiling:


caliptra-util-host/command-types/src/crypto_hmac.rs
```diff
@@ -0,0 +1,235 @@
...
+//! HMAC operations:
+//! - `HmacRequest` - Compute HMAC-SHA384/SHA512 over data using a key
+//!
+//! KDF operations:
+//! - `HmacKdfCounterRequest` - Derive a key using HMAC-based KDF in counter mode
+
+use crate::{CaliptraCommandId, CommandRequest, CommandResponse, CommonResponse};
+use zerocopy::{FromBytes, Immutable, IntoBytes};
+
+pub const MAX_HMAC_INPUT_SIZE: usize = 4096;
```
swenson: Maybe for a follow-up, but is the goal of this to be completely independent of `caliptra-api`? It seems like some of this could be imported and re-exported from `caliptra-api`.

PR #860, opened 2026-02-08
caliptra-util-lib: add the support for crypto cmds ECDSA/ECDH via mailbox transport

helloxiling:


mlvisaya (approved):
this is great, LGTM!

nit: you might also want to add the corresponding C rust bindings for these commands probably in another PR

PR #869, opened 2026-02-10
fpga: Route interrupts from flash controllers to VeeR PIC

swenson:
These can be turned on by the hw-model and the FW can clear them (and mask).

These should be tied to PIC interrupts 4 and 5

### DIFF
--- a/hw/fpga/src/caliptra_fpga_realtime_regs.rdl
+++ b/hw/fpga/src/caliptra_fpga_realtime_regs.rdl
@@ -252,12 +252,12 @@ regfile flash_ctrl_regs {
     reg {
         field {
             sw = rw;
-            hw = na;
+            hw = r;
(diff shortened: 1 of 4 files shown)


jlmahowa-amd (approved):
LGTM for once it's tested.

swenson:
Confirmed as working!

```
     90,098 UART: [mcu-rom] Flash interrupt status before read: 0x00000000
    104,430 UART: [mcu-rom] PIC pending[0..31]: 0x00000000
    116,139 UART: [mcu-rom] Flash interrupt status after read: 0x00000000
Flash op finish, set interrupt to 2
Finished flash op
  1,073,354 UART: [mcu-rom] Flash interrupt status: 0x00000002
  1,092,876 UART: [mcu-rom] PIC pending[0..31]: 0x00000010
```

PR #874, opened 2026-02-11
firmware-bundler: Integration

ternst-nv:
Switch the default build profiles for platform and fpga runtime environments to use the `firmware-bundler` to build the applications.

### DIFF
--- a/builder/src/rom.rs
+++ b/builder/src/rom.rs
@@ -12,10 +12,16 @@ use mcu_config::McuMemoryMap;
 use mcu_firmware_bundler::args::{BuildArgs, Commands, Common, LdArgs};
 
 pub fn rom_build(platform: Option<String>, features: Option<String>) -> Result<PathBuf> {
-    let rom = format!(
+    let feature_suffix = match &features {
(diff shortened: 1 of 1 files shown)


firmware-bundler/src/bundle.rs
```diff
@@ -67,8 +80,8 @@ pub fn bundle(
         runtime.extend(app.into_iter());
     }
 
-    // Firmware validated by Caliptra is required to be 4 byte aligned.
-    let aligned_bin_len = runtime.len().next_multiple_of(4);
+    // Firmware validated by Caliptra is required to be 256 byte aligned.
+    let aligned_bin_len = runtime.len().next_multiple_of(256);
```
pjhades: I might have missed some context. Why is this alignment changed? I searched in the document but didn't find a solid answer. I mean the alignment requirement should be fixed right? But from the logic point of view it makes sense no matter the requirement is 4 or 256 bytes because it is still 4-byte aligned.

pjhades (approved):
LGTM, and it's good to see the removal of code. 👍

PR #904, opened 2026-02-17
[xtask] Install libclang-dev in the FPGA build container

timothytrippel:
The lwip-rs crate requires bindgen to generate Rust bindings for the lwIP C library. bindgen depends on libclang to parse C headers. However, the caliptra-build-image does not have libclang installed, causing the FPGA test build to fail in the container.

This adds a conditional installation of libclang-dev in the build container before running the test archive command. The installation is only performed if libclang.so is not already present, minimizing the impact on build times.

This fixes an error that was introduced in #858.

### DIFF
--- a/xtask/src/fpga/utils.rs
+++ b/xtask/src/fpga/utils.rs
@@ -208,7 +208,8 @@ impl NextestArchiveCommand {
     }
 
     pub fn build(self) -> String {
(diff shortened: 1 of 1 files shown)


clundin25:
@timothytrippel opened https://github.com/chipsalliance/caliptra-infra/pull/26

timothytrippel:
Thanks I'll close this.

PR #909, opened 2026-02-17
Feature/reenable fast update test

ternst-nv:
When the system enters reset disable global interrupts.  This prevents
unexpected interrupts from:

* Triggering during MCI reset, accidently bringing the system to an
invalid execution state.
* Triggering prior to mtvec being set in an update reset causing an
infinite exception condition.

With this patch re-enable the fast firmware update test.


swenson (approved):
LGTM

PR #923, opened 2026-02-18
fpga: Install clang as part of xtask-fpga fpga build-test

swenson:
`clang` is not part of the base Docker image, but is required for lwip, which is a dependency of the network boot code. This code will probably be moved out of the caliptra-mcu-sw repo sometime soon, but for now, this is required to build tests for the FPGA locally for me.

### DIFF
--- a/xtask/src/fpga/utils.rs
+++ b/xtask/src/fpga/utils.rs
@@ -208,7 +208,9 @@ impl NextestArchiveCommand {
     }
 
     pub fn build(self) -> String {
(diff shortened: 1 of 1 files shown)


clundin25:
@swenson looks like a dupe of https://github.com/chipsalliance/caliptra-mcu-sw/pull/904

I opened https://github.com/chipsalliance/caliptra-infra/pull/26 to update the build image. Once it's merged and the deployed, it can be updated locally with `docker pull ghcr.io/chipsalliance/caliptra-build-image:latest`

PR #925, opened 2026-02-18
fix: replace mutex unwrap with safe error handling in cbinding

iyernaveenr:
Replace .lock().unwrap() calls with match-based error handling in emulator_send_uart_char and emulator_uart_rx_ready to prevent potential panics from crossing the FFI boundary, which is undefined behavior.


swenson (approved):
LGTM

PR #926, opened 2026-02-18
fix: replace transmute with typed function pointers in cbinding

iyernaveenr:
Replace *const c_void callback fields in CEmulatorConfig with Option<CExternalReadCallback> and Option<CExternalWriteCallback>, eliminating the need for std::mem::transmute. This provides compile-time type safety for callback function pointers.


swenson (approved):
LGTM

PR #928, opened 2026-02-18
tests: Actually use random ports for I3C

swenson:
I added `PortPicker` to change the ports used by each test for I3C. I thought that would pick random ports by default, but it turns out, it will just pick port 1024, then 1025, etc., which can cause race conditions if two tests are running at once (they both check that it is unused, see that it is, and then attempt to use it at the same time).

This changes the uses of `PortPicker` to actually use random ports, which drastically reduces the chances of two tests simultaneously picking the same port.


mlvisaya (approved):
LGTM

swenson:
Thanks!

PR #932, opened 2026-02-19
rom: Set WDTs on warm reset

### LINKED ISSUE #733: WDT timeout configuration
Hey,
I am getting on my FPGA setup a WDT1 and WDT2 timeout.
After some debugging I am seeing that it is being set to 1 cycle since the CORE ROM read 0 from some configuration on caliptra-sw/rom/dev/src/wdt.rs at the start_wdt function. When I increased it from 0x1 to higher value it passed.
I see it sets register:
https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.soc_ifc_reg.CPTRA_WDT_TIMER1_TIMEOUT_PERIOD%5B0%5D

Do you expect the SoC vendor to set it before starting? if so I see on the web I shared it is getting reset after de-assertion of reset so how come is it a bug?
If not who should set it up? MCU ROM?
If you can assist with understanding how and from where to set this configuration correctly and same for WDT2.

swenson:
We were missing some ROM configuration pieces on warm reset, like WDT configuration setting and verifying settings after the config done registers are set.

Fixes #733.

### DIFF
--- a/rom/src/warm_boot.rs
+++ b/rom/src/warm_boot.rs
@@ -15,8 +15,8 @@ Abstract:
 #![allow(clippy::empty_loop)]
 
 use crate::{
-    fatal_error, AxiUsers, BootFlow, McuBootMilestones, McuRomBootStatus, RomEnv, RomParameters,
-    MCU_MEMORY_MAP,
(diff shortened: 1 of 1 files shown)


swenson:
Thanks!

PR #940, opened 2026-02-24
rom: mtvec must be 256-byte aligned

swenson:
This is a requirement in the RISC-V spec.

Without this, the current caliptra-sw fails on update reset because the FPGA MCU ROM gets into a crash loop, due to an illegal instruction exception in the default MCU FW, which triggers an interrupt, which then resets due to a bad `mtvec`.


swenson:
Thanks!

PR #942, opened 2026-02-24
[xtask] add gen-fuse-config subcommand

### LINKED ISSUE #871: [tooling] develop tool to autogenerate fuse values from human readable form
We are developing reference provisioning firmware to write fuses during various lifecycle states. Each program is responsible for writing a select set of fuse fields. To facilitate writing / customizing these programs (which will be especially useful for downstream integrators), we should adopt OpenTitan's approach by developing tooling that can take as input a table of `fuse field`-->`value` entries (i.e., a key/value container) in a human readable form (e.g. JSON), and generate a rust library that statically defines the table, e.g. using the `phf` [crate](https://crates.io/crates/phf) (or something similar), that can be linked to a generic fuse programming wrapper program to enable creating customized fuse provisioning firmware images.

This tool could be a standalone rust tool or an xtask flow.

timothytrippel:
Adds a `gen-fuse-config` command to xtask to autogenerate a rust library that contains a static map of fuse field to value mappings from an input HJSON file that describes the fuse configuration for a chip.

Fixes #871.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -35,6 +35,26 @@ jobs:
       - name: Run precheckin
         run: cargo xtask precheckin
 
+  xtask-flows:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Install required packages
+        run: |
+          sudo apt-get update -qy && \
+          sudo apt-get install -qy build-essential curl git && \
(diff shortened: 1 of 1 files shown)


clundin25 (approved):
@timothytrippel would it be more ergonomic to have a crate + build script that generates this library? 

That way an application would just need to import the crate instead of copying this file around.

timothytrippel:
> @timothytrippel would it be more ergonomic to have a crate + build script that generates this library?
> 
> That way an application would just need to import the crate instead of copying this file around.

That is a fantastic idea. Let me tackle this in a follow up.

PR #970, opened 2026-02-26
recovery: Exit streaming boot after MCU image loaded; update Caliptra SW dependency

amit-kumarh:
And fix a compatibility issue to exit streaming boot state machine on booting MCU image instead of waiting for the recovery interface to be closed


swenson (approved):
LGTM

PR #972, opened 2026-02-27
rom: Add encrypted firmware boot support

swenson:
Implement encrypted MCU firmware boot flow gated behind the `core_test` feature flag. When the encrypted boot GPIO wire is set, MCU ROM:

- Sends RI_DOWNLOAD_ENCRYPTED_FIRMWARE to Caliptra RT, which loads firmware to MCU SRAM without activating the MCU
- Computes SHA-384 of ciphertext via streaming CM_SHA mailbox commands
- Imports AES key via CM_IMPORT
- Decrypts firmware in-place via CM_AES_GCM_DECRYPT_DMA

Supporting changes:
- Add error codes for encrypted firmware decrypt/activate failures
- Add boot status codes for the encrypted firmware flow
- Refactor mailbox helpers to use byte-oriented APIs (start_mailbox_req_bytes, finish_mailbox_resp_bytes)
- Add fw_exec_ctrl() and ready_for_runtime() to SoC manager
- Adjust FPGA ROM stack size

For now have the Mbox struct locally as we have a chicken and egg problem as caliptra-sw CI also needs this code working.

### DIFF
--- a/error/src/lib.rs
+++ b/error/src/lib.rs
@@ -323,6 +323,31 @@ impl McuError {
             "SOC field entropy length mismatch"
         ),
         (
+            ROM_COLD_BOOT_ENCRYPTED_FW_DECRYPT_START_ERROR,
+            0x1_0011,
+            "Cold boot encrypted firmware decrypt DMA start error"
+        ),
+        (
+            ROM_COLD_BOOT_ENCRYPTED_FW_DECRYPT_FINISH_ERROR,
+            0x1_0012,
+            "Cold boot encrypted firmware decrypt DMA finish error"
+        ),
+        (
+            ROM_COLD_BOOT_ENCRYPTED_FW_DECRYPT_TAG_MISMATCH,
+            0x1_0013,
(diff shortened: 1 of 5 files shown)


rom/src/device_ownership_transfer.rs
```diff
@@ -334,21 +343,31 @@ fn cm_hmac(env: &mut RomEnv, key: &Cmk, data: &[u8]) -> McuResult<[u32; 16]> {
         return Err(McuError::ROM_COLD_BOOT_DOT_ERROR);
     }
     // should be impossible for this slice to fail but the compiler seems to generate a panic
-    req.data
+    let src = data.get(..len).ok_or(McuError::ROM_COLD_BOOT_DOT_ERROR)?;
+    let dst = req
+        .data
         .get_mut(..len)
-        .ok_or(McuError::ROM_COLD_BOOT_DOT_ERROR)?
```
swenson: I don't think this code was panicking before. Can we revert these unrelated changes? I think this is also less performant since we are no longer transmuting.

PR #973, opened 2026-02-27
rom: Add OTP SW digest to driver

swenson:
We refactor the OTP digest code (PRESENT cipher) so that it is no_std and able to be used by the emulator or by the firmware. We also added an iterator-based digest method so that the entire OTP partition doesn't need to be copied to the stack, but can be read a dword at a time.


swenson:
Thanks!

PR #981, opened 2026-03-02
firmware-bundler: Re-enable MRAC definition

ternst-nv:
The firmware-bundler integration inadvertently removed MRAC calculation for ROM targets.  Re-enable the functionality by providing a const calculation of the value within the config libraries.  This enables the assembly start instructions to depend on the values, without requiring the firmware-bundler build process to be aware of MMIO maps it doesn't manage/interact with.

As part of this make the `compute_mrac` function constant by removing iterators, closures, and debug statements which are not supported in const functions.


swenson (approved):
LGTM

PR #996, opened 2026-03-03
rom: Add SW digest write lock

swenson:
Add a method to `Otp` that writes the SW digest of the partition data, similar to how the HW digest works. This is required for any partitions that have the SW digest marked in the OTP.

Also fix the emulator OTP to support 64-bit DAI writes by tracking the high word (wdata_1) alongside the existing low word buffer.


mlvisaya (approved):
LGTM

PR #1004, opened 2026-03-04
rom: Calculate ROM digest and stash

### LINKED ISSUE #931: rom: Add rom integrity check + stash option after Caliptra core runtime has booted
We should do a MCU ROM integrity check (SHA384) after Caliptra core runtime services are available, similar to Caliptra ROM's: https://github.com/chipsalliance/caliptra-sw/blob/main/rom/dev/src/main.rs#L260

This will require adding an expected SHA384 to an info block at the end of the MCU ROM.

And if we have already computed that SHA384, we can add the option to stash it to Caliptra as well at point.

mtimkovich:
Calculate and store SHA384 digest of ROM. Also update emulator to have ROM pad to its full size.
    
Also discard .eh_frame from the ROM linker script — this unused exception handling data was consuming ~4 KiB of DCCM. The ROM is a no_std bare-metal target with no stack unwinder, so .eh_frame serves no purpose.

Fixes #931


rom/src/cold_boot.rs
```diff
@@ -426,6 +606,60 @@ impl ColdBoot {
...
         }
     }
+
+    /// Calculate SHA384 hash of ROM and stash it.
+    fn stash_rom_digest(soc_manager: &mut CaliptraSoC) {
+        let rom_offset = unsafe { MCU_MEMORY_MAP.rom_offset } as usize;
+        const DIGEST_SIZE: usize = 48;
+        let rom_size = unsafe { MCU_MEMORY_MAP.rom_size } as usize - DIGEST_SIZE;
+        let digest = Self::cm_sha384_ciphertext(soc_manager, rom_offset, rom_size);
+        romtime::println!("[mcu-rom] MCU ROM digest: {}", HexBytes(&digest));
```
swenson: The next step in this would be we need to add the SHA384 of the ROM to the end of the ROM in the `builder/`, so that we can compare it here.

rom/src/cold_boot.rs
```diff
@@ -426,6 +606,60 @@ impl ColdBoot {
...
+
+        let mut req = StashMeasurementReq {
```
swenson: I think we want to make the SHA384 check above mandatory, but have the actual stash to Caliptra be optional.

rom/src/cold_boot.rs
```diff
@@ -940,6 +707,8 @@ impl BootFlow for ColdBoot {
             mci.set_flow_checkpoint(McuRomBootStatus::CaliptraRuntimeReady.into());
         }
 
+        Self::rom_digest_integrity(soc_manager, true);
```
swenson: Let's have whether or not we stash added to the `RomParameters`, with a default value of `true`.

firmware-bundler/src/build.rs
```diff
@@ -146,12 +146,20 @@ impl<'a> BuildPass<'a> {
...
-            .build_definition
-            .rom
-            .as_ref()
-            .map(|r| self.build_binary(r, &self.build_args.rom_features, ROM_OBJCOPY_FLAGS))
-            .transpose()?;
+        // Skip rebuilding ROM.
+        let rom = {
+            if let Some(app) = self.build_definition.rom.as_ref() {
+                let elf = self.binary_dir.join(&app.name);
+                let binary = elf.with_extension("bin");
```
swenson: I think this is too simple a check?

It feels like maybe we might want to eliminate the other code path that builds the ROM and add the digest appending here.

Maybe @ternst-nv has an idea.

PR #1005, opened 2026-03-04
[bugfix] fix STACK_TOP in baremetal ld script

amit-kumarh:
currently, we allocate the stack immediately following the text, but STACK_TOP is set to ORIGIN(RAM) + STACK_SIZE, which doesn't account for the text and thus isn't the actual top of the stack.


amit-kumarh:
Actually, the smarter thing here is probably to put the stack on the bottom, so an overflow (even if the stack is sized correctly) throws a fault and doesn't overwrite text

timothytrippel (approved):
Ah good catch.

PR #1011, opened 2026-03-06
emulator: Fix cross-thread timer scheduling race in I3C and DOE peripherals

swenson:
The I3C controller thread calls `PollScheduler::incoming()` which invokes `timer.schedule_poll_in(1)`. Inside `caliptra-sw`'s `Clock` `schedule_poll_in` reads `now()` once to compute the absolute time, then `schedule_action_at` reads `now()` again to assert the delta is `< i64::MAX`. When the main CPU step thread advances the clock between these two reads, the scheduled time appears to be in the past, wrapping to a huge delta that trips the assertion.

Add a shared step_lock (`Arc<Mutex<()>>`) that both the `PollScheduler` callbacks and the CPU step code acquire, ensuring the clock cannot advance while a timer action is being scheduled from another thread.

Also fix a lock-ordering inversion in `I3cTarget::send_command()` that would otherwise deadlock: release the target device mutex before calling `incoming()` (which now acquires the step lock).

### DIFF
--- a/emulator/app/src/emulator.rs
+++ b/emulator/app/src/emulator.rs
@@ -290,6 +290,9 @@ pub struct Emulator {
     pub doe_mbox_fsm: doe_mbox_fsm::DoeMboxFsm,
     pub i3c_address: Option<u8>,
     pub i3c_controller_join_handle: Option<JoinHandle<()>>,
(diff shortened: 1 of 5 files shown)


ternst-nv (approved):

swenson:
Thanks!

PR #1013, opened 2026-03-06
Add mcu runtime identity attestation design

parvathib:


docs/src/mcu_rt_identity_attestation.md
```diff
@@ -0,0 +1,187 @@
...
+- **Minimal Caliptra 2.x RT firmware changes** — Caliptra RT's will continue to function as before for the Auth Manifest version 2. For Version 3, the lower 16 bits are interpreted as Auth Manifest SVN and the upper 16 bits are interpreted as MCU RT FW SVN and passed to `DeriveContextCmd`, but the core logic of manifest verification and image authentication remains intact.
+- **Works for encrypted firmware** — MCU RT FW SVN is in the Auth Manifest preamble (always cleartext), not embedded in the MCU image.
+- **Cryptographically protected** — Both SVN sub-fields are covered by the Auth Manifest vendor/owner signatures. Tampering with either sub-field invalidates the manifest signature.
+- **No MCU image format dependency** — Caliptra RT remains agnostic to the MCU image internal structure.
+- **<Backward compatible>** — Existing deployments that set `preamble.svn` as a simple 16-bit value (upper bits zero) continue to work. MCU RT FW SVN defaults to 0 in that case.
+
+### Limitations
+
+- **MCU RT FW SVN range limited to 16 bits (u16)** — Maximum value of 65,535. Sufficient for practical SVN lifetimes but a reduction from the theoretical u32 range.
```
fdamato: 16 bits for SVN typically means 16 iterations... fuse can`t be erased

docs/src/mcu_rt_identity_attestation.md
```diff
@@ -0,0 +1,187 @@
...
+
+- **Keeping Attestation scoped to MCU** — `CertifyKey` and `Sign` operations on the default handle derive keys solely from the MCU RT context and its ancestors in PL0 PAUSER locality. SoC firmware contexts, if any, are derived as children in different (PL1 PAUSER) locality and are excluded from key derivation.
+
+- **Hitless MCU RT updates** — Caliptra RT updates the MCU RT TCI directly using its privileged access to the DPE context array (similar to RTMR). The default handle remains stable at `[0x00; 16]` across updates.
+
+- **SoC firmware isolation** — The MCU derives SoC firmware contexts using RETAIN_PARENT and CHANGE_LOCALITY to PL1, while keeping the MCU RT context unchanged in PL0. These SoC firmware contexts are optional, created as child contexts in PL1 PAUSER locality, and do not affect attestation key derivation.
+
+Creating the MCU RT DPE context requires invoking `DeriveContextCmd` with both a **measurement** (the SHA-384 digest of the MCU RT image) and an **SVN** (the security version number of the firmware). The measurement is already available from Caliptra RT's image authentication step, but the SVN presents a sourcing challenge described below.
```
zhalvorsen: Does this mean we want to skip stashing at AuthorizeAndStash?

docs/src/mcu_rt_identity_attestation.md
swenson: @fdamato it's a 16-bit value that we put in the manifest, the fuse can be any number of fuses. If the SVN is 7, you write 0x0007 here, you don't write 0x007f / 0b1111111 / 127, even though that is what would be in fuses. Someone could allocate 65,536 fuses if they wanted, but we would still only need to record 16 bits to capture the value.

docs/src/mcu_rt_identity_attestation.md
```diff
@@ -0,0 +1,187 @@
...
+
+3. **SVN field encoding** — OEM tooling must populate `AuthManifestPreamble.svn` as:
+   ```
+   preamble.svn = (mcu_rt_fw_svn << 16) | auth_manifest_svn
+   ```
+   where `auth_manifest_svn` is the value enforced against Caliptra Core fuses and `mcu_rt_fw_svn` is the MCU runtime firmware security version.
+
+### Boot Flow
+
+```
```
zhalvorsen: nit: putting this is a codeblock makes reading it in doc format a bit awkward because it doesn't wrap lines so you have to scroll sideways on some lines

docs/src/mcu_rt_identity_attestation.md
```diff
@@ -0,0 +1,187 @@
...
+
+---
+
+## Required Implementation Changes
+
+1. **No changes to Auth Manifest tooling or structure** — The Auth Manifest generation tooling and binary structure require no modifications. Only the input values differ — specifically, the `svn` field is populated with the partitioned encoding described above.
+
+2. **Caliptra RT (`caliptra-sw`)** — During the Set Auth Manifest flow, Caliptra RT must check the manifest version. If version >= 3, it enforces anti-rollback using only `preamble.svn[15:0]` against Caliptra Core fuses and stores `preamble.svn[31:16]` (MCU RT FW SVN) in persistent data for later use during `DeriveContextCmd`.
+
+3. **MCU-side DPE context creation (`caliptra-mcu-sw`)** — When creating the MCU RT DPE context during recovery boot after Caliptra RT authenticates the image:
```
zhalvorsen: These steps should be done by Caliptra RT not MCU

PR #1024, opened 2026-03-09
add evidence retrieval stage in e2e attestation flow

### LINKED ISSUE #991: [ATTESTATION] Extract Evidence in EAT format
Extract evidence from Caliptra device in EAT format

parvathib:
Fixes #991 
Runs an SPDM requester that issues GET_MEASUREMENTS with measurement index 0xFD.
Uses the nonce from the SPDM_NONCE environment variable in the request.
Writes the certificate chain (slot 0 only) and the resulting EAT to files for remote verification.


emulator/app/src/tests/spdm_responder_validator/mctp.rs
```diff
@@ -195,3 +196,52 @@ pub fn run_mctp_spdm_conformance_test(
...
+    });
+
+    thread::spawn(move || {
+        wait_for_runtime_start();
+
+        if !MCU_RUNNING.load(Ordering::Relaxed) {
+            exit(-1);
+        }
+        let listener =
+            TcpListener::bind("127.0.0.1:2323").expect("Could not bind to the SPDM listerner port");
```
fdamato: np: I`ll be careful to map to a specific port in the code, in some machine it can conflict with some other service... perhaps it should come from a configuration file

Issue #1027, opened 2026-03-10
Implement PROT_CAP OCP Recovery Data Field


ternst-nv:
The OCP Recovery Spec: https://www.opencompute.org/documents/ocp-recovery-document-1p1-final-pdf, defines a number of memory blocks required for managing recovery.  This issue should create a common library for defining those structures, and implement the first of these PROT_CAP, which defines the capabilities available on the device.

PR #1031, opened 2026-03-10
ocp: Add ocp PROT_CAP protocol (#1027)

ternst-nv:
Add the basis of the OCP Protocol in an ocp crate.  This includes an error reporting structure as well as a definitionof the protocol command numbers.

The spec for the data structures can be found here: https://www.opencompute.org/documents/ocp-recovery-document-1p1-final-pdf

### DIFF
--- a/common/ocp/src/protocol/prot_cap.rs
+++ b/common/ocp/src/protocol/prot_cap.rs
@@ -7,14 +7,18 @@
 //! Device's recovery protocol version, capabilities, and timing parameters.
 
 use bitfield::bitfield;
+use zerocopy::{Immutable, IntoBytes};
 
 use crate::error::OcpError;
 
 /// Expected magic string in bytes 0-7: "OCP RECV" in ASCII.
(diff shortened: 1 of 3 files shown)


common/ocp/src/protocol/prot_cap.rs
```diff
@@ -0,0 +1,236 @@
...
+
+use bitfield::bitfield;
+
+use crate::error::OcpError;
+
+/// Expected magic string in bytes 0-7: "OCP RECV" in ASCII.
+pub const MAGIC: [u8; 8] = [0x4F, 0x43, 0x50, 0x20, 0x52, 0x45, 0x43, 0x56];
+
+/// Wire size of a PROT_CAP response in bytes.
+pub const RESPONSE_LEN: usize = 15;
```
zhalvorsen: It would be good to make sure this is what is expected with a compile time assertion:

```Rust
let _: () = assert!(RESPONSE_LEN == size_of::<ProtCap>());
```

PR #1032, opened 2026-03-10
Add BootFSM timeout to ROM

### LINKED ISSUE #677: [rom] implement Caliptra Core boot timeout in MCU ROM
To improve reliability we should implement a boot timeout mechanism in the MCU ROM that monitors the boot FSM status of Caliptra Core to ensure it reaches execution of the first ROM instruction and timeout with an error otherwise.

amit-kumarh:
Fixes #677

and adds an integration test for this timeout (FPGA only, since we're using the bootfsm_break parameter to halt the FSM, which isn't yet supported by the emulator

dependent on https://github.com/chipsalliance/caliptra-sw/pull/3462


swenson (approved):
LGTM;

I think it wouldn't be too bad to add support for this test in the emulator, where we just halt the Caliptra core CPU and resume it based on some condition, and set the appropriate register, but maybe in a follow-up.

timothytrippel:
looks like you need to rebase then we can merge

PR #1046, opened 2026-03-11
emulator: Add Rom Sideload Support

chrisherczeg:
## Summary

Add support for sideloading MCU ROM images into the emulator before stepping the core.

When the ROM is built into an ELF file (e.g., containing a trampoline to runtime), this allows the ELF to be left intact and loaded via an external ELF loader rather than being extracted and passed in separately.


swenson (approved):
LGTM

PR #1055, opened 2026-03-12
ocp: recovery command structures

ternst-nv:
Add the command structures for the device reset and recovery control/status memory blocks, along with appropriate unit tests.

### DIFF
--- a/common/ocp/src/protocol/recovery_status.rs
+++ b/common/ocp/src/protocol/recovery_status.rs
@@ -101,9 +101,8 @@ impl RecoveryStatus {
     }
 
     /// Byte 0, bits 0-3: Device recovery status.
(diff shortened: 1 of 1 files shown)


common/ocp/src/protocol/recovery_status.rs
```diff
@@ -0,0 +1,206 @@
...
+        Ok(Self {
+            byte0,
+            vendor_status,
+        })
+    }
+
+    /// Byte 0, bits 0-3: Device recovery status.
+    pub fn status(&self) -> DeviceRecoveryStatus {
+        DeviceRecoveryStatus::try_from(self.byte0.status())
+            .expect("constructed with a valid status")
```
zhalvorsen: Is there a reason you aren't using a Result here?

PR #1057, opened 2026-03-12
Make xtask depend on mcu-rom-common only for fpga_realtime

chrisherczeg:
## Summary

Make `mcu-rom-common` optional in `xtask`.

## Details

`mcu-rom-common` is only needed for `fpga_realtime`, so this change:
- marks it as an optional dependency
- enables it through the `fpga_realtime` feature

This avoids pulling it in for `xtask` configurations that do not use that feature.


swenson (approved):
LGTM

PR #1058, opened 2026-03-12
Support extended external interrupt range (0-255] in emulator C binding

chrisherczeg:
This change updates emulator_set_external_interrupt to accept IRQ numbers in the range 1–255 (previously 0–31), and rejects IRQ 0 as invalid. This aligns the C binding with hardware that supports a wider set of external interrupt lines, where interrupt 0 is reserved/unused


swenson (approved):
LGTM

PR #1062, opened 2026-03-12
Fix lc_clk_byp_req_o port connection in FPGA wrapper

zahralak:
Disconnecting output port from an external driver. 

@swenson


swenson (approved):
LGTM

PR #1091, opened 2026-03-17
emulator: Support lifecycle controller transitions

### LINKED ISSUE #842: emulator: Lifecycle controller should support transitions
Most of the logic for doing the cSHAKE128 is already present in caliptra-sw: https://github.com/chipsalliance/caliptra-sw/blob/main/hw-model/src/otp_provision.rs#L572

So it would be a matter of:

* Moving that code to the right place so the MCU emulator can use it
* Hooking up the Lifecycle controller emulator peripheral to the OTP peripheral
* Implement the LC registers for the transition command and accepting the token
* Using that cSHAKE128 to check the passed in token against the register
* Updating the lifecycle state and fuses on success.
* Setting the appropriate registers on failure
* Adding tests for these scenarios
* Updating any existing tests

Additional references:
* https://opentitan.org/book/hw/ip/otp_ctrl/
* https://opentitan.org/book/hw/ip/lc_ctrl/index.html

swenson:
Implement LC state transition logic in the emulator LcCtrl peripheral with token verification (cSHAKE128), writing to OTP fuses, and comprehensive integration tests for all transitions on both emulator and FPGA.

I did notice on FPGA that I was not able to transition the lifecycle controller from the hw-model directly (probably due to AXI user permissions locked to MCU only), so a test ROM is created to facilitate the lifecycle transitions.

Tested as working locally on my FPGA, and added to the FPGA CI tests.

Merging blocked on https://github.com/chipsalliance/caliptra-sw/pull/3497 for supporting more flexible lifecycle states on the Caliptra hw-model.

Fixes #842

### DIFF
--- a/platforms/emulator/runtime/userspace/apps/user/src/spdm/device_measurements/ocp_eat/claims.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/spdm/device_measurements/ocp_eat/claims.rs
@@ -15,7 +15,7 @@ use ocp_eat::ocp_profile::{
 };
 use ocp_eat::{
     ClassIdTypeChoice, ClassMap, ConciseEvidence, ConciseEvidenceMap, DigestEntry, EnvironmentMap,
-    EvTriplesMap, EvidenceTripleRecord, MeasurementMap, MeasurementValue, TaggedBytes,
+    EvTriplesMap, EvidenceTripleRecord, MeasurementMap, MeasurementValue, TaggedBytes, VersionMap,
 };
 use spdm_lib::measurements::{MeasurementsError, MeasurementsResult};
 use zerocopy::IntoBytes;
@@ -204,7 +204,10 @@ pub async fn generate_claims(claims_buf: &mut [u8], nonce: &[u8]) -> Measurement
         measurement_maps[i] = MeasurementMap {
             key: i as u64,
             mval: MeasurementValue {
-                version: Some(&version_fields[i].buf),
+                version: Some(VersionMap {
+                    version: &version_fields[i].buf,
+                    version_scheme: None,
+                }),
                 svn: Some(svns[i] as u64),
                 digests: Some(&digest_entries_arr[i]),
                 integrity_registers: Some(&integrity_registers_arr[i]),


emulator/periph/src/lc_ctrl.rs
```diff
@@ -9,29 +9,155 @@ File Name:
...
+    let mut output = [0u8; 16];
+    hasher.finalize_xof_into(&mut output);
+    output
+}
+
+/// What token (if any) a valid transition requires.
+enum TokenRequirement {
+    /// No token needed (unconditional transition).
+    None,
+    /// Raw unlock token (hardcoded, not from OTP).
```
timothytrippel: nit:
```suggestion
    /// Raw unlock token (netlist constant, not from OTP).
```

hw/model/test-fw/lc-raw-unlock/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+    // Wait for go-bit.
+    while env.mci.registers.mci_reg_generic_input_wires[0].get() & GO_BIT == 0 {}
+
+    let wires = env.mci.registers.mci_reg_generic_input_wires[0].get();
+    let target_index = (wires >> 8) & 0x1F;
+    let token_type = (wires >> 15) & 0x3;
+    let expect_error = (wires >> 17) & 1 != 0;
+
+    // Wait for LC controller to be ready.
+    while !regs.status.is_set(lc_ctrl::bits::Status::Initialized)
```
timothytrippel: might be nice to wrap all the ops in a lib like https://github.com/chipsalliance/caliptra-mcu-sw/blob/ddc2f0f96a6cbc6c7ef56b24050c708906969c51/hw/model/src/lcc.rs#L64 did for doing LC transitions over JTAG.

hw/model/test-fw/lc-raw-unlock/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+    regs.transition_token_0.set(token[0]);
+    regs.transition_token_1.set(token[1]);
+    regs.transition_token_2.set(token[2]);
+    regs.transition_token_3.set(token[3]);
+
+    // Trigger transition.
+    regs.transition_cmd.set(1);
+
+    // Poll for result.
+    let success = loop {
```
timothytrippel: should you also check the final LC state? (after a reset)

emulator/periph/src/lc_ctrl.rs
```diff
@@ -41,34 +167,573 @@ impl Default for LcCtrl {
...
     fn read_lc_transition_cnt(
         &mut self,
     ) -> ReadWriteRegister<u32, lc_ctrl::bits::LcTransitionCnt::Register> {
         ReadWriteRegister::new(self.lc_transition_cnt)
     }
+
+    fn read_claim_transition_if(
+        &mut self,
+    ) -> ReadWriteRegister<u32, lc_ctrl::bits::ClaimTransitionIf::Register> {
+        let val = if self.mutex_claimed { MUTEX_TRUE } else { 0 };
```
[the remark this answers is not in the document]
swenson: Great catch, thanks!

timothytrippel (approved):

PR #1108, opened 2026-03-20
Add default PK Hash selection logic to ROM

amit-kumarh:
Implement a default key policy that selects the first key slot that is marked valid and contains unrevoked ECC and PQC keys.

This does add some code size due to the addition of new fuse constants to the generated list. I plan to explore refactoring some of this out in a follow-up, but had to include it here so that the FuseEntries didn't contain incorrect fuse encoding schemes.

### DIFF
--- a/tests/integration/src/rom/mod.rs
+++ b/tests/integration/src/rom/mod.rs
@@ -5,6 +5,7 @@ mod test_hitless_update;
 mod test_lc_ctrl;
 mod test_otp_blank_check;
 mod test_sw_digest_lock;
+mod test_vendor_key_policy;
 mod test_warm_reset;
 
 // testing this requires enabling the BootFSM breakpoint which is only implemented
--- a/tests/integration/src/rom/test_vendor_key_policy.rs
+++ b/tests/integration/src/rom/test_vendor_key_policy.rs
@@ -1,0 +1,249 @@
+// Licensed under the Apache-2.0 license
+
+#[cfg(test)]
+mod test {
+    use anyhow::Result;
+    use mcu_error::McuError;
+    use mcu_hw_model::{new, DefaultHwModel, InitParams, McuHwModel};
+    use mcu_rom_common::{
+        pqc_key_type_entry, vendor_ecc_revocation_entry, vendor_lms_revocation_entry,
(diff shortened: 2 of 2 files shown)


hw/fuses.hjson
```diff
@@ -43,36 +40,404 @@
...
+        name: "vendor_mldsa_revocation_0",
+        bits: 4,
+        description: "Revocation bits for MLDSA keys in slot 0.",
+        partition: "VENDOR_REVOCATIONS_PROD_PARTITION",
+        otp_item: "CPTRA_CORE_MLDSA_REVOCATION_0",
+        layout: {type: "LinearMajorityVote", duplication: 3},
+    },
+    {
+        name: "vendor_lms_revocation_0",
+        bits: 16,
```
timothytrippel: Again dont we want to use all 32 bits? or am I miss interpretting how this field is parsed by the tooling?

rom/src/fuses.rs
```diff
@@ -135,3 +137,58 @@ impl Default for McuFuseLayoutPolicy {
...
+
+impl Default for DefaultVendorKeyPolicy {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl VendorKeyPolicy for DefaultVendorKeyPolicy {
+    fn select_key(&self, otp: &Otp) -> McuResult<usize> {
+        let valid_mask = otp.read_entry(generated_fuses::VENDOR_PK_HASH_VALID)?;
```
amit-kumarh: this is mistakenly assuming the VALID field is 1 word. Need to fix

RaunakGu:
Hi @amit-kumarh ,

Thanks for the reply, Yes, definitely we need to sync the document.
but before that,
We have a case where PK Hash not required to be provisioned when leaves Vendor Manufacturing and in field (PROD State) it could be still provisioned, 
In this case considering 0 as 'Valid' will impact the selection logic, we may end up selecting a non-provisioned key as Valid, right?
image

amit-kumarh:
In this case, all of the remaining unprovisioned slots will be above the selected one due to the key policy,

The first key slot is required to be provisioned in manufacturing (in the HASHES_MANUF partition), so we should always begin in a valid state.

If we, in the field, invalidate all of the key slots without provisioning any more, then we will try and load a key slot of all 0s, which would be an error case when it doesn't match what's in the FW preamble.

PR #1110, opened 2026-03-23
ocp: usb packet definitions

ternst-nv:
Add support for the Setup and Configuration packets which are required for OCP USB Recovery.  This is not intended to be a comprehensive implementation of the USB spec, but focuses solely on those elements required for OCP Recovery to function.  It also provides utility methods for generating the OCP Configuration blob, with parameters for device specific values.

### DIFF
--- a/common/ocp/src/usb/descriptors.rs
+++ b/common/ocp/src/usb/descriptors.rs
@@ -1,0 +1,556 @@
+// Licensed under the Apache-2.0 license
+
+//! USB descriptor structures for OCP Recovery device enumeration.
+//!
+//! Defines the minimal set of USB 2.0 descriptors required to enumerate
+//! a composite device exposing the OCP Secure Firmware Recovery interface
+//! on EP0 (OCP spec Section 8.5).
+//!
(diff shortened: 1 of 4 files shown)


common/ocp/src/usb/descriptors.rs
```diff
@@ -0,0 +1,511 @@
...
+    /// Total length of all descriptors in this configuration.
+    pub w_total_length: u16,
+    /// Number of interfaces in this configuration.
+    pub b_num_interfaces: u8,
+    /// Value used by SET_CONFIGURATION to select this configuration. Must be nonzero.
+    pub b_configuration_value: u8,
+    /// Index of string descriptor for this configuration, or 0.
+    pub i_configuration: u8,
+    /// Configuration attributes. Bit 7 must be 1; bit 6 = self-powered; bit 5 = remote wakeup.
+    pub bm_attributes: u8,
```
clundin25: Does it make sense to use a bitflag here?

https://docs.rs/bitflag/latest/bitflag/

PR #1123, opened 2026-03-25
Set I3C CmdEmptyBufThld to 1 (spec minimum)

### LINKED ISSUE #946: [Bug] Invalid `CMD_EMPTY_BUF_THLD` value in HCI queue threshold setup
In `rom/src/i3c.rs`, the HCI queue threshold configuration sets `CmdEmptyBufThld` to `0`:

```rust
regs.piocontrol_queue_thld_ctrl.modify(
    QueueThldCtrl::CmdEmptyBufThld.val(0)
        + QueueThldCtrl::RespBufThld.val(1)
        + QueueThldCtrl::IbiStatusThld.val(1),
);
```

According to the [QUEUE_THLD_CTRL register spec](https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.I3CCSR.PIOControl.QUEUE_THLD_CTRL), the `CMD_EMPTY_BUF_THLD` field (bits [7:0]) accepts values **1–255**. A value of `0` is out of range and invalid per the spec. This field controls when the `CMD_QUEUE_READY_STAT` interrupt fires (when the command queue has N or more free entries), so setting it to 0 means the threshold interrupt behavior is undefined. The fix should be to set this value to at least `1`. 

### Typo in timing register comment

In the same file (`rom/src/i3c.rs`), line 59 has an incorrect comment:

```rust
regs.soc_mgmt_if_t_f_reg.set(clocks); // rise time of both SDA and SCL in clock units
```

The `t_f` register represents the **fall** time, not the rise time. The comment should read:

```rust
regs.soc_mgmt_if_t_f_reg.set(clocks); // fall time of both SDA and SCL in clock units
```

giladliber:
QUEUE_THLD_CTRL.CMD_EMPTY_BUF_THLD was set to 0, which is outside the valid range (1–255) per the I3C [HW spec](https://chipsalliance.github.io/caliptra-ss/main/regs/?p=soc.I3CCSR.PIOControl.QUEUE_THLD_CTRL). Changed to 1 to match the register reset default. 
Also fixed a copy-paste comment typo on t_f_reg ("rise time" → "fall time").

Fixes #946

### DIFF
--- a/rom/src/i3c.rs
+++ b/rom/src/i3c.rs
@@ -60,7 +60,7 @@ impl I3c {
         // TODO: pass this timing information in
         let clocks = 0;
         regs.soc_mgmt_if_t_r_reg.set(clocks); // rise time of both SDA and SCL in clock units
-        regs.soc_mgmt_if_t_f_reg.set(clocks); // rise time of both SDA and SCL in clock units
+        regs.soc_mgmt_if_t_f_reg.set(clocks); // fall time of both SDA and SCL in clock units
 
         // if this is set to 6+ then ACKs start failing
         regs.soc_mgmt_if_t_hd_dat_reg.set(clocks); // data hold time in clock units


giladliber:
@clundin25  - can you please take a look on that small fix? Thanks!

Issue #905, opened 2026-02-17
[rom] develop mailbox commands to facilitate provisioning fuse programming


timothytrippel:
Early in the manufacturing lifecycle (when a device is in `TEST_UNLOCKED*` or `MANUF` lifecycle states, it would be convenient to expose a set of mailbox commands in the MCU ROM to facilitate fuse programming for the subset of partitions that should be programmed in these LC states (see https://chipsalliance.github.io/caliptra-mcu-sw/provisioning.html#provisioning-sequence-1).

timothytrippel:
Thanks @taytek-mrvl !

PR #1126, opened 2026-03-25
romtime: Add files for otp provisioning mailbox calls.

cdupuis-mrvl:
This commit adds support for the MC_FUSE_READ, MC_FUSE_WRITE and MC_FUSE_LOCK_PARTITION MCI mailbox opcodes as specified in https://chipsalliance.github.io/caliptra-mcu-sw/external_mailbox_cmds.html.  This is the start of the implementation of RFC #905: https://github.com/chipsalliance/caliptra-mcu-sw/issues/905 by adding the necessary mailbox opcode primitives.

This pull request only implements the actual handlers for the mailbox commands but does not currently integrate them into either the rom or runtime firmware flows.

Integration tests with the software emulator are still pending.


romtime/src/mci_mbox_cmd.rs
```diff
@@ -0,0 +1,521 @@
...
+// iterating over arbitrarily large SRAM payloads during checksum verification)
+// ---------------------------------------------------------------------------
+const MAX_DLEN_FUSE_READ: u32 = 12; // chksum + partition + entry
+const MAX_DLEN_FUSE_LOCK: u32 = 8; // chksum + partition
+const MAX_DLEN_FUSE_WRITE: u32 = 20 + (MAX_FUSE_DATA_WORDS * 4) as u32; // header + data
+
+// ---------------------------------------------------------------------------
+// Mailbox register bundle (works for either mbox0 or mbox1)
+// ---------------------------------------------------------------------------
+
```
timothytrippel: nit: remove newline

romtime/src/mci_mbox_cmd.rs
```diff
@@ -0,0 +1,521 @@
...
+// ---------------------------------------------------------------------------
+// Checksum helpers operating directly on SRAM registers
+// ---------------------------------------------------------------------------
+
+/// Verify the caliptra-style request checksum.
+///
+/// Layout: SRAM\[0\] = chksum, SRAM\[1..\] = payload.
+/// chksum == calc_checksum(cmd, LE-bytes-of(SRAM\[1..dlen/4\]))
+fn verify_request_checksum(
```
timothytrippel: Could we split the summing of bytes of a u32 into a helper function to make this a bit more readable, like:

```rust
fn sum_u32_bytes(val: u32, count: usize) -> u32 {
    val.to_le_bytes().iter().take(count).map(|&b| b as u32).sum()
}

fn verify_request_checksum(
    cmd: u32,
    sram: &[tock_registers::registers::ReadWrite<u32>],
    dlen: u32,
) -> bool {
    if dlen < 4 {
        return false;
    }
    let payload_len = (dlen - 4) as usize;
    let full_words = payload_len / 4;
    let tail_bytes = payload_len % 4;

    let mut sum = sum_u32_bytes(cmd, 4);

    for reg in sram.iter().skip(1).take(full_words) {
        sum = sum.wrapping_add(sum_u32_bytes(reg.get(), 4));
    }

    if tail_bytes > 0 {
        sum = sum.wrapping_add(sum_u32_bytes(sram[1 + full_words].get(), tail_bytes));
    }

    sum.wrapping_add(sram[0].get()) == 0
}

fn write_response_checksum(sram: &[tock_registers::registers::ReadWrite<u32>], resp_words: usize) {
    let sum: u32 = sram
        .iter()
        .take(resp_words)
        .skip(1)
        .map(|reg| sum_u32_bytes(reg.get(), 4))
        .sum();
    sram[0].set(0u32.wrapping_sub(sum));
}
```

romtime/src/mci_mbox_cmd.rs
```diff
@@ -0,0 +1,508 @@
...
+            MC_FUSE_READ_CMD => handle_fuse_read(cmd, input_dlen, &active_mbox, otp),
+            MC_FUSE_WRITE_CMD => handle_fuse_write(cmd, input_dlen, &active_mbox, otp),
+            MC_FUSE_LOCK_PARTITION_CMD => {
+                handle_fuse_lock_partition(cmd, input_dlen, &active_mbox, otp)
+            }
+            _ => handle_unknown_cmd(cmd, &active_mbox),
+        }
+
+        // Wait for SoC to release the mailbox (MBOX_EXECUTE → 0).
+        crate::println!(
```
zhalvorsen: You could do something similar to https://github.com/chipsalliance/caliptra-sw/blob/a05a9075ed2f91224b126137b8dbc739520020d0/runtime/src/lib.rs#L586 to populate the response checksum so you don't have to do it in each handler as well.

error/src/lib.rs
```diff
@@ -242,6 +242,71 @@ impl McuError {
             0x3_000c,
             "OTP failure in OCP LOCK"
         ),
+        (
+            ROM_OTP_FUSE_INVALID_PARTITION,
+            0x3_0010,
```
zhalvorsen, line 247: Any particular reason this skips `0x3_000d` - `0x3_000f`?

error/src/lib.rs
```diff
@@ -242,6 +242,71 @@ impl McuError {
...
+            0x3_001a,
+            "OTP fuse provisioning: data too large"
+        ),
+        (
+            ROM_OTP_FUSE_INPUT_TOO_SHORT,
+            0x3_001b,
+            "OTP fuse provisioning: input too short"
+        ),
+        (
+            ROM_OTP_FUSE_UNKNOWN_COMMAND,
```
zhalvorsen, line 306: We may want to add more commands to the ROM mailbox. Can you make this name more generic?

romtime/src/mci_mbox_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+            size_of::<u32>(),
+            MAX_FUSE_REQ_BYTES
+        );
+        return Err(McuError::ROM_OTP_FUSE_INVALID_LENGTH);
+    }
+
+    sram_to_buf(sram, buf, input_dlen);
+
+    let chksum = u32::from_le_bytes(buf[..4].try_into().unwrap());
+    if !verify_checksum(chksum, cmd, &buf[size_of::<u32>()..input_dlen]) {
```
zhalvorsen, line 125: Just something to keep an eye on. This line might panic which will cause some failures in CI. You can use `.get` logic to ensure it can't panic.

romtime/src/mci_mbox_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+    crate::println!(
+        "[mci-mbox] IFPR: partition={}, entry={}",
+        HexWord(partition),
+        entry
+    );
+
+    let params = fuse_read_dai_params(partition, entry, MAX_FUSE_DATA_WORDS)?;
+
+    // Response: [chksum(4)][fips_status(4)][length_bits(4)][data…]
+    buf[4..8].copy_from_slice(&MailboxRespHeader::FIPS_STATUS_APPROVED.to_le_bytes());
```
zhalvorsen, line 174: Because this is currently hardcode, feel free to do this logic in `send_response`. Also, using `zerocopy` for these assignments can make it easier to read. It can become self documenting code.

```Rust

        let hdr: &mut MailboxRespHeader = MailboxRespHeader::mut_from_bytes(
            &mut mut_resp_bytes[..size_of::<MailboxRespHeader>()],
        )
        .map_err(|_| McuError::ROM_OTP_FUSE_INVALID_LENGTH)?;

        // Set the chksum field
        hdr.chksum = checksum;
        hdr.fips_status = MailboxRespHeader::FIPS_STATUS_APPROVED;
```

romtime/src/mci_mbox_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+    buf[8..12].copy_from_slice(&params.valid_bits.to_le_bytes());
+    for i in 0..params.words_to_read {
+        match otp.read_word(params.base_word_addr + i) {
+            Ok(word) => {
+                buf[12 + i * 4..16 + i * 4].copy_from_slice(&word.to_le_bytes());
```
zhalvorsen, line 179: zerocopy will make these assignments easier as well. If you use a u32 array in the response struct, this line would become:

```Rust
params.words[i] = word;
```

romtime/src/mci_mbox_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+        length
+    );
+
+    let data_bytes = length
+        .checked_add(7)
+        .map(|v| (v / 8) as usize)
+        .ok_or_else(|| {
+            crate::println!("[mci-mbox] IFPW: length overflow");
+            McuError::ROM_OTP_FUSE_INVALID_LENGTH
+        })?;
```
zhalvorsen, line 230: It could be helpful to have a comment here saying that you are rounding up to the nearest multiple of 8? Alternatively, you could use a Rust u32 function https://doc.rust-lang.org/std/primitive.u32.html#method.next_multiple_of:

```suggestion
    let data_bytes = length.next_multiple_of(8);
```

romtime/src/mci_mbox_cmd.rs
```diff
@@ -0,0 +1,404 @@
...
+// Receives fuse provisioning commands from MCI mailbox CSRs, dispatches to the
+// DAI-backed handlers in otp_provision, and writes caliptra-compatible responses.
+//
+// Checksum verification and response formatting are centralized in the dispatch
+// loop (process_fuse_mbox_commands), following the patterns established in
+// caliptra-sw runtime and mcu-mbox-lib transport.
+
+use crate::otp_provision::{fuse_lock_partition_dai, fuse_read_dai_params, fuse_write_dai};
+use crate::{HexWord, Mci, Otp};
+use caliptra_api::mailbox::{populate_checksum, MailboxReqHeader, MailboxRespHeader};
```
zhalvorsen, line 14: FYI: `MailboxReqHeader` and `MailboxRespHeader` are both redefined in `mcu_mbox_common::messages::` It would probably be good to use those versions so they stay conventional with the rest of this repo

timothytrippel:
@cdupuis-mrvl could you squash your changes into a single commit? also has all feedback been addressed? can we move forward with this?

romtime/src/otp_provision.rs
```diff
@@ -0,0 +1,449 @@
...
+pub const PARTITION_VENDOR_SECRET_PROD: u32 = 0x0000_000D;
+pub const PARTITION_VENDOR_NON_SECRET_PROD: u32 = 0x0000_000E;
+pub const PARTITION_CPTRA_SS_LOCK_HEK_PROD_0: u32 = 0x0000_000F;
+pub const PARTITION_CPTRA_SS_LOCK_HEK_PROD_1: u32 = 0x0000_0010;
+pub const PARTITION_CPTRA_SS_LOCK_HEK_PROD_2: u32 = 0x0000_0011;
+pub const PARTITION_CPTRA_SS_LOCK_HEK_PROD_3: u32 = 0x0000_0012;
+pub const PARTITION_CPTRA_SS_LOCK_HEK_PROD_4: u32 = 0x0000_0013;
+pub const PARTITION_CPTRA_SS_LOCK_HEK_PROD_5: u32 = 0x0000_0014;
+pub const PARTITION_CPTRA_SS_LOCK_HEK_PROD_6: u32 = 0x0000_0015;
+pub const PARTITION_CPTRA_SS_LOCK_HEK_PROD_7: u32 = 0x0000_0016;
```
zhalvorsen, line 42: Can we make this an enum?

romtime/src/otp_provision.rs
```diff
@@ -0,0 +1,449 @@
...
+    pub valid_bits: u32,
+}
+
+// Validates a fuse-read request and returns the parameters needed to perform
+// the actual DAI word reads.  No data is read here — the caller drives the
+// read loop so it can target any destination (stack buffer, SRAM registers, …).
+//
+// `entry` is a *byte offset* within the partition (must be word-aligned).
+// `max_words` caps the number of words that will be read.
+
```
zhalvorsen, line 195: Lets convert the public function comments to doc comments:


```suggestion
/// Validates a fuse-read request and returns the parameters needed to perform
/// the actual DAI word reads.  No data is read here — the caller drives the
/// read loop so it can target any destination (stack buffer, SRAM registers, …).
///
/// `entry` is a *byte offset* within the partition (must be word-aligned).
/// `max_words` caps the number of words that will be read.
```

cdupuis-mrvl:
> @cdupuis-mrvl could you squash your changes into a single commit? also has all feedback been addressed? can we move forward with this?

There's a few more comments we need to address that we just got from @zhalvorsen so we need to address those.  Otherwise everything previous has been caught up.

timothytrippel:
Sg, ping me when these have been addressed and I will get this approved/merged.

cdupuis-mrvl:
> Sg, ping me when these have been addressed and I will get this approved/merged.

Will do.

error/src/lib.rs
cdupuis-mrvl: We had initially wanted to leave room since we were starting a new error subtype.  We made them consecutive in fc64f22.

error/src/lib.rs
cdupuis-mrvl: Sure, we changed it to ROM_MCI_MBOX_UNKNOWN_COMMAND in fc64f22.

romtime/src/mci_mbox_cmd.rs
cdupuis-mrvl: Sure we fixed this up in fc64f22.  Now we replaced both direct index operations with .get():. If buf is too short for either slice, the code now returns Err(McuError::ROM_OTP_FUSE_INVALID_LENGTH) instead of panicking. The remaining .unwrap() on try_into() is safe because .get(..4) guarantees exactly 4 bytes when it returns Some.

romtime/src/mci_mbox_cmd.rs
cdupuis-mrvl: Moved the FIPS status assignment into send_response using MailboxRespHeader::mut_from_bytes(), so it's set once centrally instead of being hardcoded in each handler. Removed the three duplicated buf[4..8].copy_from_slice(...) lines in fc64f22.

romtime/src/mci_mbox_cmd.rs
cdupuis-mrvl: Added a FuseReadResp struct and used mut_from_bytes() so the response is built via named fields (resp.valid_bits, resp.data[i] = word) instead of raw byte offsets. Also converted the remaining buf[4..8] slicing in send_response and the error path to use MailboxRespHeader::mut_from_bytes() in fc64f22.

romtime/src/mci_mbox_cmd.rs
cdupuis-mrvl: Thanks for the suggestion. Just a heads up -- next_multiple_of(8) would round up to the nearest multiple of 8 without dividing, so it wouldn't quite work here since we're converting bits to bytes (ceiling division by 8). We went with length.div_ceil(8) instead, which reads much better than the manual (length + 7) / 8 and also sidesteps the overflow concern without needing checked_add.

romtime/src/otp_provision.rs
cdupuis-mrvl: Yes, changed to an enum in fc64f22.

romtime/src/otp_provision.rs
cdupuis-mrvl: Sure, converted to doc comments in fc64f22.

cdupuis-mrvl:
> > Sg, ping me when these have been addressed and I will get this approved/merged.
> 
> Will do.

Hi @timothytrippel.  I've addressed the comments from @zhalvorsen.  I can squash this into once patch when we want to move forward.

zhalvorsen (approved):

PR #1144, opened 2026-03-30
rom: Add firmware manifest DOT command processing (2.0)

### LINKED ISSUE #886: Device Ownership Transfer: Add firmware manifest command flows
Reading from the MCU firmware, check for a **optional** header/manifest that contains instructions on any fuses to burn or device ownership transfer commands to execute. This allows DOT commands to be executed as part of firmware updates, if an integrator chooses to do so. (Some prefer this, some don't.)

This can happen in the ROM or at the beginning of runtime (or an FMC).

This should support basic DOT commands (UNLOCK, LOCK, ROTATE, DISABLE), and they should be idempotent (so if the command does not apply, then it is ignored).

swenson:
Add support for embedding DOT commands (LOCK, UNLOCK, ROTATE, DISABLE) in an optional section within the MCU firmware image header, allowing ownership transfer state changes to be executed during cold boot as part of firmware updates. All commands are idempotent.

Key changes:
- Add FwManifestDotSection struct and process_fw_manifest_dot_commands() to device_ownership_transfer.rs with support for 5 command types
- Add fw_manifest_dot_offset field to RomParameters for opt-in by platform integrators
- Integrate manifest processing in cold_boot.rs after firmware validation, with guards against encrypted boot (in-place decryption would corrupt the header)
- Refactor burn_dot_lock_fuse to accept &Otp for split-borrow compatibility
- Add test-fw-manifest-dot feature to emulator ROM
- Add custom_mcu_rom and firmware_prefix to test infrastructure
- Add 9 integration tests covering lock, unlock, rotate, disable, idempotency, magic skip, and bad version error paths

* rom: Fix HMAC checksum computation and error code collision

Fix two bugs that caused all DOT integration tests to fail:

1. cm_hmac checksum bug: The manual checksum computation added the command as a raw u32 value instead of summing its individual LE bytes (as calc_checksum does). This caused Caliptra to reject every CM_HMAC request with FW_PROC_MAILBOX_INVALID_CHECKSUM (0x01020005). Rewrite cm_hmac to build the request in a flat byte buffer and use calc_checksum + start_mailbox_req_bytes, matching the proven pattern used by cm_derive_stable_key.

2. Error code collision: ROM_COLD_BOOT_FW_MANIFEST_DOT_ERROR and ROM_COLD_BOOT_HEK_REPORT_ERROR both used 0x1_0016, causing the test_error_constants_uniqueness test to fail. Assign 0x1_0017 to ROM_COLD_BOOT_FW_MANIFEST_DOT_ERROR.

* tests: Fix DOT test infrastructure for full fw_manifest_dot support

- Add DOT flash write handler in emulator root_bus (was missing, causing Store/AMO access fault during FwBoot blob persistence)
- Add OTP flags to test-fw-manifest-dot ROM parameters
- Fix test fuse assertions to use DOT_FUSE_ARRAY.byte_offset instead of incorrect VENDOR_NON_SECRET_PROD_PARTITION_BYTE_OFFSET+1
- Cache ROM_FW_MANIFEST_DOT bytes eagerly to prevent the compiled binary from being overwritten by subsequent default ROM builds
- Remove stale HMAC cache that could return invalid HMACs across hardware models with different Caliptra keys

* Add DOT recovery mechanism

* rom: replace copy_from_slice with panic-free helpers and fix test ROM clobbering

Replace copy_from_slice calls in DOT mailbox helpers (cm_hmac, cm_derive_stable_key) and SimpleFlash::write with a copy_slice helper that returns an error on length mismatch instead of panicking.  The ROM binary must be provably panic-free; copy_from_slice contains an internal length check the compiler sometimes cannot optimise away.

Fix compile_rom() in test infrastructure to return the feature-suffixed ROM binary path directly from rom_build instead of copying to a generic mcu-rom-emulator.bin.  Subsequent build steps (CaliptraBuilder / SOC manifest) re-create that intermediate without the appended ROM digest, silently clobbering the copy and causing ROM digest integrity failures in tests that exercise warm reset.

Fixes #886 

---------

(cherry picked from commit 33b024af1b00909ef6ef14451c392b3212665999)


rom/src/device_ownership_transfer.rs
```diff
@@ -1048,3 +1048,314 @@ pub fn dot_recovery_challenge_flow(
...
+/// Creates, HMAC-seals, and writes a DOT blob to flash storage.
+///
+/// This is used by manifest DOT commands (LOCK, DISABLE, ROTATE) to persist
+/// ownership credentials alongside the fuse burn.
+fn create_and_seal_dot_blob(
+    env: &mut RomEnv,
+    dot_fuses: &DotFuses,
+    cak: &OwnerPkHash,
+    lak: &LakPkHash,
+    dot_flash: Option<&dyn crate::flash::hil::FlashStorage>,
```
mlvisaya: qq: should we make flash storage mandatory or return error if None is passed? if the integrator doesn't pass flash storage, they may change the fuse, but the blob won't be written

PR #1146, opened 2026-03-31
emulator mci: fix WDT interrupt assertion + add missing ERROR0 intr handlers

v-ramalleti:
What changed:

Make WDT timeout set ERROR0 internal interrupt status in MMIO-visible register, then recompute IRQ.
Add missing ERROR0 MMIO handlers: error0_intr_trig, error0_internal_intr (W1C), error0_intr_en, global_intr_en.
Add safe poll scheduling helpers (reschedule_poll/cancel_poll) to avoid stale ActionHandles.
Why:
Timer expiry was configured but interrupt wasn’t asserted/servable reliably; fix brings model behavior in line with expected WDT interrupt behavior.

Notes:

Unhandled interrupt 38 still appears in some test logs; likely separate from MCI ERROR0/WDT interrupt handling and not fixed here.

### DIFF
--- a/emulator/periph/src/mci.rs
+++ b/emulator/periph/src/mci.rs
@@ -28,7 +28,6 @@ pub struct Mci {
     ext_mci_regs: caliptra_emu_periph::mci::Mci,
     generated: MciGenerated,
 
-    error0_internal_intr_r: ReadWriteRegister<u32, Error0IntrT::Register>,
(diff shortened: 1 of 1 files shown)


swenson (approved):
LGTM; smoke test failures are unrelated as this is an emulator only change

PR #920, opened 2026-02-18
FPGA External I3C Host 2.1

jlmahowa-amd:
Adds runtime-configurable connection to allow the I3C to be driven by an external I3C host controller using a special adapter board.

PR #1154, opened 2026-03-31
FPGA use axi_reset for sram_req_if 2.1

jlmahowa-amd:


jlmahowa-amd:
Change went in with #920

PR #1157, opened 2026-03-31
ocp: simple commands

ternst-nv:
Expand the capabilities of the OCP state machine to handle a collection of simple requests for capabilities, device id, and device status.  As part of this, update the vendor and process command to better handle the integration.  Include a number of tests validating the functionality.

### DIFF
--- a/common/ocp/src/error.rs
+++ b/common/ocp/src/error.rs
@@ -55,7 +55,8 @@ pub enum OcpError {
     /// The value does not match a known USB Descriptor Type (Table 9-5).
     InvalidDescriptorType = 23,
     /// A transport-level error occurred (e.g. USB hardware failure).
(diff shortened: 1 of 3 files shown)


common/ocp/src/error.rs
```diff
@@ -52,6 +54,23 @@ pub enum OcpError {
     InvalidStandardRequest = 22,
     /// The value does not match a known USB Descriptor Type (Table 9-5).
     InvalidDescriptorType = 23,
+    /// A transport-level error occurred (e.g. USB hardware failure).
+    TransportError = 24,
```
clundin25: What do you think about this being a sum type wrapping the underlying error code? (Maybe a u32?) 

That would avoid papering over the underlying transport error

PR #1172, opened 2026-04-02
[builder]: Update Caliptra builder to use an argument struct

clundin25:
This greatly simplifies call sites since generally not all features are required at once.

Additionally, features can be added incrementally without touching every single call site.


zhalvorsen:
This is a great improvement!

PR #1217, opened 2026-04-10
ocp: activation flow

ternst-nv:
This implements the functions from both the Recovery Agent and integrator perspective for moving through the final, activation stage of a recovery image.

This concludes the chain of functionality required for an OCP integrator.  Integration style tests will be added within a follow up PR.

### DIFF
--- a/common/ocp/src/interface.rs
+++ b/common/ocp/src/interface.rs
@@ -56,6 +56,28 @@ pub struct RecoveryDeviceConfig<'a> {
     pub local_c_image_support: bool,
 }
 
+/// Result of an activation attempt, reported by the integrator via
+/// [`RecoveryStateMachine::complete_activation`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ActivationResult {
+    /// The activation succeeded and no more image stages are expected.
+    /// Sets recovery status to `Success` and device status to
+    /// `RunningRecoveryImage`.
+    Complete,
+
+    /// This activation stage succeeded but more image stages are needed.
+    /// Increments `recovery_image_index` and sets recovery status to
+    /// `AwaitingImage` so the host can push the next image.
+    StageSuccess,
+
+    /// The image failed its authentication requirement.  Sets recovery
+    /// status to `AuthenticationError`.
+    AuthenticationError,
+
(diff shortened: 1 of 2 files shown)


common/ocp/src/interface.rs
```diff
@@ -241,6 +263,82 @@ impl<'a, U: UsbDeviceDriver, V: VendorHandler> RecoveryStateMachine<'a, U, V> {
...
+
+    /// Transition the device into recovery mode with the given reason code.
+    ///
+    /// Sets `device_status_value` to `RecoveryMode`, stores the reason in
+    /// `recovery_reason` (reported via DEVICE_STATUS bytes 2-3), and sets
+    /// `recovery_status` to `AwaitingImage` with image index 0.
+    pub fn enter_recovery(&mut self, reason: RecoveryReasonCode) {
+        self.state.device_status_value = DeviceStatusValue::RecoveryMode;
+        self.state.recovery_reason = reason;
+        self.state
```
zhalvorsen, line 328: Should this also set the index to 0?

common/ocp/src/interface.rs
```diff
@@ -241,6 +263,82 @@ impl<'a, U: UsbDeviceDriver, V: VendorHandler> RecoveryStateMachine<'a, U, V> {
...
+                    vendor_status,
+                )?;
+                self.state.device_status_value = DeviceStatusValue::RecoveryMode;
+            }
+            ActivationResult::AuthenticationError => {
+                self.state
+                    .recovery_status
+                    .set_status(DeviceRecoveryStatus::AuthenticationError);
+                self.state.device_status_value = DeviceStatusValue::BootFailure;
+                self.state.recovery_reason = RecoveryReasonCode::AuthFailureRecoveryFirmware;
```
zhalvorsen: nit: you could use `set_boot_failure` for this and `ActivationResult::Failed`

zhalvorsen (approved):

common/ocp/src/interface.rs
ternst-nv: Yeah, good point.  When reviewing the spec for this question, I also found we were not setting the `RecoveryStatus` to booting as the spec states during the Activation transition.

zhalvorsen (approved):

PR #1218, opened 2026-04-10
[OCP-LOCK]: Initial code for HPKE Pub Key endorsement

clundin25:
Creates a certificate signed by the MCU FW DPE Context, currently only using EC.

There will be several follow up changes to keep fleshing this out.


runtime/userspace/api/caliptra-api/src/ocp_lock.rs
```diff
@@ -94,13 +145,159 @@ impl<'a> OcpLock<'a> {
...
+            version: 2,
+            serial_number,
+            signature: AlgorithmIdentifier {
+                oid: ECDSA_WITH_SHA_384,
+                parameters: None,
+            },
+            // TODO(clundin): Get issuer from DPE Certify Key
+            issuer: subject,
+            validity: Validity {
+                // TODO(clundin): Use a newer date ?
```
zhalvorsen, line 240: Note for a later PR, we could take this from a platform defined trait like Caliptra does with its certificates

runtime/userspace/api/caliptra-api/src/ocp_lock.rs
```diff
@@ -94,13 +145,159 @@ impl<'a> OcpLock<'a> {
...
+        HashContext::hash_all(HashAlgoType::SHA384, &tbs_der[..tbs_len], &mut digest).await?;
+
+        let label = {
+            let mut label = [0; 48];
+            label[..DPE_LABEL.len()].clone_from_slice(DPE_LABEL);
+            label
+        };
+
+        let dpe_cmd = SignP384Cmd {
+            handle: ContextHandle::default(),
```
zhalvorsen, line 270: This might need to come from the platform as well, depending on how they construct their DPE tree

zhalvorsen (approved):

runtime/userspace/api/caliptra-api/src/ocp_lock.rs
clundin25: I think the easiest way is to plumb a signature callback through as a function param

clundin25: WDYT? @zhalvorsen

zhalvorsen: Yeah, maybe through a trait or similar

Issue #1220, opened 2026-04-10
Emulator MCI mbox lock, SRAM clearing on reset are not equivalent to HW


swenson:
The emulator handles how the lock is set on reset differently (HW should start with the mbox locked by the root user).

We should make it consistent with the RTL (and, therefore, FPGA), and ensure that all mbox-related tests pass on both emulator and FPGA.

PR #1221, opened 2026-04-10
rom: add DOT commands to I3C services with packetized transport

### LINKED ISSUE #1078: DOT: I3C commands for recovery in ROM

swenson:
Implement DOT_STATUS, DOT_RECOVERY, DOT_UNLOCK_CHALLENGE, and DOT_OVERRIDE as ROM I3C service commands. Adds a packetization protocol to handle payloads exceeding the 256-byte I3C TTI FIFO, using MCI mailbox SRAM for reassembly.

There is an inconsistency with how the MCI mbox lock and SRAM clearing works on emulator and FPGA. I went down the rabbit hole on making them consistent, but it ended up spiraling into a bunch of unrelated changes, so I am saving that for a follow-up PR. (Filed as issue #1220.)

Fixes #1078


rom/src/cold_boot.rs
```diff
@@ -281,15 +282,48 @@ fn enter_i3c_services(
...
+    // (RTL: mcu_mbox.sv) forces lock=1 on reset with mbox_user set to
+    // strap_root_axi_user. The emulator should be updated to match this
+    // behavior, and all firmware paths that use the mailbox should perform
+    // the initial release (write execute=0) before first use. See issue #1220.
+    let lock_val = mci
+        .registers
+        .mcu_mbox0_csr_mbox_lock
+        .read(registers_generated::mci::bits::MboxLock::Lock);
+    if lock_val != 0 {
+        romtime::println!("[mcu-rom-i3c-svc] Warning: could not acquire mailbox lock");
```
mlvisaya, line 299: i believe acquiring the lock is required to be able to write to the mbox sram. in my test before, i see that when the lock is not acquired, writing to the sram won't change its contents. i think it's better to loop here for some number of iterations until we acquire a lock or fail, otherwise.

rom/src/cold_boot.rs
```diff
@@ -281,15 +282,48 @@ fn enter_i3c_services(
...
+    }
```
mlvisaya, line 300: nit: another thing i found is sometimes, MCU could lock mbox but  not release it. it could be from the previous run and MCU got reset.

it maybe better to release the lock before acquiring it here. 

I found this logic helped: https://github.com/chipsalliance/caliptra-sw/blob/10dd5e7542b7d24cf11115c2edd3d034294963d6/runtime/tests/runtime_integration_tests/test_authorize_and_stash.rs#L1060

rom/src/cold_boot.rs
```diff
@@ -599,7 +633,15 @@ impl BootFlow for ColdBoot {
                     // Try I3C services if DOT_RECOVERY enabled
                     if let Some(services) = params.i3c_services {
                         if services.contains(I3cServicesModes::DOT_RECOVERY) {
-                            enter_i3c_services(&env.mci, i3c_base, services);
+                            let dot_ctx = crate::DotContext {
```
ternst-nv, line 636: It looks like there is a `DotRecoveryHandler` trait.  It seems like it would be good to define the I3C dot recovery in terms of it, so that integrators could pick and choose the DOT recovery mechanism which is most suitable for them.

It seems like from the design doc, both I3C and MCI mailbox are intended recovery flows, as well as falling back to runtime.  I could imagine OCP Recovery over USB would also be a viable mechanism for providing a recovery DOT blob.

rom/src/i3c_mailbox.rs
```diff
@@ -102,14 +186,403 @@ impl I3cMailboxHandler {
...
+    }
+
+    /// Handle DOT_OVERRIDE: receive signed challenge response and complete override.
+    ///
+    /// Payload:
+    ///   bytes 0..95:    ECC P-384 public key (x: 48, y: 48)
+    ///   bytes 96..191:  ECC signature (r: 48, s: 48)
+    ///   bytes 192..2783: MLDSA-87 public key (2592 bytes)
+    ///   bytes 2784..7411: MLDSA-87 signature (4628 bytes)
+    fn handle_dot_override(&mut self, payload_len: usize) -> DispatchResult {
```
ternst-nv, line 376: This logic looks very similar to that in `rom/src/device_ownership_transfer.rs:dot_override_challenge_flow`.  Is there any way to deduplicate between them?

docs/src/rom_i3c_services.md
```diff
@@ -36,24 +36,67 @@ The ROM reports boot milestones via MCI when entering and leaving I3C services:
 
 ## Framing
 
+### Packetized Transport
```
ternst-nv, line 39: This spec seems to be doing something very similar to the OCP recovery flow.  Is there anyway we could take advantage of the i3c hardware OCP recovery flow to define a 4th image for DOT?  This may not be possible due to how the i3c hardware is defined, but curious. 

Also has this flow been tested with utilizing the i3c as well for the axi bypass mode for flash boot or normal i3c boot controlled by the BMC?

tests/integration/src/rom/test_i3c_services.rs
```diff
@@ -78,12 +149,314 @@ mod test {
...
+        let flush = hw.cycle_count() + 100_000;
+        hw.step_until(|m| m.cycle_count() >= flush);
+
+        assert_eq!(
+            hw.mci_fw_fatal_error(),
+            None,
+            "ROM hit fatal error during challenge"
+        );
+
+        // Try reading the response via private read (works on FPGA).
```
ternst-nv, line 312: Should we determine why the emulator i3c socket isn't delivering this? It feels concerning to not validate this using the flow integrators would expect (understand FPGA is working as expected).

tests/integration/src/rom/test_i3c_services.rs
```diff
@@ -78,12 +149,314 @@ mod test {
...
+    /// 3. Receive challenge from ROM via IBI
+    /// 4. Sign challenge with both ECDSA P-384 and MLDSA-87
+    /// 5. Send DOT_OVERRIDE with public keys and signatures
+    /// 6. Verify ROM logs success
+    #[test]
+    #[cfg_attr(
+        feature = "fpga_realtime",
+        ignore = "FPGA does not support otp_memory provisioning for DOT fuses"
+    )]
+    fn test_i3c_services_dot_override_full_flow() {
```
ternst-nv, line 233: If we need to support the protocol defined above, it would be good to add some error in the middle tests, verifying the recovery flow can still continue (whether retry the packet or the entire flow).

docs/src/rom_i3c_services.md
swenson: Theoretically, yes, we could use OCP Secure Firmware Recovery for this, and perhaps have the status and challenge come over a vendor-defined message (type 44) or INDIRECT_DATA (types 41 to 43), and then send the payloads as INDIRECT_FIFO_DATA packets.

However, today, I don't see a great way to communicate the nonce in particular -- the i3c-core doesn't support the vendor-defined message, and we don't have any other good way to communicate back to the BMC those 48 bytes.

We also don't have a DMA guaranteed in MCU, though with a tight enough firmware loop reading the data port and writing to DCCM, we could probably keep up with I3C (as long as we didn't do much of anything else).

tests/integration/src/rom/test_i3c_services.rs
swenson: Oh, this is my bad. I thought I had gotten this sorted before pushing. I'll work on this.

rom/src/i3c_mailbox.rs
swenson: Updated to add shared logic.

tests/integration/src/rom/test_i3c_services.rs
swenson: Added.

rom/src/cold_boot.rs
swenson: I took a stab at this.

rom/src/cold_boot.rs
swenson: Added some fixes there. Still need to fix some of the mbox lock logic in the emulator I think.

rom/src/cold_boot.rs
```diff
@@ -281,15 +284,62 @@ fn enter_i3c_services(
...
+        }
+        // First failed attempt: release a possibly stale lock from a prior run
+        if !released_stale {
+            released_stale = true;
+            mci.registers
+                .mcu_mbox0_csr_mbox_execute
+                .write(MboxExecute::Execute::CLEAR);
+        }
+    }
+    if !lock_acquired {
```
ternst-nv, line 310: Should we error out if the lock isn't acquired?

rom/src/cold_boot.rs
ternst-nv: Hmm, this looks like it still requires all implementors would need to support the i3c recovery as the last choice.  For the ImageProvider I defined a `Manager` which allowed the integrator to determine the provider which should be utilized based on the number of retries required: https://github.com/ternst-nv/caliptra-mcu-sw/blob/917467b7d5685d8c631a4bef7f2b2b3d524e6c78/rom/src/recovery.rs#L74, allowing any function which implemented the trait to get a DOT blob.

rom/src/i3c_mailbox.rs
```diff
@@ -102,14 +199,307 @@ impl I3cMailboxHandler {
...
+                self.send_status(STATUS_ERROR);
+                return DispatchResult::Continue;
+            }
+        };
+
+        // Reset override state
+        self.override_state = OverrideState::Idle;
+
+        // Expected payload: ECC PK (96) + ECC sig (96) + MLDSA PK (2592) + MLDSA sig (4628) = 7412
+        let expected_payload = 96
```
ternst-nv: Are there constants for the ECC pk and signature as well?

rom/src/i3c_mailbox.rs
```diff
@@ -195,36 +598,122 @@ impl I3cMailboxHandler {
...
-            .write(TtiResetControl::IbiQueueRst::SET);
+        // IBI was not acknowledged — hold the IBI queue in reset so
+        // the pending entry is cleared and no retries are attempted.
+        // These TTI reset bits are persistent ("hold-in-reset"), not
+        // self-clearing, so the IBI queue stays disabled for the
+        // remainder of the services loop.
+        caliptra_mcu_romtime::println!(
+            "[mcu-rom-i3c-svc] IBI not acknowledged, flushing IBI queue"
+        );
+        regs.tti_tti_reset_control.set(
```
ternst-nv: Are there register constants that could be used for this?

rom/src/i3c_mailbox.rs
```diff
@@ -11,60 +11,139 @@ const MDB_SERVICES: u8 = 0x1F;
...
 const MAX_RX_WORDS: usize = 64;
 
+/// Packet header size: [cmd, payload_len, seq_num, total_seqs].
+const PACKET_HEADER_SIZE: usize = 4;
+
+/// Maximum payload bytes per packet. The TTI RX FIFO is 256 bytes (64 DWORDs).
+/// The on-wire size includes a 4-byte header, payload, and 1-byte PEC
+/// (CRC-8/SMBus appended by the I3C private write). The chunk must also be
+/// 4-byte aligned for u32 word reassembly. 256 - 4 - 1 = 251, aligned → 248.
+#[allow(dead_code)]
```
ternst-nv: Looks like this `dead_code` can be removed.

rom/src/cold_boot.rs
swenson: We should, but there is a bug in the emulator on how the lock works. After that is fixed, we should make this an error. See #1220

rom/src/device_ownership_transfer.rs
```diff
@@ -1346,3 +1424,130 @@ pub fn process_fw_manifest_dot_commands(
     caliptra_mcu_romtime::println!("[mcu-rom-dot] Manifest DOT processing complete");
     Ok(())
 }
+
+/// Policy for handling errors from a [`DotLockedRecoveryHandler`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DotLockedRecoveryErrorPolicy {
```
ternst-nv, line 1430: Thanks Chris!  I think when we get to 2.1 we should likely write a Generic implementation of this logic and share it amongst multiple possible users (Dot, Image, etc.).

rom/src/cold_boot.rs
ternst-nv: Okay, could we attempt to fix that prior to landing, as the behavior we'd want on device is to error if the lock is not acquired.

PR #1224, opened 2026-04-12
rom/lifecycle: release transition mutex on error paths

### LINKED ISSUE #1223: [Bug] Lifecycle transition mutex not released on error paths
In ```lifecycle.rs```, the ```Lifecycle::transition()``` function acquires the claim_transition_if mutex before initiating a lifecycle state transition but fails to release it when the transition encounters an error.

The polling loop at Step 7 uses ```return Err(...)``` for all error conditions which exits the function immediately. The mutex release (```self.registers.claim_transition_if.set(0)```) only executes after the loop on the success path.

This means any failed lifecycle transition will leave the claim_transition_if mutex permanently held, preventing any subsequent transition attempts from acquiring it.

Also, the steps jump from 1 to 3, 4, 6, 7 - missing 2, 5, and the release step is unnumbered

giladliber:
Introduce TransitionMutexGuard to ensure the lifecycle transition mutex is always released when the function returns, including on error paths.
This replaces the manual release at the end of transition() and allows using early returns instead of break statements in the status polling loop.

Fixes #1223

### DIFF
--- a/rom/src/lifecycle.rs
+++ b/rom/src/lifecycle.rs
@@ -62,6 +62,36 @@ pub struct LifecycleHashedTokens {
 
 pub use caliptra_mcu_otp_lifecycle::LifecycleControllerState;
 
+/// RAII guard that releases the lifecycle transition mutex on drop.
+struct TransitionMutexGuard {
+    registers: StaticRef<lc_ctrl::regs::LcCtrl>,
+}
+
+impl TransitionMutexGuard {
+    /// Spin until the transition mutex is claimed, then return a struct
+    /// that will release it on drop.
+    fn claim(registers: StaticRef<lc_ctrl::regs::LcCtrl>) -> Self {
+        const MULTI_TRUE: u32 = 0x96;
+        while registers.claim_transition_if.get() != MULTI_TRUE {
+            registers.claim_transition_if.set(MULTI_TRUE);
+            let reg_value = registers.claim_transition_if.get();
+            caliptra_mcu_romtime::println!(
+                "[mcu-rom-lcc] Claim Mutex Register: {} (should be {})",
+                HexWord(reg_value),
+                HexWord(MULTI_TRUE)
+            );
+        }
+        caliptra_mcu_romtime::println!("[mcu-rom-lcc] Mutex successfully acquired.");
+        Self { registers }
+    }
+}
+
+impl Drop for TransitionMutexGuard {
+    fn drop(&mut self) {
+        self.registers.claim_transition_if.set(0);
+    }
+}
+
 pub struct Lifecycle {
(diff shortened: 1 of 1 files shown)


rom/src/lifecycle.rs
```diff
@@ -124,89 +124,92 @@ impl Lifecycle {
...
             if status.is_set(lc_ctrl::bits::Status::OtpPartitionError) {
-                caliptra_mcu_romtime::println!("[mcu-rom-lcc] OTP partition error detected.");
-                return Err(McuError::ROM_LC_OTP_PARTITION_ERROR);
+                romtime::println!("[mcu-rom-lcc] OTP partition error detected.");
+                break Err(McuError::ROM_LC_OTP_PARTITION_ERROR);
             }
-        }
+        };
 
+        // Step 6: Release the mutex, regardless of success or failure
```
clundin25: Does it make sense to wrap this in a type that releases the mutex on drop?

giladliber: Yes, that makes sense indeed. Then we can use "return" also for the Ok case exactly like in the Err cases.
I'll update the PR. Thanks!

clundin25: Thanks! If it turns into a larger refactor we can revisit, using a type seems like a good approach to prevent occurrences of this bug

giladliber: Updated. Let me know if that's ok and according to your coding guidelines.

clundin25 (approved):

clundin25:
@giladliber looks like there are some merge conflicts, do you mind taking a look?

rom/src/lifecycle.rs
```diff
@@ -111,102 +141,86 @@ impl Lifecycle {
...
+        romtime::println!(
             "[mcu-rom-lcc] Transitioning Lifecycle state... to {}",
             u8::from(state)
         );
         let next_lc_state_mne = Self::calc_lc_state_mnemonic(state);
         let token = u128::from_le_bytes(token.0);
 
-        const MULTI_TRUE: u32 = 0x96;
+        // Step 1: Acquire the transition mutex (released on drop)
+        let _mutex = TransitionMutexGuard::claim(self.registers);
```
clundin25: This mutex could be passed to a function that wraps the protected block? That makes it clear to the reader why and where it is held

zhalvorsen (approved):

rom/src/lifecycle.rs
giladliber: Done. Is that what you meant?

clundin25: Looks great! Thanks @giladliber !

RaunakGu:
@clundin25 - Could you please also port this to main-2.1?

clundin25:
@RaunakGu will do 👍

PR #1233, opened 2026-04-13
[test] fix failing test_new_unbooted in CI

timothytrippel:
Update test_new_unbooted to use pre-built binaries from the environment when available. This prevents the test from attempting to build MCU and Caliptra components from source in CI environments where the source code is not present.

Partially addresses #1226; Depends on #1233 (only review the last commit).

### DIFF
--- a/firmware-bundler/src/ld.rs
+++ b/firmware-bundler/src/ld.rs
@@ -602,49 +602,50 @@ INCLUDE $BASE_LD_CONTENTS
 
 /// Output a linker file for the application.
 fn content_aware_write(prefix: &str, content: &str, linker_dir: &Path) -> Result<PathBuf> {
-    // Determine if a previous file matching this prefix has already been generated.
-    // First read through the linker-script directory
-    let maybe_previous_file = std::fs::read_dir(linker_dir)?
-        .find(|f| {
-            f.as_ref()
-                .map(|f| {
-                    // Then check if each entry has a name which starts with the same name as
-                    // this linker file.  If so return it as the previous file.
-                    f.file_name()
-                        .to_str()
-                        .map(|n| n.starts_with(prefix))
-                        .unwrap_or(false)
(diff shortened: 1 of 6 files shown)


clundin25 (approved):

zhalvorsen (approved):

firmware-bundler/src/ld.rs
```diff
@@ -602,49 +602,50 @@ INCLUDE $BASE_LD_CONTENTS
...
+        // We found a file with matching content. We can ignore other files with the same prefix.
+        return Ok(file);
+    }
+
+    // If we didn't find a matching file, we need to write a new one.
+    // Use a unique UUID with each linker script generated.
     let output_file = linker_dir.join(format!("{}-{}.ld", prefix, uuid::Uuid::new_v4()));
     std::fs::write(&output_file, content)?;
+
+    // Only remove OLD files AFTER we've successfully written the new one,
```
ternst-nv, line 639: I'm curious if the tests in question are using the locks appropriately to avoid parallel builds?  This change seems to reduce the likelihood of clobbering a linker file, but not remove it.  As such I'm concerned it could result in non-deterministic test failures that will be hard to debug.

swenson: I think the parallel builds run in separate target directories, so hopefully couldn't clobber each other?

swenson (approved):

firmware-bundler/src/ld.rs
timothytrippel: I simplified this PR to only contain fixes to the test_new_unbooted test. These changes were moved to #1233.

timothytrippel:
Simplified this PR greatly to focus on the test_new_unbooted test to deal with a bunch of compounding CI failures. Now depends on #1233 which fixes the SVN tests.

PR #1236, opened 2026-04-14
tests: emulator OTP scrambled partition handling

swenson:
Two bugs prevented the DOT override tests from passing in CI:

1. Emulator OTP DAI read bug: reads from scrambled partitions at non-8-byte-aligned addresses unscrambled the wrong 64-bit block. Fix by aligning to 8-byte boundaries and returning the correct half-word based on the address offset.

2. Test bug: create_challenge_recovery_otp_memory wrote the vendor recovery PK hash as plaintext into partition 13 (VendorSecretProdPartition), which is scrambled. The DAI read path would unscramble it, producing garbage. Fix by pre-scrambling each 8-byte block before writing to OTP memory.

Re-enable test_dot_override_challenge_success,
test_dot_override_invalid_signatures_fails, and
test_dot_unlocked_state_custom_owner_keys in the nightly-emulator nextest profile.


zhalvorsen (approved):

clundin25 (approved):

swenson:
Thanks!

PR #1237, opened 2026-04-14
Fix PLDM firmware update: use correct MCTP endpoint ID

swenson:
The emulator's PLDM daemon socket used EndpointId(1) as the MCU
destination, but the MCU runtime MCTP mux starts with local_eid=0.
This was a latent bug since commit 6326d984 (PR #270), but was masked
because the test transport's set_pkt_payload_size(1024) caused all
PLDM messages to fit in a single MCTP packet. Single-packet messages
(SOM=1) are dispatched via process_first_packet which does not check
the destination EID.
    
Commit f0037300 (PR #1166) removed the set_pkt_payload_size override,
dropping the packet payload size from 1024 to the 64-byte default.
PLDM messages now span multiple MCTP packets, and the continuation
packets (SOM=0) go through process_packet which checks dest_eid against
local_eid. Since dest_eid=1 != local_eid=0, all continuation packets
were dropped with 'Packet not for this Endpoint', causing every
firmware update and streaming SoC boot test to time out.
    
Fix by using EndpointId(0) to match the MCU runtime's default EID,
consistent with the working test_fw_update_e2e test.
    
Re-enable 12 tests: 7 firmware_update_flash and 5 streaming_soc_boot.


zhalvorsen (approved):

clundin25 (approved):

swenson:
Thanks!

mlvisaya (approved):
LGTM

PR #1238, opened 2026-04-14
[rom] skip first functional PK hash slot on strap assertion

### LINKED ISSUE #1228: MCU needs to be able to rotate PK hash without disabling rollback
When we want to rotate a PK hash to a new one, we'd mark this in fuses in the VENDOR_PK_HASH_VALID field. However, we want to provide a reference to perform this rotation without disabling rollback to the old image if the new one is faulty, as this is a potential bricking risk.

A proposal for accomplishing this is [here](https://docs.google.com/document/d/1IUzV9pQu7mN81lZd-LnyOqCIErd7Hi54f51eAMXbS7U/edit?tab=t.0), using a strap pin to boot from 1 PK Hash slot forward without burning the fuse so as to not disable the possibility to roll back.

amit-kumarh:
To allow PK hash rotation without disabling rollback in fuses. includes emulator tests (and slight enhancements to the overall vendor key selection tests).

We are using ss_strap_generic[3]  - this is fully unused and not copied to Caliptra Core

### DIFF
--- a/docs/src/rom.md
+++ b/docs/src/rom.md
@@ -283,7 +283,7 @@ The verification process:
 The MCU ROM supports locking the selected vendor public key hash slot (and all higher order slots) to prevent post-ROM manipulation. This is controlled by writing to the `VENDOR_PK_HASH_VOLATILE_LOCK` register.
 
 **Strapping Override**:
-The locking behavior is gated by bit 0 of the hardware strapping register `SS_STRAP_GENERIC[2]`.
+The locking behavior is gated by bit 0 of the hardware strapping register `SS_STRAP_GENERIC[3]`.
 *   **Production Mode** (Bit 0 is `0`): The ROM automatically applies the volatile lock to the selected key slot index (locking that slot and all higher slots).
 *   **Provisioning Mode** (Bit 0 is `1`): The ROM skips applying the lock, allowing potential post-ROM code to provision higher order hash slots.
 
--- a/rom/src/fuses.rs
+++ b/rom/src/fuses.rs
@@ -148,23 +148,28 @@ pub trait VendorKeyPolicy {
 ///
 /// This policy selects the first key slot that is both marked valid
 /// in the VENDOR_PK_HASH_VALID mask and is functional (not fully revoked).
-pub struct DefaultVendorKeyPolicy;
+pub struct DefaultVendorKeyPolicy {
+    rotate_pk_hash: bool,
+}
 
 impl DefaultVendorKeyPolicy {
-    pub const fn new() -> Self {
-        Self
+    pub const fn new(rotate_pk_hash: bool) -> Self {
+        Self { rotate_pk_hash }
     }
 }
 
 impl Default for DefaultVendorKeyPolicy {
     fn default() -> Self {
-        Self::new()
+        Self::new(false)
     }
 }
 
 impl VendorKeyPolicy for DefaultVendorKeyPolicy {
     fn select_key(&self, otp: &Otp) -> McuResult<usize> {
         let valid_mask = otp.read_entry_multi::<1>(generated_fuses::VENDOR_PK_HASH_VALID)?[0];
+        let target_slot_count = if self.rotate_pk_hash { 2 } else { 1 };
+        let mut slots_found = 0;
+        let mut first_functional: Option<usize> = None;
 
         for i in 0..16 {
             if (valid_mask >> i) & 1 == 1 {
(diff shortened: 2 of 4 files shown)


docs/src/rom.md
```diff
@@ -283,7 +283,7 @@ The verification process:
 The MCU ROM supports locking the selected vendor public key hash slot (and all higher order slots) to prevent post-ROM manipulation. This is controlled by writing to the `VENDOR_PK_HASH_VOLATILE_LOCK` register.
 
 **Strapping Override**:
-The locking behavior is gated by bit 0 of the hardware strapping register `SS_STRAP_GENERIC[2]`.
+The locking behavior is gated by bit 0 of the hardware strapping register `SS_STRAP_GENERIC[3]`.
```
zhalvorsen, line 286: We should probably document the use of this new GPIO

rom/src/fuses.rs
```diff
@@ -148,23 +148,28 @@ pub trait VendorKeyPolicy {
...
     fn default() -> Self {
-        Self::new()
+        Self::new(false)
     }
 }
 
 impl VendorKeyPolicy for DefaultVendorKeyPolicy {
     fn select_key(&self, otp: &Otp) -> McuResult<usize> {
         let valid_mask = otp.read_entry_multi::<1>(generated_fuses::VENDOR_PK_HASH_VALID)?[0];
+        let target_slot_number = if self.rotate_pk_hash { 2 } else { 1 };
```
zhalvorsen: nit: can we use 0 based indexing here?

rom/src/rom.rs
```diff
@@ -62,7 +62,8 @@ pub struct Soc {
 
 impl Soc {
     pub const BOOT_FSM_DONE: u32 = 4;
-    pub const PK_HASH_STRAPPING_MASK: u32 = 0x1;
+    pub const PK_HASH_LOCK_STRAPPING_MASK: u32 = 0x1;
```
zhalvorsen: nit: this name to me implies the opposite. Can we use something like `PK_HASH_SKIP_LOCK_STRAPPING_MASK`?

docs/src/rom.md
amit-kumarh: Do you know a good place for it? I have this note in the ROM PR, and I additionally have a followup in the pipeline that adds a full breakdown of the selection and revocation/rotation flow to the integrator-guide.md file as well.

rom/src/fuses.rs
amit-kumarh: This is meant to be the count of functional slots - so in normal operation, take the first functional slot, and in rotation mode, take the second. Renamed the variable to hopefully clarify

rom/src/rom.rs
amit-kumarh: Done

docs/src/rom.md
zhalvorsen: That should be good enough I would guess. Thanks!

rom/src/fuses.rs
zhalvorsen: That works for me. I'm good with that naming.

zhalvorsen (approved):

tests/integration/src/rom/test_vendor_key_policy.rs
```diff
@@ -271,4 +241,51 @@ mod test {
...
+        hw.caliptra_soc_manager()
+            .soc_ifc()
+            .ss_strap_generic()
+            .get(3)
+            .expect("failed to get strapping register")
+            .write(|_| 0x2);
+
+        // Slot 0 is the first functional slot which we skip
+        // Slot 1 is marked invalid
+        // Slot 2 should be selected
```
timothytrippel, line 287: Should we also just test the most simple case of:
- slot 0 being valid
- slot 1 being valid (i.e. newly provisioned)

to ensure slot 1 is selected?

rom/src/rom.rs
```diff
@@ -203,7 +204,9 @@ impl Soc {
         self.registers.ss_strap_generic[1].set(OTP_DIRECT_ACCESS_CMD_REG_OFFSET);
 
         // Select the vendor public key slot to use.
-        let default_policy = DefaultVendorKeyPolicy::new();
+        let default_policy = DefaultVendorKeyPolicy::new(
+            self.registers.ss_strap_generic[3].get() & Self::PK_HASH_ROTATION_STRAPPING_MASK != 0,
```
timothytrippel, line 208: in a follow up PR, it might be nice to have all the generic strap indices be named constants; just noting here, maybe we can file an issue on this

timothytrippel (approved):

RaunakGu:
@amit-kumarh @timothytrippel,

I am trying to understand this change to port to our code.

Can you clarify how this programming is expected to be set by any firmware component? Since it is consumed by the MCU ROM during cold boot, is there a valid window in which this strap can be programmed and then executed? How do we plan to use this mechanism?

it also does not seem appropriate to allow any other vendor cores(outside CSS) to influence PK hash selection—what is  the significance of this change?

amit-kumarh:
Hey Raunak - sorry I missed this last week.

The rationale here is to provide some flexibility to address https://github.com/chipsalliance/caliptra-mcu-sw/issues/1228 , so provide an avenue to update to firmware that is signed with a new PK hash slot without burning away in fuses the ability to roll back until confirming that image can boot.

The strap pin could be actuated by the actor that's performing the update if it also wishes to rotate the key (or if an integration doesn't want to use this feature, the strap can be no-connected internally, and the hash selection will work as normal).

PR #1243, opened 2026-04-14
tests: fix DOT recovery invalid HMAC test for OTP scrambling

swenson:
create_locked_otp_memory() left the VendorSecretProdPartition region (partition index 13) as all-zeros in raw OTP memory. Since this partition is scrambled, the DAI read path unscrambles those zeros into non-zero values, causing recovery_pk_hash to appear non-empty. This made the ROM enter the DOT override challenge/response flow (which blocks waiting for I3C commands) instead of immediately fatal-erroring after the backup blob HMAC check failed.

Pre-scramble the zero blocks in the VendorSecretProdPartition so they unscramble back to zeros, matching the pattern used by create_challenge_recovery_otp_memory(). Re-enable the test in nextest.


mlvisaya (approved):

clundin25 (approved):

swenson:
Thanks!

PR #1244, opened 2026-04-14
Document pk hash selection and rotation in integrator guide

### LINKED ISSUE #1228: MCU needs to be able to rotate PK hash without disabling rollback

amit-kumarh:
Expand the integrator guide with a section on the Vendor PK Hash rotation and selection implemented in #1108 , #1238

### DIFF
--- a/docs/src/integrator-guide.md
+++ b/docs/src/integrator-guide.md
@@ -65,3 +65,111 @@ the integration-specific fuse map.
 - If you are **not** using DOT, then `CPTRA_SS_OWNER_PK_HASH` is the sole
   source of the owner PK hash and must be provisioned or another integrator-
   specific mechanism must be used.
+
+## Vendor Public Key Selection and Rotation
+Caliptra MCU supports a vendor public key selection and rotation scheme
+based on fuses and hardware strapping pins. This section describes how the ROM
+selects the active vendor public key slot and how integrators can manage
+rotation and revocation.
+
+### Key Policy and Selection Process
+The ROM follows this process to select a vendor public key hash slot (out of 16
+available slots):
+1.  **Validity Check**: The ROM reads the `VENDOR_PK_HASH_VALID` fuse mask.
+    Each bit corresponds to a slot. If a bit is set to `1`, the slot is
+    considered invalid and skipped.
+2.  **Revocation Check**: For each valid slot, the ROM checks the revocation
+    status of the keys in the `CPTRA_CORE_ECC_REVOCATION_X`,  
+    `CPTRA_CORE_MLDSA_REVOCATION_X`, and `CPTRA_CORE_LMS_REVOCATION_X` fields:
+    -   **ECC Keys**: Checked against the ECC revocation fuses (4 bits per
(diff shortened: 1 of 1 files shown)


docs/src/integrator-guide.md
```diff
@@ -65,3 +65,111 @@ the integration-specific fuse map.
...
+3.  **Default Selection**: By default, the ROM selects the **first functional
+    slot** it encounters (searching from slot 0 to 15). However, this logic can
+    be overridden by passing a different implementation of the `VendorKeyPolicy`
+    into the ROM parameters.
+
+### Key Rotation via Strapping
+Integrators can force the ROM to rotate to the next available key by using a
+hardware strapping pin:
+-   **Strap Register**: `SS_STRAP_GENERIC[3]`
+-   **Bit 1 (Rotation)**: If this bit is set to `1`, the ROM will **skip the
```
timothytrippel, line 99: Bit 1 of the `SS_STRAP_GENERIC[3]` register? or bit 0 of said register?

docs/src/integrator-guide.md
```diff
@@ -65,3 +65,111 @@ the integration-specific fuse map.
...
+    first functional slot** it finds and select the **second functional slot**.
+This allows a platform to switch to a new key without burning fuses, simply
+by changing a strapping resistor or GPIO state, provided that a second valid and
```
timothytrippel:
```suggestion
by changing a strapping register or GPIO state, provided that a second valid and
```

docs/src/integrator-guide.md
```diff
@@ -65,3 +65,111 @@ the integration-specific fuse map.
...
+```
+
+#### Case 2: Revocation Across Different PK Hash Slots
+
+This flow is used when the entire PK hash slot is compromised or needs to be
+replaced, requiring a transition to a new PK hash slot.
+
+**Process**:
+1.  Push a new Runtime (RT) firmware signed with a key from a new PK hash slot.
+2.  Additionally assert the hardware strapping pin (Bit 1 of
```
timothytrippel, line 151: same here, is this supposed to be 
```suggestion
2.  Additionally assert the hardware strapping pin (Bit 0 of
```

or is bit 0 used for something else?

timothytrippel (approved):

docs/src/integrator-guide.md
amit-kumarh: Should be bit 1 - bit 0 controls the PK Hash locking

Issue #1245, opened 2026-04-15
[xtask,ci] use xtask commands for running FPGA tests in CI workflows


timothytrippel:
To enable local test running and debugging that matches the environment used in CI, we should refactor the xtask FPGA test commands to be usable by the CI FPGA test flow. This will require:

Addressed in #1246:
- [x] enhance the xtask FPGA bootstrap command to allow loading a pre-downloaded bitstream
- [x] enable the xtask FPGA flows to read the environment variables that are normally hard coded but configurable in CI YAML configs

Addressed in #1247:
- [x] Introduce new flags, e.g., `--rom-features`, `--separate-runtimes`, `--mcu_cfg`, to `xtask fpga build` command and `--no-container` to `xtask fpga build-test` command as these are used by CI flows

Addressed in #1263:
- [x] add the ability to specify an FPGA's configuration (ss, core, or core on ss) since the `xtask fpga build[-test]` commands are run in CI environments where FPGAs are not always assumed to be pre-bootstrapped

Addressed in #1264:
- [x] refactor the FPGA test CI workflow to use `cargo xtask fpga ...` commands instead of invoking `cargo nextest ...` directly

PR #1247, opened 2026-04-15
[xtask] add FPGA build options for CI parity

timothytrippel:
Introduce new flags to streamline FPGA build operations:

- Add --rom-features and --separate-runtimes to 'build' for parity
  with previous all-build flows.
- Add --mcu_cfg to 'build' to allow custom SoC manifest configurations.
- Add --no-container to 'build-test' to allow native cross-compilation
  in CI environments.

Partially addresses #1245 ; depends on #1246

### DIFF
--- a/xtask/src/fpga/mod.rs
+++ b/xtask/src/fpga/mod.rs
@@ -4,7 +4,7 @@ use anyhow::{anyhow, bail, Result};
 use caliptra_image_gen::to_hw_format;
 use caliptra_image_types::FwVerificationPqcKeyType;
 use caliptra_mcu_builder::flash_image::build_flash_image_bytes;
-use caliptra_mcu_builder::FirmwareBinaries;
+use caliptra_mcu_builder::{FirmwareBinaries, ImageCfg};
 use caliptra_mcu_hw_model::{InitParams, McuHwModel, ModelFpgaRealtime};
 use caliptra_mcu_rom_common::LifecycleControllerState;
 use clap::Subcommand;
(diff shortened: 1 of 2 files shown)


xtask/src/fpga/mod.rs
```diff
@@ -23,18 +23,26 @@ mod utils;
 struct BuildArgs<'a> {
     mcu: bool,
     fw_id: &'a Option<String>,
+    rom_features: &'a Option<String>,
+    separate_runtimes: bool,
```
clundin25, line 27: Do we need this? Isn't this always true?

xtask/src/fpga/mod.rs
```diff
@@ -23,18 +23,26 @@ mod utils;
...
+    mcu_cfgs: &'a Option<Vec<ImageCfg>>,
```
clundin25, line 28: out of scope but this type is complex enough to be a JSON or TOML file IMO :)

xtask/src/fpga/mod.rs
```diff
@@ -121,6 +151,9 @@ pub(crate) enum Fpga {
         /// Uses a `cargo-nextest` package filter-set, e.g. `package(caliptra-rom)`.
         #[arg(long)]
         package_filter: Option<String>,
+        /// Build tests without using a container.
+        #[arg(long)]
+        no_container: bool,
```
clundin25: I thought the plan was to use a container for the build?

clundin25 (approved):

PR #1249, opened 2026-04-15
ocp: integrate ocp recovery boot

ternst-nv:
This PR takes the work done on the OCP state machine and integrates it into the MCU Rom.  This includes providing a mechanism for an integrator to specify their boot order, as well as the mechanisms to fallback on in case of failure and what a failure definition looks like.

A basic test was added validating some common OCP commands work.  Follow on PRs will validate both full load image paths as well as various negative tests.


rom/src/recovery.rs
```diff
@@ -36,12 +37,90 @@ pub trait ImageProvider {
...
     fn bytes_loaded(&self) -> usize;
 }
 
+/// Policy for handling errors from an [`ImageProvider`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ErrorPolicy {
+    /// On error, skip this provider and try the next one.
+    Continue,
+    /// Retry up to N times, then fall through to the next provider.
+    Retry(u32),
```
pjhades, line 53: nit: maybe using a smaller integer type?

rom/src/recovery.rs
```diff
@@ -36,12 +37,90 @@ pub trait ImageProvider {
...
+                ErrorPolicy::Retry(n) => {
+                    if self.retries < n {
+                        self.retries += 1;
+                    } else {
+                        self.current += 1;
+                        self.retries = 0;
+                    }
+                }
+                ErrorPolicy::RetryForever => {
+                    self.retries += 1;
```
pjhades, line 109: nit: this seems to be a good place for `saturating_add`.

rom/src/recovery.rs
```diff
@@ -36,12 +37,90 @@ pub trait ImageProvider {
...
+    ///
+    /// - First call for a provider: returns it (retries = 0).
+    /// - `Continue`: immediately advances to the next provider on next call.
+    /// - `Retry(n)`: stays on current while retries < n, then advances.
+    /// - `RetryForever`: stays on current indefinitely.
+    ///
+    /// Returns `None` if all providers are exhausted.
+    fn provider(&mut self) -> Option<&mut (dyn ImageProvider + 'a)> {
+        if self.retries == 0 {
+            self.retries += 1;
```
pjhades, line 93: Do we need to consider the corner case where the policy is `Retry(0)`?

rom/src/cold_boot.rs
```diff
@@ -762,19 +759,19 @@ impl BootFlow for ColdBoot {
...
+        let recovery_boot = ((mci.registers.mci_reg_generic_input_wires[1].get() & (1 << 29)) != 0)
+            || params.request_recovery_boot;
 
-        if flash_boot && (params.flash_partition_driver.is_none() || !cfg!(feature = "hw-2-1")) {
+        if recovery_boot && (params.image_provider_manager.is_none() || !cfg!(feature = "hw-2-1")) {
             romtime::println!(
-                "Flash boot requested but missing flash driver or AXI bypass not enabled in ROM"
+                "Recovery boot requested but missing image provider or AXI bypass not enabled"
             );
             fatal_error(McuError::ROM_COLD_BOOT_FLASH_NOT_CONFIGURED_ERROR);
```
pjhades, line 769: Should this error enum be updated as well?

platforms/emulator/rom/src/riscv.rs
```diff
@@ -309,19 +398,27 @@ pub extern "C" fn rom_entry() -> ! {
...
-            // request_flash_boot defaults to false - emulator sets the wire when flash boot is requested
+            // request_recovery_boot defaults to false - emulator sets the wire when flash boot is requested
             cptra_mbox_axi_users: mbox_axi_users,
             cptra_fuse_axi_user: axi_user0,
             cptra_trng_axi_user: axi_user0,
             cptra_dma_axi_user: axi_user0,
             mci_mbox0_axi_users: mbox_axi_users,
             mci_mbox1_axi_users: mbox_axi_users,
-            #[cfg(feature = "ocp-lock")]
-            ocp_lock_config,
```
clundin25, line 324: Looks like this was accidentally removed

pjhades (commented):
Left some comments on changes outside common/ocp. Will take another look later.

common/ocp/src/interface.rs
```diff
@@ -339,10 +368,69 @@ impl<'a, U: UsbDeviceDriver, V: VendorHandler> RecoveryStateMachine<'a, U, V> {
...
+
+    /// Returns the CMS index from the current RECOVERY_CTRL register.
+    pub fn recovery_ctrl_cms(&self) -> u8 {
+        self.state.recovery_ctrl.cms
+    }
+
+    /// Returns the image size from the current INDIRECT_FIFO_CTRL register.
+    pub fn fifo_image_size(&self) -> u32 {
+        // Translate from 4 byte words to bytes.
+        self.state.indirect_fifo_ctrl_image_size * 4
```
pjhades: nit: define 4 as a constant of this type?

rom/src/recovery.rs
ternst-nv: Good catch.  I'll change the Retry enum to take `NonZeroU8` so we can't encounter it.

rom/src/cold_boot.rs
ternst-nv: Yupp, will do.

pjhades (approved):
I did a 2nd round, focusing on changes in common/ocp and they look good to me.

pjhades (approved):
LGTM!

PR #1250, opened 2026-04-15
runtime: enable MCI interrupts before kernel-level tests

swenson:
The MCI interrupt enablement (global_intr_en, notif0_intr_en) was positioned after the kernel-level test dispatch block in board.rs. This meant test_mcu_mbox_driver and test_mcu_mbox_soc_requester_loopback ran without mailbox command-available interrupts enabled, causing them to hang waiting for the MCU mailbox driver callback to fire.

Move the interrupt enablement before the test dispatch block so all kernel-level tests have MCI interrupts available. Remove the duplicate enablement that was positioned just before the main kernel loop.

This allows the mbox tests to pass.

### DIFF
--- a/platforms/emulator/runtime/src/board.rs
+++ b/platforms/emulator/runtime/src/board.rs
@@ -793,13 +793,66 @@ pub unsafe fn main() {
         BOARD = Some(board_kernel);
     }
 
-    // Disable WDT1 before running the loop or tests
+    // Disable WDT and enable MCI interrupts before tests and the kernel loop.
     let mci: StaticRef<mci::regs::Mci> =
         unsafe { StaticRef::new(MCU_MEMORY_MAP.mci_offset as *const mci::regs::Mci) };
     let mci_wdt = caliptra_mcu_romtime::Mci::new(mci);
     mci_wdt.disable_wdt();
 
-    // Run any requested test
+    mci.intr_block_rf_global_intr_en_r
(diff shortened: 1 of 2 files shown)


platforms/emulator/runtime/src/board.rs
```diff
@@ -799,6 +799,14 @@ pub unsafe fn main() {
...
+    // Enable MCI Interrupts
+    mci.intr_block_rf_global_intr_en_r
+        .modify(mci::bits::GlobalIntrEnT::NotifEn::SET + mci::bits::GlobalIntrEnT::ErrorEn::SET);
+    mci.intr_block_rf_notif0_intr_en_r.modify(
+        mci::bits::Notif0IntrEnT::NotifCptraMcuResetReqEn::SET
+            + mci::bits::Notif0IntrEnT::NotifMbox0CmdAvailEn::SET,
+    );
+
     // Run any requested test
     let exit = if cfg!(feature = "test-exit-immediately") {
```
clundin25: Does it make sense to wrap these in a function that documents the pre-requisites for this block? E.g. the fix being done in this PR

clundin25 (approved):

platforms/emulator/runtime/src/board.rs
swenson: Good call. Will do

swenson:
Thanks!

PR #1251, opened 2026-04-15
Optimize PLDM transfer performance and add benchmark

swenson:
- Make [profile.test] inherit from release for optimized test binaries
- Reduce I3C HCI polling interval from 1000 to 100 ticks for faster I/O
- Replace 10ms recv_timeout with non-blocking try_recv + 100μs condvar wait in i3c_socket_server, reducing I3C socket communication latency
- Add test_firmware_update_benchmark_64k test for measuring PLDM transfer throughput (transfers 64KB via PLDM and reports KB/s)

Combined effect: ~42% PLDM transfer speedup (1114 -> 1585 B/s).


swenson:
Thanks!

PR #1252, opened 2026-04-15
[firmware-bundler,tests] fix race condition in linker script generation and renable svn tests

timothytrippel:
Modified content_aware_write to handle multiple linker scripts with the same prefix. The function now scans all files in the linker directory, returning an existing file if its content matches. If no match is found, a new file with a unique UUID is written, and only then are the non-matching files removed. This prevents parallel build processes from deleting linker scripts currently in use by other processes.

Fixed the mcu svn integration tests by:

- adding vendor_test_partition to InitParams and ModelEmulated to allow provisioning of vendor test fuses in the emulator, and
- using absolute paths for firmware/emulator bundles in xtask test command.

Partially addresses #1226.


clundin25 (approved):
@timothytrippel how was this manifesting? Flaky tests?

timothytrippel:
yes; I was seeing random failures locally when debugging

timothytrippel:
closing as changes already merged in #1233

PR #1254, opened 2026-04-15
dot: Add locks prior to build

ternst-nv:
A build occurs in `compute_hmac` which is commonly called in `create_valid_dot_blob` which in a number of tests was called prior to the test lock being acquired.  This resulted in a number of parallel builds occuring, resulting in unexpected behavior.


swenson (approved):
LGTM, great catch!

PR #1256, opened 2026-04-15
[mailbox] Consolidate populating checksum logic

zhalvorsen:
This removes populating the checksum from each command handler and populates it for all commands after execution. This simplifies flow and follows the pattern the Caliptra Core command handler uses.


runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -80,8 +85,20 @@ impl<'a> CmdInterface<'a> {
...
                         if status == MbxCmdStatus::Complete {
+                            // guarantee it is big enough to hold the header
+                            let resp_len = resp_len.max(size_of::<MailboxRespHeader>());
+
+                            if resp_len > msg_buf.len() {
+                                // should be impossible
+                                return Err(MsgHandlerError::InvalidParams);
+                            }
+
+                            let msg_buf = &mut msg_buf[..resp_len];
```
clundin25: nit: could use zerocopy here to condense the code?

clundin25 (approved):

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
zhalvorsen: Done

zhalvorsen: I tried it, but it turns out the logic would have been more complex with zerocopy in this instance

PR #1261, opened 2026-04-15
[rom] Set FW protected region boundary

### LINKED ISSUE #82: ROM: when starting, lock the AXI SRAM to be protected data region using MCI registers

amit-kumarh:
Per https://chipsalliance.github.io/caliptra-web/docs/2.1/subsystem/ss_hardware_spec.html#mcu-sram-1

Set the FW_SRAM_EXEC_REGION_SIZE MCI register to set the protected region of the SRAM from a value in the RomParameters.

Note that if this value is too small, the MCU RT will fail, as the instruction fetch unit does NOT have access to the protected region; only the executable region. By default, make the SRAM executable up to the last 32k

### DIFF
--- a/rom/src/cold_boot.rs
+++ b/rom/src/cold_boot.rs
@@ -512,6 +512,13 @@ impl BootFlow for ColdBoot {
         );
         mci.set_flow_checkpoint(McuRomBootStatus::McuMboxAxiUsersConfigured.into());
 
+        let size_value = params.mcu_fw_sram_exec_region_size.unwrap_or(
+            unsafe { MCU_MEMORY_MAP.sram_size }
+                - crate::MCU_SRAM_DEFAULT_PROTECTED_REGION_BLOCKS * 4096
+                - 1,
+        );
+        mci.set_fw_sram_exec_region_size(size_value);
+
         // Set SS_CONFIG_DONE_STICKY to lock MCI configuration registers
         caliptra_mcu_romtime::println!(
             "[mcu-rom] Setting SS_CONFIG_DONE_STICKY to lock configuration"
--- a/rom/src/rom.rs
+++ b/rom/src/rom.rs
@@ -45,6 +45,8 @@ const LMS_CALIPTRA_VALUE: u8 = 3;
 const OTP_DAI_IDLE_BIT_OFFSET: u32 = 22;
 const OTP_DIRECT_ACCESS_CMD_REG_OFFSET: u32 = 0x60;
 
+pub const MCU_SRAM_DEFAULT_PROTECTED_REGION_BLOCKS: u32 = 8; // 32 kB / 4 kB chunks
+
 /// Trait for different boot flows (cold boot, warm reset, firmware update)
 pub trait BootFlow {
     /// Execute the boot flow
@@ -739,6 +741,8 @@ pub struct RomParameters<'a> {
     /// Note that in 2.0, Caliptra already sets recovery status as successful so there may be a race
     /// condition depending on when a BMC reads the recovery status.
     pub recovery_status_open: bool,
+    /// Size of the executable SRAM region to pass into FW_SRAM_EXEC_REGION_SIZE
+    pub mcu_fw_sram_exec_region_size: Option<u32>,
     /// Valid AXI users for Caliptra mailbox. 0 values are ignored.
     pub cptra_mbox_axi_users: [u32; 5],
     /// Valid AXI user for Caliptra fuse registers. 0 = don't configure.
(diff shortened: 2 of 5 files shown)


rom/src/cold_boot.rs
```diff
@@ -509,6 +509,12 @@ impl BootFlow for ColdBoot {
         );
         mci.set_flow_checkpoint(McuRomBootStatus::McuMboxAxiUsersConfigured.into());
 
+        // Configure SRAM Execution Region Size before locking configuration
+        let size_value = params
```
swenson: Perfect.

We should probably also add this to the ROM cold boot documentation. (Not blocking this PR though.)

swenson (approved):
LGTM

rom/src/cold_boot.rs
amit-kumarh: done!

amit-kumarh:
Confirmed on my dev FPGA this passes testing when we give the different actors different AXI user IDs - MCU LSU, MCU IFU, and SRAM Config (Caliptra).

In a follow-up, I can (and think we should) change the FPGA test setup so all the user IDs are different

rom/src/rom.rs
```diff
@@ -45,6 +45,8 @@ const LMS_CALIPTRA_VALUE: u8 = 3;
 const OTP_DAI_IDLE_BIT_OFFSET: u32 = 22;
 const OTP_DIRECT_ACCESS_CMD_REG_OFFSET: u32 = 0x60;
 
+pub const MCU_SRAM_DEFAULT_PROTECTED_REGION_BLOCKS: u32 = 8; // 32 kB / 4 kB chunks
```
swenson, line 48: I think this should be a RomParameter that can be overriden, right? Otherwise it will be more difficult for integrators to override.

amit-kumarh: It's still a parameter:

```
let size_value = params.mcu_fw_sram_exec_region_size.unwrap_or(
            unsafe { MCU_MEMORY_MAP.sram_size }
                - crate::MCU_SRAM_DEFAULT_PROTECTED_REGION_BLOCKS * 4096
                - 1,
        );
```

I just pulled out the default value to a constant so that we didn't have the magic number in both the cold and warm boot flows

swenson: Nevermind, I can't read :)

swenson (approved):

timothytrippel (approved):

rom/src/warm_boot.rs
```diff
@@ -94,6 +94,13 @@ impl BootFlow for WarmBoot {
         );
         mci.set_flow_checkpoint(McuRomBootStatus::McuMboxAxiUsersConfigured.into());
 
+        let size_value = params.mcu_fw_sram_exec_region_size.unwrap_or(
+            unsafe { MCU_MEMORY_MAP.sram_size }
```
timothytrippel, line 98: one thing I noticed (we can harden this in a follow up PR), but should we be checking that "crate::MCU_SRAM_DEFAULT_PROTECTED_REGION_BLOCKS * 4096" is smaller than the sram_size somewhere to prevent overflow/underflow? (ideally statically though challenging with current extern C nature of MCU_MEMORY_MAP struct.

timothytrippel (approved):

rom/src/warm_boot.rs
amit-kumarh:
```
// Check if there was overflow due to fw_sram_exec_region_size being larger than the entire
// SRAM.
assign exec_region_overflow  = |exec_region_end_calc[28:MCU_SRAM_CIF_ADDR_W];
// If there was overflow set to the MCU SRAM size.
// Otherwise take the calculated value
assign exec_region_end  = exec_region_overflow ? (MCU_SRAM_SIZE_BYTES-1) : 
                            exec_region_end_calc;  
```

This case should be handled in HW for the actual region size, but probably worth doing to prevent underflow to the Rust integer - although the RomParam is immutable, right? so if the value gets set to something that over/underflows we're hosed anyway

timothytrippel: ah that is good to know. Yeah probably worth doing, but in a follow up. Let's just file an issue for now.

tonywangxy:
Hi @amit-kumarh @swenson 

The unit of FW_SRAM_EXEC_REGION_SIZE is in multiples of 4KiB. Looks like this PR passes raw byte count to the register. Can you help to check? Thanks!

amit-kumarh:
Hey @tonywangxy - thanks for the catch. Yep, that register is specified in 4k chunks - just an oversight on my part. Sent out #1428 and #1429 to fix the default boundary calculation, and we'll probably patch release for 2.0.

PR #1262, opened 2026-04-16
[romtime]: Add handoff table

clundin25:
Reserves 4K of DCCM space for a handoff table. This is used to pass state from ROM to Runtime.

OCP LOCK HEK seed state is passed from ROM to Firmware using this mechanism


mgbharath:
Hi @clundin25 

When do you plan to merge this code to main2.1?

clundin25:
Hey @mgbharath I need to verify that this is passing CI and then get the team to review. My hope is to have this wrapped up by the end of the week

mgbharath:
Thanks for the quick response :)

common/config/src/lib.rs
```diff
@@ -67,6 +70,9 @@ impl Default for McuMemoryMap {
             dccm_size: 256 * 1024,
             dccm_properties: MemoryRegionType::MEMORY,
 
+            handoff_offset: 0x5000_3000,
```
ternst-nv: Hmm, this structure concerns me as we then have to keep the config and linker scripts in order to guarantee the correct resolution of the Handoff table.  When I've done things like this previously I've used the `link_section` attribute: https://doc.rust-lang.org/reference/abi.html#the-link_section-attribute.  That can be used to choose a specific location in memory via the linker, and then assign variables to it.

That would allow us to define a static buffer in both the Rom and Runtime, ensure it has the same location, and then not have to define the value within the config library.

platforms/emulator/runtime/src/handoff.rs
```diff
@@ -0,0 +1,34 @@
...
+
+pub struct HandOff {
+    data: romtime::handoff::HandoffData,
+}
+
+impl HandOff {
+    /// Read the handoff data from DCCM.
+    pub fn from_dccm() -> Option<Self> {
+        let handoff_addr = HandoffData::dccm_address() as *const HandoffData;
+        let data = unsafe { core::ptr::read_volatile(handoff_addr) };
```
ternst-nv: See comment about `link_section`.  This would allow us to use the standard zero copy APIs over the buffer rather than having to do read/write volatile.

romtime/src/handoff.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub fht_marker: u32,
+
+    /// Major version of FHT.
+    pub fht_major_ver: u16,
+
+    /// Minor version of FHT.
+    pub fht_minor_ver: u16,
+
+    /// HEK state from OCP LOCK fuse population.
+    #[cfg(feature = "ocp-lock")]
```
ternst-nv, line 31: Is there a gain to only including HEK if OCP lock is enabled?  If we do then we are effectively saying we need to support two or more tables depending on what set of features are enabled in the firmware.  Would it be possible to make this an Option, and then only assign a value if we are in OCP mode?

ternst-nv: It isn't going to save us any space unless we dynamically size the buffer based on what features we have enabled, which seems like more complexity than what we'd want for this.

romtime/src/handoff.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// Handoff data produced by ROM.
+#[repr(C, packed)]
+#[derive(Debug, TryFromBytes, IntoBytes, KnownLayout, Immutable, Clone, Copy)]
+pub struct RomHandoffTable {
+    /// Magic Number marking start of table.
+    pub fht_marker: u32,
+
+    /// Major version of FHT.
+    pub fht_major_ver: u16,
```
ternst-nv, line 25: 👍

common/config/src/lib.rs
clundin25: Awesome, I will dig into this, I wasn't aware of this but this seems much safer

romtime/src/handoff.rs
clundin25: I am a bit unsure what the best approach is given integrators can modify this as they wish.

Right now OCP LOCK is gated behind a feature flag, so using an `Option` won't work because the type won't be present if it's not enabled.

I've feature gated OCP LOCK since it's code that is really only relevant to storage vendors.

re: Dynamically sizing the buffer, I think we should avoid this since the data is defined in ROM

platforms/emulator/runtime/src/handoff.rs
clundin25: Done !

builder/src/rom.rs
```diff
@@ -146,12 +146,15 @@ OUTPUT_ARCH( "riscv" )
...
 {
-  ROM   (rx) : ORIGIN = $ROM_OFFSET, LENGTH = $ROM_SIZE
-  RAM  (rwx) : ORIGIN = $DCCM_OFFSET, LENGTH = $DCCM_SIZE /* dedicated SRAM for the ROM stack */
+  ROM     (rx) : ORIGIN = $ROM_OFFSET, LENGTH = $ROM_SIZE
+  RAM     (rwx) : ORIGIN = $DCCM_OFFSET, LENGTH = $DCCM_SIZE /* dedicated SRAM for the ROM stack */
 }
 
 SECTIONS
 {
+  .handoff ORIGIN(RAM) (NOLOAD) : {
```
ternst-nv: I think`ORIGIN` will place this at the bottom of the DCCM, instead of the top like we discussed, might be good to check the assigned address via readelf/objdump and make sure.

builder/src/rom.rs
```diff
@@ -146,12 +146,15 @@ OUTPUT_ARCH( "riscv" )
...
+    KEEP(*(.handoff))
+  } > RAM
```
ternst-nv: [nit] Might be nice to put this with the other RAM sections for future readability.  I like placing the `.handoff` in its own section.

common/config/src/lib.rs
```diff
@@ -284,6 +290,15 @@ impl McuMemoryMap {
         );
         map.insert("DCCM_SIZE".to_string(), format!("0x{:x}", self.dccm_size));
 
+        map.insert(
```
ternst-nv, line 293: I think with the linker statements we can get rid of this in both the firmware-bundler and config files?

romtime/src/handoff.rs
ternst-nv: Hmm, that is a good point.  Dynamically sizing was the wrong term to use, I meant more the fields would vary based on underlying firmware, for example the `padding` field won't actually pad to 64 bytes for any firmware that isn't compiled with OCP Lock enabled.

My concern is that you'd get a MCU Runtime compiled with one set of options, which doesn't correlate with a ROM compiled from another set of options.  While that shouldn't happen, it would be good to make sure that in the inadvertent case it does the system remains in a recoverable state.

I don't think we need to solve it in this PR, but I think it is something that bears further thought.  Maybe something like a list of entries detailing which "optional" structs have been added and their offset?

platforms/emulator/runtime/src/handoff.rs
```diff
@@ -0,0 +1,45 @@
...
+use romtime::ocp_lock::HekState;
+
+pub struct HandOff {
+    data: HandoffData,
+}
+
+impl HandOff {
+    /// Read the handoff data from DCCM.
+    pub fn from_dccm() -> Option<Self> {
+        let data = unsafe {
```
swenson: nit: add `// Safety` comment

romtime/src/handoff.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Top-level handoff structure stored in DCCM.
+/// Resident at a well-known location in DCCM.
+///
+/// SAFETY: This structure MUST NOT exceed the reserved memory region size (4KB)
+/// defined in the linker scripts and memory map. Exceeding this size will cause
+/// memory corruption or linker errors.
+///
+/// ALIGNMENT: This structure is explicitly 4-byte aligned.
+#[repr(C, align(4))]
+#[derive(Debug, TryFromBytes, IntoBytes, KnownLayout, Immutable, Clone, Copy, Default)]
```
swenson: I think we don't want this to be `Copy`, because we don't want to accidentally copying it on to the stack

platforms/emulator/runtime/src/handoff.rs
```diff
@@ -0,0 +1,45 @@
...
+            romtime::println!(
+                "[mcu-runtime] Reading handoff table from DCCM at {:p}",
+                &raw const romtime::handoff::HANDOFF
+            );
+            romtime::handoff::HANDOFF
+        };
+
+        if data.rom.fht_marker == romtime::handoff::FHT_MARKER {
+            Some(Self { data })
```
swenson: I'm not entirely convinced that this isn't being copied onto the stack.

Wouldn't the more correct pattern be to have the handoff table live in a `StaticRef`, so that we know it doesn't accidentally get copied onto the stack?

romtime/src/handoff.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Top-level handoff structure stored in DCCM.
+/// Resident at a well-known location in DCCM.
+///
+/// SAFETY: This structure MUST NOT exceed the reserved memory region size (4KB)
+/// defined in the linker scripts and memory map. Exceeding this size will cause
+/// memory corruption or linker errors.
+///
+/// ALIGNMENT: This structure is explicitly 4-byte aligned.
+#[repr(C, align(4))]
```
swenson, line 78: align(4) is 16-byte aligned

romtime/src/handoff.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub rom: RomHandoffTable,
+
+    /// Runtime handoff table.
+    pub runtime: RuntimeHandoffTable,
+}
+
+// Enforce that the handoff data structure fits within the reserved 4KB region.
+const _: () = assert!(core::mem::size_of::<HandoffData>() <= 4096);
+
+// Enforce 4-byte alignment of the data structure.
```
swenson, line 90:
```suggestion
// Enforce 16-byte alignment of the data structure.
```

platforms/emulator/runtime/src/handoff.rs
clundin25: Done!

romtime/src/handoff.rs
clundin25: Done!

romtime/src/handoff.rs
clundin25: Done!

platforms/emulator/runtime/src/handoff.rs
clundin25: Hmm `StaticRef` seems incompatible with https://doc.rust-lang.org/reference/abi.html#the-link_section-attribute, since the linker is deciding where to place it

swenson: Ah, maybe not `StaticRef` then, but with this having `Copy` derived, it seems like this would get moved onto the stack here rather than actually referencing the static data?

rom/src/cold_boot.rs
```diff
@@ -843,6 +845,14 @@ impl BootFlow for ColdBoot {
                 ..Default::default()
             },
         );
+        let _ = fuse_state;
+
+        #[cfg(feature = "ocp-lock")]
+        if let Some(hek_state) = fuse_state.hek_state {
```
zhalvorsen: Even though the only thing in the handoff table is HEK state, I imagine we always want to write it so adding stuff later can be more seamless?

platforms/emulator/runtime/src/board.rs
```diff
@@ -785,7 +795,42 @@ pub unsafe fn main() {
...
+    let exit = if cfg!(feature = "test-handoff") {
+        debug!("Executing test-handoff");
+        #[cfg(feature = "test-handoff")]
+        {
+            if let Some(ho) = crate::handoff::HandOff::from_dccm() {
+                use romtime::ocp_lock::HekSeedState;
+                let state = ho.hek_state();
+                let ho_addr = ho.addr() as u32;
+                let expected_addr = 0x5000_0000;
+                if state.active_slot == 2
```
zhalvorsen: We should also probably check the FHT version number here. It will be good practice to always check the version number when first retrieving the handoff table to make sure the firmware knows how to interpret the underlying data

romtime/src/handoff.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub reserved_hek: [u32; 3], // 12 bytes
+
+    /// Padding to reach 64 bytes total.
+    pub padding: [u8; 44],
+}
+
+impl Default for RomHandoffTable {
+    fn default() -> Self {
+        Self {
+            fht_marker: 0,
```
zhalvorsen: For the default would it be best to fill in the constants like `marker`, and versions?

romtime/src/handoff.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/// Top-level handoff structure stored in DCCM.
+/// Resident at a well-known location in DCCM.
+///
+/// SAFETY: This structure MUST NOT exceed the reserved memory region size (4KB)
+/// defined in the linker scripts and memory map. Exceeding this size will cause
+/// memory corruption or linker errors.
+///
+/// ALIGNMENT: This structure is explicitly 16-byte aligned.
```
zhalvorsen: This comment says "explicitly" 16-byte aligned, but the attribute is only listing 4 byte aligned (`align(4)`)

romtime/src/handoff.rs
```diff
@@ -0,0 +1,132 @@
...
+// Enforce 16-byte alignment of the data structure.
+const _: () = assert!(core::mem::align_of::<HandoffData>() == 4);
+
+impl HandoffData {
+    /// Size of the handoff data structure.
+    pub const SIZE: usize = core::mem::size_of::<Self>();
+
+    /// Create a new handoff data structure with the given HEK state for ROM.
+    #[cfg(feature = "ocp-lock")]
+    pub fn new(hek_state: HekState) -> Self {
```
zhalvorsen: Would this be cleaner and more extendable if it took a struct as an argument, then you can just compile flag out the `hek_state` part of the input args?

romtime/src/handoff.rs
clundin25: This conflicts with: https://github.com/chipsalliance/caliptra-mcu-sw/pull/1262#discussion_r3156124119

Reading https://doc.rust-lang.org/reference/type-layout.html#r-layout.properties

It's unclear to me if `align(n)` MUST be aligned to `n` in bytes, or `n` is the x in 2 to the power of x

platforms/emulator/runtime/src/board.rs
clundin25: Done!

romtime/src/handoff.rs
clundin25: Done!

rom/src/cold_boot.rs
clundin25: Done!

romtime/src/handoff.rs
zhalvorsen: Ha ha that's funny. I read `Alignment is measured in bytes, and must be at least 1, and always a power of 2.` as it is given in bytes, but MUST be a power of 2. So you can't have align(3) because it isn't a power of 2. I tried this in a playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6803db4e3d21efc1c1ee29b6d9d55720 and I get this error when building:

```
error[E0589]: invalid `repr(align)` attribute: not a power of two
 --> src/main.rs:5:17
  |
5 | #[repr(C, align(3))]
  |                 ^
```

zhalvorsen: I could definitely be wrong though

zhalvorsen: cc @swenson

clundin25: Seems like definitive evidence that it is the raw bytes and not a power of 2 :)

.config/nextest.toml
```diff
@@ -14,6 +14,17 @@ store-failure-output = true
 failure-output = "immediate-final"
 fail-fast = false
 slow-timeout = { period = "30m", terminate-after = 6 }
+default-filter = """
```
ternst-nv, line 18: Are these failing on main? If so we should file an issue against for fixing them.

platforms/emulator/runtime/src/board.rs
```diff
@@ -301,6 +301,16 @@ pub unsafe fn main() {
     #[allow(static_mut_refs)]
     romtime::set_exiter(&mut EMULATOR_EXITER);
 
+    // Read handoff table BEFORE PMP setup.
+    let _handoff = unsafe { crate::handoff::HandOff::from_dccm() };
```
ternst-nv: [nit]: Can remove the underscore

platforms/emulator/runtime/src/handoff.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Marker for read-only access to the handoff table.
+pub struct ReadOnly;
+/// Marker for read-write access to the handoff table.
+#[allow(dead_code)]
+pub struct ReadWrite;
+
+/// Access to the handoff table.
+///
+/// The `Access` generic parameter enforces read-only or read-write capabilities at compile time.
+pub struct HandOff<Access> {
```
ternst-nv, line 16: It seems like we are copying the HekState out with this structure?  I'm a bit confused by it, why not have an accessor that can return a borrow or mutable borrow to the handoff table.  I guess in other words what does this wrapper buy us?

.config/nextest.toml
clundin25: Opened https://github.com/chipsalliance/caliptra-mcu-sw/issues/1335

platforms/emulator/runtime/src/board.rs
clundin25: Done!

platforms/emulator/runtime/src/handoff.rs
clundin25: Cleaned things up a bit & refactored to avoid copies.

The wrapper creates a safe read only API, and an unsafe mutable API, where a user that wants to write the handoff data MUST ensure there is no other references to the Handoff section.

rom/src/cold_boot.rs
```diff
@@ -843,6 +843,14 @@ impl BootFlow for ColdBoot {
                 ..Default::default()
             },
         );
+        let _ = fuse_state;
```
ternst-nv: [nit] Can this be removed?

common/config/src/lib.rs
clundin25: Had to re-add to place handoff section at end of DCCM reliably.

zhalvorsen (approved):

builder/src/rom.rs
```diff
@@ -201,6 +202,11 @@ SECTIONS
         PROVIDE(ESTACK_START = . );
     }
 
+    .handoff (NOLOAD) :
```
ternst-nv, line 205: This is going to place the handoff location at the top of all stack/data sections, however it will change location based on stack/data allocations.  Is that acceptable, or would we like to assign a specific address, like DCCM_TOP - SIZEOF(handoff_table)?

firmware-bundler/data/default-rom-layout.ld
```diff
@@ -59,6 +61,15 @@ SECTIONS
         . = ALIGN(4);
     } > RAM
 
+    /* We reserve 1 KB at the end of DCCM for the handoff table.
+       This must be 1KB aligned because the runtime requires DCCM regions
+       to be 1KB aligned for MEIVT (Machine External Interrupt Vector Table). */
+    . = HANDOFF_ADDR;
```
ternst-nv, line 67: This differs from how we've defined the hand off addr in the other linker files.

platforms/emulator/runtime/src/handoff.rs
```diff
@@ -0,0 +1,89 @@
...
+            );
+            return None;
+        }
+        Some(Self {
+            _access: PhantomData,
+        })
+    }
+}
+
+#[allow(dead_code)]
```
ternst-nv: Can this be removed?

platforms/emulator/runtime/src/handoff.rs
```diff
@@ -0,0 +1,89 @@
...
+                _access: PhantomData,
+            }),
+            None => None,
+        }
+    }
+}
+
+impl<Access> HandOff<Access> {
+    /// Get the address of the handoff table.
+    #[allow(dead_code)]
```
ternst-nv: Can this be removed?

ternst-nv (approved):
Left some comments about the linker/lints that need to be addressed, otherwise LGTM.

firmware-bundler/data/default-rom-layout.ld
clundin25: Going to split this into a follow up, will address this to match the other linker pattern. Tracking in https://github.com/chipsalliance/caliptra-mcu-sw/issues/1361

builder/src/rom.rs
clundin25: Will punt this one to a follow up as well. Would the tradeoff here be that you can accidentally clobber the memory regions if you are not careful? 

I don't have a very strong opinion here, since moving these locations in ROM should not happen frequently

clundin25: Tracking in https://github.com/chipsalliance/caliptra-mcu-sw/issues/1361

ternst-nv: Yeah, my big concern would be if in the event it does change location, the firmware may have to be implemented to support multiple locations depending on which ROM it is running against.  Keeping it to a single location means we wouldn't have to worry about that.

.config/nextest.toml
timothytrippel: @amit-kumarh were these recently fixed by you?

PR #1264, opened 2026-04-16
[ci] refactor fpga test workflow to use xtask flows

### LINKED ISSUE #1245: [xtask,ci] use xtask commands for running FPGA tests in CI workflows

timothytrippel:
This fixes #1245. 

Depends on #1263, only review the last commit.

### DIFF
--- a/.github/workflows/fpga.yml
+++ b/.github/workflows/fpga.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 jobs:
   build_test_binaries:
-    runs-on: [e2-standard-8]
+    runs-on: [e2-standard-8-big-disk]
     timeout-minutes: 60
     env:
       SCCACHE_VERSION: 0.10.0
@@ -118,6 +118,11 @@ jobs:
           path: /tmp/openssl.tar
           key: openssl-${{ env.CACHE_BUSTER }}
 
+      - name: Download bitstream
+        run: |
+          cargo xtask-fpga fpga download-bitstream
+          mv caliptra-bitstream.pdi /tmp/caliptra-bitstream.pdi
+
       - name: Build test firmware
         run: |
           # In the CI we really want to avoid lengthy build steps.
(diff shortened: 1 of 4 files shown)


.github/workflows/fpga.yml
```diff
@@ -229,30 +224,24 @@ jobs:
...
-
-          sudo CPTRA_FIRMWARE_BUNDLE="${PWD}/target/all-fw.zip" cargo-nextest nextest run \
-              "${COMMON_ARGS[@]}" \
-              --test-threads=${RUST_TEST_THREADS} \
-              --no-fail-fast
+            --cargo-metadata="${TEST_BIN}/target/nextest/cargo-metadata.json" \
+            --binaries-metadata="${TEST_BIN}/target/nextest/binaries-metadata.json" \
+            --target-dir-remap="${TEST_BIN}/target" \
+            --workspace-remap=. \
+            --message-format json > /tmp/nextest-list.json
```
clundin25, line 242: Could consider dropping this, I don't think we ever inspect the test list for this job.

I can't tell if this command is failing or the `xtask fpga test` command below, based on the logs.

timothytrippel: yeah wasn't sure if we wanted to keep it as a means to eventually run a sanity check for which FPGA tests were run so we don't run into a similar situation that happened with the emulator tests where they accidently get disabled?

timothytrippel: also happy to remove and only add that feature back in when it is completed E2E

clundin25: No urgency from my end to make any changes. Validating that all the tests we expect is something we should be doing. I'll cut a ticket

clundin25 (approved):
Awesome !

PR #1266, opened 2026-04-16
[userspace] Add Command Authorizer Trait for commands that need authorization

### LINKED ISSUE #897: Fuses: Authenticate commands
No auth layer exists. Spec mentions certificate-based and HMAC-based auth. At minimum need a stub/trait for authorization.

embediver:
Adds a `CommandAuthorizer` trait that is used to authorize every external command that needs authorization.
A stub is added to showcase this for a _rotate vendor PK hash_ command.

To give the implementer the option to embed the command, that is to be authorized, anywhere in the request, the handler logic was refactored to use slices for the request and response (instead of passing a `&mut` buffer and a lenght around).
This has the disadvantage, that a `MAX_DATA_SIZE` sized buffer for the response is allocated regardless of the command.
But since most command handlers used an intermediate buffer anyway this is no big deal imo. 

First step to resolve #897.

### DIFF
--- a/runtime/userspace/api/external-cmds-common/src/lib.rs
+++ b/runtime/userspace/api/external-cmds-common/src/lib.rs
@@ -6,6 +6,7 @@ extern crate alloc;
 
 use alloc::boxed::Box;
 use async_trait::async_trait;
+use caliptra_mcu_mbox_common::messages::CommandId;
 use zerocopy::{Immutable, IntoBytes};
 
 pub const MAX_FW_VERSION_LEN: usize = 32;
@@ -142,4 +143,8 @@ pub trait UnifiedCommandHandler {
         algorithm: u32,
         csr_data: &mut AttestedCsrData,
     ) -> Result<(), CommandError>;
+}
+
+pub trait CommandAuthorizer {
+    fn is_authorized(&self, cmd_id: CommandId, msg_buf: &mut [u8], req_len: usize) -> bool;
 }
--- a/runtime/userspace/api/external-cmds-common/src/lib.rs
+++ b/runtime/userspace/api/external-cmds-common/src/lib.rs
@@ -145,6 +145,24 @@ pub trait UnifiedCommandHandler {
     ) -> Result<(), CommandError>;
 }
 
+pub struct AuthorizationError;
+
 pub trait CommandAuthorizer {
-    fn is_authorized(&self, cmd_id: CommandId, msg_buf: &mut [u8], req_len: usize) -> bool;
+    /// Validates if a message is authorized.
+    ///
+    /// The request can contain authorization data (e.g. a HMAC).
+    /// This method is responsible for unpacking the contained
+    /// request message and returning it as a slice.
+    ///
+    /// # Arguments
+    /// * `cmd_id` - Command identifier
+    /// * `req` - Message to be authorized
+    ///
+    /// # Returns
+    /// * `Result<&[u8], CommandError>` - Unpacked command or Error
+    fn is_authorized<'a>(
+        &self,
+        cmd_id: CommandId,
+        req: &'a [u8],
+    ) -> Result<&'a [u8], AuthorizationError>;
 }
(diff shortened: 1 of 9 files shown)


runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -464,14 +403,14 @@ impl<'a> CmdInterface<'a> {
...
 
-    async fn handle_fw_version(
+    async fn handle_fw_version<'r>(
         &self,
-        msg_buf: &mut [u8],
-        req_len: usize,
-    ) -> Result<(usize, MbxCmdStatus), MsgHandlerError> {
+        req: &[u8],
+        resp_buf: &'r mut [u8],
+    ) -> Result<(&'r [u8], MbxCmdStatus), MsgHandlerError> {
```
zhalvorsen, line 410: I made a PR recently (#1256) that consolidates calculating the checksum after the handlers so we don't have to do it in each one. Can we have these return a mutable reference?

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -75,13 +72,14 @@ impl<'a> CmdInterface<'a> {
         let receive_result = self.transport.receive_request(msg_buf).await;
 
         let status = match receive_result {
-            Ok((cmd_id, req_len)) => {
+            Ok((cmd_id, req)) => {
                 // Process the request and prepare the response.
-                match self.process_request(msg_buf, cmd_id, req_len).await {
-                    Ok((resp_len, status)) => {
+                let mut resp_buf = [0u8; MAX_RESP_DATA_SIZE];
```
zhalvorsen, line 77: Like I mentioned on the call, it would probably be better in the long run to swap which buffer gets allocated on the stack. The responses have a potential to be really big, so it would be good to put the command on the stack instead.

runtime/userspace/api/external-cmds-common/src/lib.rs
```diff
@@ -143,3 +144,25 @@ pub trait UnifiedCommandHandler {
...
+    /// * `cmd_id` - Command identifier
+    /// * `req` - Message to be authorized
+    ///
+    /// # Returns
+    /// * `Result<& mut [u8], CommandError>` - Unpacked command or Error
+    fn is_authorized<'a>(
+        &self,
+        cmd_id: CommandId,
+        req: &'a [u8],
+    ) -> Result<&'a mut [u8], AuthorizationError>;
```
zhalvorsen, line 167: I think this is supposed to be a non-mutable reference right? Because it will be a subslice of the `req` input?

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
embediver: Currently requests sit at ~7508 bytes and responses at ~4284 bytes.
This can be easily observed by getting the size of the (`mcu_mbox::messages::`) `McuMailboxReq` and `McuMailboxResp` enums (which are supposed to have a fields for every response and request).

There are a couple of requests that can be quite large to, like AES encrypt.

I not so sure if swapping the buffers would be a benefit. Maybe just pass in both to be on the safe side?

embediver: I changed it to use two buffers now, if you have another idea just let me know.
Also I got rid of the `const` definition for the maximum command size and used `size_of::<McuMailboxResp>()`  which should be less error prone.

A way to only use one buffer could also be implemented by using some kind of managed buffer (like a `heapless::Vec`), I don't know if I like that though. In SPDM-lib a custom `MessageBuf` is implemented for such usecases, but its interface is very confusing and hard to use.

runtime/userspace/api/external-cmds-common/src/lib.rs
embediver: Jep, my mistake.

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
embediver: Rebased it.

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
zhalvorsen: We could switch to a dynamic buffer at some point maybe, but I think for now it would be best to have only one buffer for the stack savings. We can leave it like you had it before with the response getting allocated on the stack until response sizes end up growing passed requests.

zhalvorsen (approved):

PR #1267, opened 2026-04-16
dot: update lock acquire

ternst-nv:
Cherry-pick the 2.1 lock acquiring logic to main.  Also update a number of tests not yet on main-2.1 with the lock acquiring logic.


swenson (approved):
LGTM

PR #1270, opened 2026-04-16
rom: Add ROM milestone hooks

swenson:
Introduce a RomHooks trait and RomParameters::hooks field so integrators can observe major ROM milestones (cold/warm/fw-boot entry+exit, Caliptra boot, fuse writes to Caliptra, firmware load) without patching the common ROM. Hooks are opt-in (None by default); default trait methods are no-ops so integrators override only what they need.

Platform ROMs (emulator, FPGA) ship an example LoggingRomHooks implementation that prints a tagged line per milestone, gated behind a new 'test-rom-hooks' cargo feature. An integration test builds the emulator ROM with that feature and asserts all hooks reachable on the cold-boot happy path fire in the expected order.


mlvisaya (approved):
seems to be straightforward. LGTM!

timothytrippel (approved):

swenson:
Thanks!

platforms/emulator/rom/src/riscv.rs
```diff
@@ -52,6 +52,92 @@ impl DotRecoveryHandler for TestDotRecoveryHandler {
...
 }
 
+/// Example `RomHooks` implementation: prints a distinctive tag for each
+/// milestone so external observers (e.g. integration tests) can confirm
+/// that the hook fired. Also records which hooks fired as a bitmask in
+/// `mci_reg_fw_extended_error_info[0]` so a host-side test can verify
+/// completion even when the UART output buffer is drained on boot.
+/// Platforms that want to observe the boot flow (for debugging, timing,
+/// telemetry, ...) can drop in their own implementation the same way.
+struct LoggingRomHooks;
```
amit-kumarh, line 62: nit: What's the reason for defining these hooks for each platform? can we define it once in `platforms/common` or elsewhere?

swenson: Ah, yeah, we could have made them common, as logging / testing hooks, since we have tests that rely on setting those particular bits. I'll try to remember to address that in a future PR.

PR #1150, opened 2026-03-31
ci: Create backport + AI workflow

swenson:
For any PR merged to main, this will create a PR to main-2.1 with the change.

If the commit cherry picks cleanly, then Copilot (currently using the Opus 4.6 model) will check if there is anything else that should be done to adapt the commit.

If the commit does not cherry pick cleanly, then Copilot will try to resolve the merge conflicts and fix the PR.

The resultant PR will be assigned to the original author to review (or tag them if they cannot be assigned).

This is currently using a Copilot personal access token (PAT) for this action. An organization access token won't work, as it doesn't have the ability to make copilot requests.

Tested here: https://github.com/swenson/caliptra-mcu-sw/actions/runs/23807857975/job/69386316145

PR #1271, opened 2026-04-17
[CI]: AI enabled Cherry pick flow

clundin25:
Based on https://github.com/chipsalliance/caliptra-mcu-sw/pull/1150/

### DIFF
--- a/.github/workflows/cherry-picker.yml
+++ b/.github/workflows/cherry-picker.yml
@@ -1,0 +1,182 @@
+# Licensed under the Apache-2.0 license
+
+name: Backport Cherry-Pick with AI Assist
+
+on:
+  # TODO(clundin): Enable after testing
+  # pull_request:
+  #   types: [closed]
+  #   branches: [main]
+  workflow_dispatch:
+    inputs:
+      commit_sha:
+        description: 'Commit SHA to cherry-pick'
+        required: true
+      target_branch:
+        description: 'Target branch for cherry-pick'
+        required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/cherry-picker.yml
```diff
@@ -0,0 +1,182 @@
...
+          GIT_AUTHOR_NAME: "github-actions[bot]"
+          GIT_AUTHOR_EMAIL: "github-actions[bot]@users.noreply.github.com"
+          GIT_COMMITTER_NAME: "github-actions[bot]"
+          GIT_COMMITTER_EMAIL: "github-actions[bot]@users.noreply.github.com"
+          GIT_EDITOR: "true"
+        run: |
+          COMMIT_SHA="${{ steps.params.outputs.commit_sha }}"
+          TARGET_BRANCH="${{ steps.params.outputs.target_branch }}"
+
+          gemini --model gemini-3.0-pro "A cherry-pick of commit ${COMMIT_SHA} from 'main' onto \
```
swenson, line 139: nit: maybe we should make the `gemini-3.0-pro` model a variable above so that it is easier to change in the futurue

swenson (approved):
LGTM, one nit/suggestion

.github/workflows/cherry-picker.yml
clundin25: Ah thanks, that's a good point. I'll send that as a follow up. Going to merge this so I can trigger a flow and check if this is working

PR #1273, opened 2026-04-17
rom: program SS_PROD_DEBUG_UNLOCK_AUTH_PK_HASH registers

swenson:
Write SS_PROD_DEBUG_UNLOCK_AUTH_PK_HASH_REG_BANK_OFFSET (to the MCI register-bank offset of mci_reg_prod_debug_unlock_pk_hash_reg, 0x480) and SS_NUM_OF_PROD_DEBUG_UNLOCK_AUTH_PK_HASHES during fuse population, so Caliptra ROM can locate the PK hashes during prod debug unlock.

The hash count is configurable via a new RomParameters field prod_debug_unlock_auth_pk_hash_count: Option<u32>, defaulting to the number of entries in PROD_DEBUG_UNLOCK_PK_ENTRIES when None.


zhalvorsen (approved):

swenson:
Thanks!

romtime/src/mci.rs
```diff
@@ -5,6 +5,12 @@ use caliptra_mcu_registers_generated::mci;
 use caliptra_mcu_registers_generated::mci::bits::ResetRequest;
 use tock_registers::interfaces::{ReadWriteable, Readable, Writeable};
 
+/// Offset of `mci_reg_prod_debug_unlock_pk_hash_reg` within the MCI register
+/// bank. This value is programmed into
+/// `SS_PROD_DEBUG_UNLOCK_AUTH_PK_HASH_REG_BANK_OFFSET` so Caliptra ROM can
+/// locate the PK hashes via AXI at `MCI_BASE + offset + (level - 1) * 48`.
+pub const MCI_PROD_DEBUG_UNLOCK_PK_HASH_REG_BANK_OFFSET: u32 = 0x480;
```
amit-kumarh: Something to consider - is it worth fetching this programmatically so we don't have to modify this in subsequent HW revs?.

perhaps something like https://doc.rust-lang.org/core/mem/macro.offset_of.html ?

```
offset_of!(Mci, mci_reg_prod_debug_unlock_pk_hash_reg)
```

amit-kumarh (approved):

romtime/src/mci.rs
swenson: Oh, good call. I didn't see anything initially that was a good constant for this, but didn't think to use `offset_of!`. Will do.

rom/src/rom.rs
```diff
@@ -362,6 +362,26 @@ impl Soc {
...
+        // many are valid. The offset is the address of
+        // `mci_reg_prod_debug_unlock_pk_hash_reg` within the MCI register
+        // bank; Caliptra reads `MCI_BASE + offset + (level - 1) * 48` during
+        // prod debug unlock authentication.
+        let num_pk_hashes = params
+            .prod_debug_unlock_auth_pk_hash_count
+            .unwrap_or(PROD_DEBUG_UNLOCK_PK_ENTRIES.len() as u32);
+        caliptra_mcu_romtime::println!(
+            "[mcu-fuse-write] Setting prod debug unlock PK hash bank offset to {:x}, count {}",
+            caliptra_mcu_romtime::MCI_PROD_DEBUG_UNLOCK_PK_HASH_REG_BANK_OFFSET,
```
timothytrippel, line 375: to ensure a test exercises this, should we remove the hw model's  defaults here: https://github.com/chipsalliance/caliptra-mcu-sw/blob/d33e753d492e514b21b093f77c5360dca457e0d4/hw/model/src/lib.rs#L282?

(we can do in follow up)

timothytrippel (approved):

rom/src/rom.rs
swenson: Cool, I'm working on it.

swenson: Done, seemed to work well for me locally.

PR #1274, opened 2026-04-17
[CI]: Add bitstream manifest bundler tool

clundin25:
Bundles bitstreams into a well known tarball with a manifest.

### DIFF
--- a/.github/workflows/bitstream-build.yml
+++ b/.github/workflows/bitstream-build.yml
@@ -41,3 +41,46 @@ jobs:
             /tmp/*.xsa
             /tmp/*.pdi
           retention-days: 7
+
+      - name: Bundle bitstream
+        run: |
+          SS_COMMIT=$(git submodule status hw/caliptra-ss | awk '{print $1}' | tr -d '+-')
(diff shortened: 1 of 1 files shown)


.github/workflows/bitstream-build.yml
```diff
@@ -41,3 +41,46 @@ jobs:
...
+          if [ -n "$PDI_FILE" ]; then
+            PDI_ARG="--pdi-file $PDI_FILE"
+          fi
+
+          # Leave a copy for the uploader
+          mkdir manifest-bundle
+          
+          caliptra-bitstream-downloader bundle-manifest \
+            --repository "${{ github.repository }}" \
+            --hw-major-version "2.0" \
```
zhalvorsen, line 67: Should this come from a parameter so it will work with 2.1?

zhalvorsen (approved):

.github/workflows/bitstream-build.yml
clundin25: 2.1 has a separate job, so it's okay to hardcode in this case. When I cherry-pick it, I will hard code 2.1

PR #1281, opened 2026-04-20
runtime: track precise PMP entry usage in PMPRegionList

### LINKED ISSUE #1265: PMPRegionList capacity should be based on PMP entry count, not region count
### Problem

PMPRegionList in `runtime/kernel/veer/src/pmp.rs` has a hardcoded maximum of 16 logical regions. This limit is based on a worst-case assumption that every region uses TOR addressing mode, which consumes 2 PMP entries each (16 × 2 = 32 entries, filling the kernel's half of the 64 available PMP entries).

In practice, the emulator platform already defines up to 9 platform regions in `board.rs`, plus up to 7 MMIO regions from the MCU memory map — potentially reaching 16 total.
Any SoC vendor needing additional peripheral MMIO regions would be blocked by the region count limit despite having ample PMP entry budget remaining. The bottleneck is the region count, not the actual hardware resource.

### Proposed Solution

Replace the fixed 16-region cap with a limit based on actual PMP entry consumption:

1. Add an entry_count() method to PMPRegion that returns 2 for TOR-based variants and 1 for NAPOT-based variants.
2. Track the cumulative entry usage in PMPRegionList via a new total_entries field, and reject add_region() calls only when the 32-entry budget (AVAILABLE_ENTRIES - MPU_REGIONS * 2) would be exceeded.
3. Increase the PMPRegionList array size from 16 to MAX_KERNEL_ENTRIES (32) to support the theoretical maximum when all regions are NAPOT.
4. In VeeRProtectionMMLEPMP::new(), replace the conservative count * 2 estimate with the precise total_entries value for calculating kernel_start_entry.

This change is backward-compatible — existing platform configurations work identically with a more accurate capacity check. The User MPU / Kernel region split remains unchanged.

### Proposed patch

runtime/kernel/veer/src/pmp.rs
```
diff --git a/runtime/kernel/veer/src/pmp.rs b/runtime/kernel/veer/src/pmp.rs
index 5333c0d..1ad6629 100644
--- a/runtime/kernel/veer/src/pmp.rs
+++ b/runtime/kernel/veer/src/pmp.rs
@@ -16,6 +16,7 @@ use rv32i::pmp::{pmpcfg_octet, NAPOTRegionSpec, TORRegionSpec, TORUserPMP, TORUs
 
 const MPU_REGIONS: usize = 16;
 pub const AVAILABLE_ENTRIES: usize = 64;
+const MAX_KERNEL_ENTRIES: usize = AVAILABLE_ENTRIES - MPU_REGIONS * 2;
 
 pub type VeeRPMP = PMPUserMPU<MPU_REGIONS, VeeRProtectionMMLEPMP>;
 
@@ -109,32 +110,56 @@ pub enum PMPRegion {
     MachineMMIO(MMIORegion),
 }
 
+impl PMPRegion {
+    /// Returns the number of PMP hardware entries this region consumes.
+    /// TOR regions use 2 entries (start + end), NAPOT regions use 1 entry.
+    pub fn entry_count(&self) -> usize {
+        match self {
+            PMPRegion::KernelText(_) | PMPRegion::ReadOnly(_) | PMPRegion::Data(_) => 2,
+            PMPRegion::UserMMIO(_) | PMPRegion::MachineMMIO(_) => 1,
+        }
+    }
+}
+
 /// Configuration result containing all PMP regions in a simple list
+///
+/// The capacity is limited by the number of available kernel PMP entries
+/// (`MAX_KERNEL_ENTRIES`), not by an artificial region count. NAPOT regions
+/// consume 1 PMP entry each while TOR regions consume 2, so the maximum
+/// number of logical regions depends on the mix of addressing modes.
 pub struct PMPRegionList {
     /// Fixed-size array of regions (no heap allocation)
-    /// Size 16 assumes worst case that all regions are TOR regions (using 2 PMP entries each)
-    /// User regions: MPU_REGIONS, Kernel regions: ~3-4, total fits within AVAILABLE_ENTRIES
-    pub regions: [Option<PMPRegion>; 16],
-    /// Number of actual regions used
+    /// Size equals `MAX_KERNEL_ENTRIES` to support the theoretical maximum
+    /// (all NAPOT regions, 1 entry each).
+    pub regions: [Option<PMPRegion>; MAX_KERNEL_ENTRIES],
+    /// Number of logical regions used
     pub count: usize,
+    /// Total PMP hardware entries consumed (TOR=2, NAPOT=1 per region)
+    pub total_entries: usize,
 }
 
 impl PMPRegionList {
     /// Create a new empty region list
     pub fn new() -> Self {
         Self {
-            regions: [None; 16],
+            regions: [None; MAX_KERNEL_ENTRIES],
             count: 0,
+            total_entries: 0,
         }
     }
 
-    /// Add a region to the list
+    /// Add a region to the list.
+    ///
+    /// Returns `Err(())` if adding this region would exceed the available
+    /// kernel PMP entries (`MAX_KERNEL_ENTRIES`).
     pub fn add_region(&mut self, region: PMPRegion) -> Result<(), ()> {
-        if self.count >= 16 {
+        let entries_needed = region.entry_count();
+        if self.count >= MAX_KERNEL_ENTRIES || self.total_entries + entries_needed > MAX_KERNEL_ENTRIES {
             return Err(());
         }
         self.regions[self.count] = Some(region);
         self.count += 1;
+        self.total_entries += entries_needed;
         Ok(())
     }
 
@@ -193,7 +218,11 @@ impl fmt::Display for PMPRegion {
 
 impl fmt::Display for PMPRegionList {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        writeln!(f, "PMPRegionList ({} regions):", self.count)?;
+        writeln!(
+            f,
+            "PMPRegionList ({} regions, {}/{} kernel PMP entries used):",
+            self.count, self.total_entries, MAX_KERNEL_ENTRIES
+        )?;
         writeln!(f, "  Format: [MODE, RWX, LOCK] where:")?;
         writeln!(
             f,
@@ -277,17 +306,16 @@ impl VeeRProtectionMMLEPMP {
             reset_entry(i);
         }
 
-        // Conservative approach: assume worst-case that all regions are TOR regions (2 entries each)
-        let max_kernel_entries = pmp_regions.count * 2;
+        // Use precise entry counting: TOR regions consume 2 entries, NAPOT regions consume 1
+        let actual_kernel_entries = pmp_regions.total_entries;
 
         // Ensure we don't exceed available PMP entries (reserve MPU_REGIONS*2 for user MPU)
-        if max_kernel_entries > (AVAILABLE_ENTRIES - MPU_REGIONS * 2) {
-            return Err(()); // Too many kernel regions
+        if actual_kernel_entries > MAX_KERNEL_ENTRIES {
+            return Err(()); // Too many kernel entries
         }
 
         // Calculate starting entry for kernel regions (at the end of PMP entries)
-        // We'll use conservative allocation but only consume what we actually need
-        let kernel_start_entry = AVAILABLE_ENTRIES - max_kernel_entries;
+        let kernel_start_entry = AVAILABLE_ENTRIES - actual_kernel_entries;
 
         // Process regions from PMPRegionList in order, writing directly to PMP registers
         // This creates a 1:1 mapping from PMPRegionList to PMP hardware entries.

### PR DESCRIPTION
**Pull request by chunyanh-mrvl**
Fix #1265

## Summary
Refactors the kernel-side PMP region list (PMPRegionList) to track the
actual number of PMP hardware entries consumed (TOR=2, NAPOT=1) instead
of a worst-case count * 2 estimate. Frees PMP entries that were
over-reserved when NAPOT regions were used, and replaces the previous
hardcoded array size of 16 with MAX_KERNEL_REGIONS derived from the
hardware budget (MAX_KERNEL_ENTRIES = AVAILABLE_ENTRIES - MPU_REGIONS * 2).
Also tightens encapsulation so downstream code (VeeRProtectionMMLEPMP::new)
can rely on the total_entries == sum(region.entry_count()) invariant.
Scoped to runtime/kernel/veer/src/pmp.rs only; no other files change.
## Changes
runtime/kernel/veer/src/pmp.rs
- Add MAX_KERNEL_ENTRIES and MAX_KERNEL_REGIONS consts (module-private)
  that bound the kernel-side PMPRegionList — its array size and the PMP
  hardware entry budget.
- Add PMPRegion::entry_count() (const fn) reporting hardware entry cost
  per addressing mode (TOR=2, NAPOT=1).
- Make PMPRegionList fields private; add count() / total_entries()
  accessors, Default impl, and const fn new().
- Split add_region failure conditions into the array bound and the PMP
  entry budget for clarity.
- VeeRProtectionMMLEPMP::new now uses pmp_regions.total_entries()
  directly; the kernel start entry is computed precisely.
- Fix ReadOnlyRegion / DataRegion doc comments that incorrectly described
  them as NAPOT regions (they are TOR), and clarify ReadOnlyRegion's
  purpose post the CodeRegion → ReadOnlyRegion rename in #237.
## Test plan
- [ ] cargo xtask precheckin
- [ ] cargo xtask runtime

## Notes
- Boot stack: the PMPRegionList array grows 16→32 entries. (The earlier
  ~960 B figure also counted a platform-region buffer 32→64 change; that
  is no longer part of this PR since pmp_config.rs is unchanged, so the
  remaining increase is only the PMPRegionList growth.) Still a comfortable
  margin against the 8 KB emulator runtime stack.
- Public API additions: PMPRegion::entry_count(), PMPRegionList::count(),
  PMPRegionList::total_entries(), PMPRegionList::default().
  (MAX_KERNEL_ENTRIES / MAX_KERNEL_REGIONS are internal, not public.)
- Public API removals: PMPRegionList::regions / count / total_entries
  fields are no longer accessible directly (use accessors / iter()).

### DIFF
--- a/runtime/kernel/veer/src/pmp.rs
+++ b/runtime/kernel/veer/src/pmp.rs
@@ -16,6 +16,16 @@ use rv32i::pmp::{pmpcfg_octet, NAPOTRegionSpec, TORRegionSpec, TORUserPMP, TORUs
 
 const MPU_REGIONS: usize = 16;
 pub const AVAILABLE_ENTRIES: usize = 64;
+pub const MAX_KERNEL_ENTRIES: usize = AVAILABLE_ENTRIES - MPU_REGIONS * 2;
+
+/// Maximum number of logical kernel PMP regions that can fit in a
+/// `PMPRegionList`.
+///
+/// Equal to `MAX_KERNEL_ENTRIES` because each region consumes at least one
+/// PMP entry (NAPOT).  This is used to size the in-list array; the binding
+/// hardware constraint is the *entry* budget (`MAX_KERNEL_ENTRIES`), which
+/// `add_region` enforces precisely via `total_entries`.
+pub const MAX_KERNEL_REGIONS: usize = MAX_KERNEL_ENTRIES;
 
 pub type VeeRPMP = PMPUserMPU<MPU_REGIONS, VeeRProtectionMMLEPMP>;
 
@@ -74,15 +84,16 @@ fn write_pmpaddr_pmpcfg(i: usize, pmpcfg: u8, pmpaddr: usize) {
 // these memory regions as arguments. They further encode whether a region
 // must adhere to the `NAPOT` or `TOR` addressing mode constraints:
 
-/// The code (kernel + apps) RAM region address range.
+/// A read-only data region (e.g., the ROM area covering constants, strings
+/// and other non-executable read-only data from the code section).
 ///
-/// Configured in the PMP as a `NAPOT` region.
+/// Configured in the PMP as a `TOR` region with R-only permissions.
 #[derive(Copy, Clone, Debug)]
 pub struct ReadOnlyRegion(pub TORRegionSpec);
 
 /// The Data RAM region address range.
 ///
-/// Configured in the PMP as a `NAPOT` region.
+/// Configured in the PMP as a `TOR` region with R+W permissions.
 #[derive(Copy, Clone, Debug)]
 pub struct DataRegion(pub TORRegionSpec);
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by ArthurHeymans — APPROVED** [review]

File: runtime/kernel/veer/src/pmp.rs [hunk 1]
```diff
@@ -16,6 +16,16 @@ use rv32i::pmp::{pmpcfg_octet, NAPOTRegionSpec, TORRegionSpec, TORUserPMP, TORUs
 
 const MPU_REGIONS: usize = 16;
 pub const AVAILABLE_ENTRIES: usize = 64;
+const MAX_KERNEL_ENTRIES: usize = AVAILABLE_ENTRIES - MPU_REGIONS * 2;
```
mhatrevi, line 19: Can we add focused tests for the new PMPRegionList accounting behavior?

- A list with 32 NAPOT regions is accepted.
- A 33rd NAPOT region is rejected.
- A mixed TOR/NAPOT list whose cumulative entry count reaches exactly 32 is accepted.
- Adding one more region after reaching 32 entries is rejected.
- total_entries() == iter().map(PMPRegion::entry_count).sum() holds after successful additions.

chunyanh-mrvl: Done. Added focused unit tests for the `PMPRegionList` entry-budget accounting in `runtime/kernel/veer/src/pmp.rs`

- A list with 32 NAPOT regions is accepted → `accepts_napot_list_filling_entry_budget`
- A 33rd NAPOT region is rejected (and the list is left unchanged) → `rejects_napot_region_beyond_budget`
- A mixed list (14×TOR + 4×NAPOT) reaching exactly 32 entries is accepted → `accepts_mixed_list_reaching_exact_budget`
- Adding one more region after the 32-entry budget is full is rejected → `rejects_region_after_budget_is_exactly_full`
- `total_entries() == iter().map(PMPRegion::entry_count).sum()` after each successful add → `total_entries_equals_sum_of_entry_counts`


The two rejection tests intentionally cover both guards in `add_region`: the 33rd-NAPOT case trips the region-array bound (`count >= MAX_KERNEL_REGIONS`), while the mixed case (18 regions but 32 entries) trips the entry-budget guard (`total + needed > MAX_KERNEL_ENTRIES`) while the array still has room — so the entry-accounting path itself is exercised, not just the array bound.
To run these under host `cargo test`/`nextest`, two small test-only wiring changes were needed, both firmware-neutral:
- the `pmp` module is now compiled for tests as well: `#[cfg(any(target_arch = "riscv32", test))]`;
- `kernel` and `rv32i` are added as `[dev-dependencies]` (the crate's normal deps are riscv32 target-gated, so they aren't present on the host otherwise).
riscv32 firmware builds are unaffected: the crate stays `#![no_std]`, the `pmp` cfg is equivalent to the old gate when `test = false`, and dev-dependencies never enter a firmware build.
Run test: `cargo nextest run -p caliptra-mcu-tock-veer` (or `cargo test -p caliptra-mcu-tock-veer`)

mhatrevi (approved):

Issue #707, opened 2026-01-05
Add asynchronous support to OTP syscalls and driver


zhalvorsen:
The OTP driver in `rom/src/otp.rs` only supports blocking reads/writes to OTP. These may take a long time, especially writing.

We should add asynchronous support to both the driver and the syscall to avoid locking up the kernel or an app for the entire duration. This would require adding a subscribe/upcall infrastructure to the syscall added in #695.

PR #1286, opened 2026-04-20
[syscall] Add syscalls to read and write to the Caliptra SVN fuses

zhalvorsen:


runtime/kernel/capsules/src/otp.rs
clundin25, line 28: How did this end up here? OCP LOCK is only in 2.1?

runtime/kernel/capsules/src/otp.rs
```diff
@@ -58,14 +61,16 @@ impl Otp {
...
+            driver,
             total_heks,
             apps: grant,
         }
     }
 
     fn read_reg(&self, processid: ProcessId) -> CommandReturn {
         match self.apps.enter(processid, |app, _| match app.reg_offset {
+            // TODO: investigate using a cache instead of the actual fuses to reduce wear and
+            // increase performance
```
clundin25, line 73: Should we have a tracking bug?

runtime/kernel/capsules/src/otp.rs
zhalvorsen: It was cherry-picked from 2.1, but I figure if we are going to compress the branches at some point I might as well leave placeholders so they are consistent in at least this spot.

clundin25 (approved):

runtime/kernel/capsules/src/otp.rs
zhalvorsen: The more I think about it, the more I think this would be like https://github.com/chipsalliance/caliptra-mcu-sw/issues/707. I'll remove the comment.

zhalvorsen: I'm going to do it in a follow-on though

PR #1288, opened 2026-04-21
Hold Caliptra CPU until MCU ROM writes CPTRA_BOOT_GO

iyernaveenr:
On real hardware, Caliptra Core is held in reset until MCU ROM writes the CPTRA_BOOT_GO register in MCI. The emulator did not enforce this, stepping both CPUs unconditionally from the first cycle.

Wire the MCI CPTRA_BOOT_GO register write to a shared flag and skip the Caliptra CPU step until the flag is set. This applies to both the standalone emulator and SVP (which calls the same step path via the C binding).


swenson (approved):
LGTM

PR #1290, opened 2026-04-21
Replace process-wide emulator globals with per-instance state

iyernaveenr:
### Table of contents

- [Problem](#problem)
- [Solution](#solution)
- [Key properties](#key-properties)
- [C binding](#c-binding)
- [Regression test](#regression-test)
- [Files changed](#files-changed)
- [Local verification](#local-verification)

### Problem

The emulator used process-wide globals (`MCU_RUNNING`, `MCU_TICKS`,
`MCU_RUNTIME_STARTED`, `TICK_LOCK`, `TICK_COND`) in
`common/testing/src/lib.rs`. When two emulator instances run in the same
process they shared this state, so stopping one would stop the other
and concurrent step loops corrupted `MCU_TICKS`.

An earlier revision of this PR kept the globals as a silent fallback:
threads without per-instance state transparently read the globals. That
caused worker threads spawned **before** `init_emulator_state()` to
silently desync from the producer, hanging tests at the nextest
`SLOW`-to-`TIMEOUT` cap (~15 min in CI) instead of failing fast.

### Solution

Per-instance state is the **only** path. There is no process-wide
fallback.

1. `EmulatorState` bundles all per-instance coordination state
   (`running`, `ticks`, `runtime_started`, `tick_lock`, `tick_cond`).
   The struct is owned through `Arc<EmulatorState>` so worker threads
   can hold a clone for as long as they live.
2. `thread_local!` stores an `Option<Arc<EmulatorState>>` per thread.
   `init_emulator_state()` installs it on the calling thread.
3. The free functions (`is_emulator_running`, `sleep_emulator_ticks`,
   `wait_for_runtime_start`, ...) **panic** if called from a thread
   without an initialised state. There is no silent global fallback.
4. `spawn_with_emulator_state()` is `#[track_caller]` and panics at the
   spawn site if the parent thread has no state, so ordering bugs
   surface where they originate rather than as later worker desyncs.
5. Every emulator/model constructor (`Emulator::from_args_with_callbacks`,
   `ModelEmulated::new_unbooted`, `ModelFpgaRealtime::new_unbooted`)
   builds its own `Arc<EmulatorState>`, installs it on the calling
   thread **before** spawning any worker, then stores the `Arc` on the
   returned object.

### Key properties

- The five process-wide globals are gone. `TICK_NOTIFY_TICKS` is a
  `pub const`.
- Producers and consumers always agree on the same `Arc<EmulatorState>`.
- Ordering bugs panic at the offending call site instead of hanging.
- Worker thread state is captured at spawn time and outlives the
  worker because the constructor holds the parent `Arc`.

### C binding

`emulator_stop(handle)` stops a specific instance by writing
`state.running = false` through that instance's handle.

`emulator_trigger_exit()` preserves its documented "stops ALL
instances" semantics through a process-wide
`Mutex<Vec<Weak<EmulatorState>>>` registry. `emulator_init` pushes a
`Weak` reference into the registry; `emulator_trigger_exit` walks it
and calls `state.running.store(false, ...)` on every live instance.
Expired `Weak`s are pruned on registration.

### Regression test

New file `common/testing/tests/state_ordering.rs` pins the
panic-on-missing contract with three invariants:

- `spawn_before_init_panics_at_spawn_site` -- `spawn_with_emulator_state()`
  called without prior `init_emulator_state()` panics at the call site,
  not later in the worker.
- `free_fn_without_init_panics` -- `is_emulator_running()` and the
  other free functions panic when called from a thread without
  per-instance state.
- `init_before_spawn_propagates_per_instance_state` -- workers spawned
  after init observe their parent's per-instance writes (the original
  bug scenario, now a positive invariant).

A future regression to silent global fallback fails these tests fast in
CI rather than hanging.

### Files changed

Per-instance state and free-function rewrite:
- `common/testing/src/lib.rs`

Constructor-site state init:
- `emulator/app/src/emulator.rs`
- `emulator/app/src/main.rs`
- `hw/model/src/model_emulated.rs`
- `hw/model/src/model_fpga_realtime.rs`

C binding registry and ABI preservation:
- `emulator/cbinding/src/lib.rs`

Worker-spawn-site and free-function migration:
- `common/testing/src/i3c_socket.rs`
- `common/testing/src/i3c_socket_server.rs`
- `common/testing/src/mctp_util/common.rs`
- `common/testing/src/spdm_responder_validator/common.rs`
- `common/testing/src/spdm_responder_validator/doe.rs`
- `common/testing/src/spdm_responder_validator/mctp.rs`
- `emulator/app/src/doe_mbox_fsm.rs`
- `emulator/app/src/tests/*.rs`
- `tests/integration/src/*.rs`
- `hw/model/src/lib.rs`

Integration test scaffolding:
- `tests/integration/src/lib.rs` drops a caller-side
  `set_emulator_running(true)` at the top of `start_runtime_hw_model`
  that predated per-instance state. Each hw-model constructor already
  creates a fresh `EmulatorState` with `running = true` and installs it
  on the calling thread, so the line was a no-op under the new model
  and would panic because the test thread has no state until the model
  is built.

Regression test (new):
- `common/testing/tests/state_ordering.rs`

### Local verification

- `cargo check --workspace` -- OK
- `cargo clippy --workspace -- -D warnings --no-deps`
  (precheckin invocation, `RUSTFLAGS=-Cpanic=abort`) -- OK
- `cargo fmt --all --check` -- OK
- `cargo xtask cargo-lock | format | header-check | deps | clippy` --
  all OK
- `cargo test -p caliptra-mcu-testing-common --lib --tests` --
  12/12 pass
- `cargo test -p caliptra-mcu-testing-common --test state_ordering` --
  3/3 pass


common/testing/src/i3c_socket.rs
```diff
@@ -73,7 +73,7 @@ pub fn run_tests(
         );
         exit(-1);
     });
-    std::thread::spawn(move || {
+    crate::spawn_with_emulator_state(move || {
         wait_for_runtime_start();
         if !MCU_RUNNING.load(Ordering::Relaxed) {
```
swenson: It looks like the global `MCU_RUNNING` is still used here, though? Shouldn't we replace use of the globals here (and elsewhere with `spawn_with_emulator_state()`) with the thread-locals from the `EmulatorState`?

iyernaveenr: Good catch. The initial version only converted the thread spawns I introduced (spawn_with_emulator_state closures) but left the existing direct global accesses untouched, thinking backward compatibility required it. Your comment made me realize the thread-local should be authoritative everywhere, not just in the new code paths.

Fixed now. Replaced all direct global usages with thread-local equivalents across all files, removed the dual-write from step(), and refactored start_i3c_socket to use the thread-local internally instead of taking a `&'static AtomicBool` parameter.

The only remaining direct MCU_RUNNING usages are intentional:
- step() checks both per-instance and global as a fallback for emulator_trigger_exit()
- Ctrl-C handler sets the global
- emulator_trigger_exit() sets the global (global kill switch API)

tests/integration/src/lib.rs
```diff
@@ -1014,8 +1014,12 @@ mod test {
...
 
     pub fn start_runtime_hw_model(params: TestParams) -> DefaultHwModel {
-        // reset to known good state for beginning of test so that I3C socket will start correctly
-        MCU_RUNNING.store(true, Ordering::Relaxed);
+        // Each hw-model constructor now creates a fresh per-instance
+        // `EmulatorState` (with `running = true`) and installs it on the
+        // current thread via `init_emulator_state`, so no caller-side reset
+        // is required. Touching `set_emulator_running(true)` here would
+        // panic, since the test thread has no per-instance state until the
+        // model is built below.
```
swenson: nit: We don't need to provide comments like this, since it's not like we are seeing previous versions of the code when we look at it.
```suggestion
```

swenson (approved):
LGTM

PR #1292, opened 2026-04-21
rom: DOT header should be processed during FW update; add docs

swenson:
Adds docs/src/firmware_format.md describing the optional DOT header used by ROM for authorized, idempotent fuse and ownership-transfer actions, and links it from rom.md and SUMMARY.md.

Fixes the hitless firmware update path so it also processes a DOT manifest and sets FIRMWARE_BOOT_FLOW_COMPLETE before jumping to firmware, matching the cold-boot flow.

Adds an integration test (test_fw_manifest_dot_hitless_lock) that forces the ROM to re-enter via the hitless path and verifies DOT command processing still burns fuses. Uses a new ROM-only test-force-hitless-update feature to rewrite reset_reason at the end of cold boot. Enabled in the nightly-ci nextest filter.


mlvisaya (approved):
LGTM

swenson:
Thanks!

Issue #544, opened 2025-11-05
Support in-field prod PK hash rotation


zhalvorsen:
The firmware side of #536 to burn new PK hash values in the field and revoke old ones

PR #1296, opened 2026-04-23
Support in-field prod PK hash rotation

embediver:
#544

### DIFF
--- a/common/mcu-mbox/src/messages.rs
+++ b/common/mcu-mbox/src/messages.rs
@@ -21,6 +21,7 @@ pub use caliptra_api::mailbox::{
     MAX_CMB_DATA_SIZE,
 };
 pub use caliptra_api::{calc_checksum, verify_checksum};
+use caliptra_mcu_registers_generated::fuses::OTP_CPTRA_CORE_VENDOR_PK_HASH_0;
 use core::convert::From;
 use core::num::NonZeroU32;
 use zerocopy::{FromBytes, FromZeros, Immutable, IntoBytes, KnownLayout};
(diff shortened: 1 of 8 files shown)


runtime/userspace/syscall/src/otp.rs
```diff
@@ -38,6 +39,47 @@ impl<S: Syscalls> Otp<S> {
...
+        slot: u32,
+        new_hash: &[u8; OTP_CPTRA_CORE_VENDOR_PK_HASH_0.byte_size],
+    ) -> Result<(), ErrorCode> {
+        if slot > 15 {
+            return Err(ErrorCode::Invalid);
+        }
+
+        let valid_mask = self.read(reg::VENDOR_PK_HASH_VALID, 0)?;
+        let slot_used = (valid_mask & (1 << slot)) != 0;
+        if slot_used {
```
zhalvorsen: Feel free to do this on a follow up, but it could be good to check if the slot is valid and the requested `new_hash` is already burned into fuses return `Ok`. If the `new_hash` doesn't match return an error.

This would make the API idempotent.

runtime/userspace/syscall/src/otp.rs
```diff
@@ -38,6 +39,47 @@ impl<S: Syscalls> Otp<S> {
...
+            return Err(ErrorCode::Invalid);
+        }
+
+        let reg = reg::vendor_pk_hash_reg_by_slot(slot).ok_or(ErrorCode::Invalid)?;
+
+        // Write fuse
+        for (i, chunk) in new_hash.chunks(4).enumerate() {
```
zhalvorsen: I'd recommend using `chunks_exact` here so the compiler can make better assumptions about the size of the chunk later to avoid adding panic code

zhalvorsen (approved):

runtime/userspace/syscall/src/otp.rs
embediver: Done.

common/mcu-mbox/src/messages.rs
```diff
@@ -1441,6 +1442,30 @@ impl Default for ExportAttestedCsrResp {
 }
 impl McuResponseVarSize for ExportAttestedCsrResp {}
 
+/// MC_ROTATE_VENDOR_PK_HASH request: Rotate the vendor PK hash
+#[repr(C)]
+#[derive(Debug, IntoBytes, FromBytes, KnownLayout, Immutable, PartialEq, Eq)]
+pub struct RotateVendorPkHashReq {
```
zhalvorsen: We should probably rename this structure to correspond to the new API usage. Similar to the `MC_PROVISION_VENDOR_PK_HASH` constant name change

zhalvorsen (approved):

zhalvorsen (approved):

zhalvorsen (approved):

PR #1297, opened 2026-04-23
Enable spdm tests on fpga

parvathib:


.github/workflows/fpga.yml
```diff
@@ -161,28 +163,54 @@ jobs:
           tar xvf caliptra-test-binaries.tar.zst -C /tmp/caliptra-test-binaries/
           mksquashfs /tmp/caliptra-test-binaries /tmp/caliptra-test-binaries.sqsh -comp zstd
 
-      - name: 'Upload bitstream'
+      - name: Build spdm-emu for aarch64
+        if: github.event_name != 'pull_request'
```
swenson: Rather than adding all this conditional logic to the normal FPGA tests, I think it would make more sense to move this to a separate CI file, like we have for the normal SPDM validator -- we don't have that part of the normal build and test, after all.

parvathib: moved to a separate workflow file

swenson (approved):
LGTM

PR #1298, opened 2026-04-23
test_ecc_errors: fall back to building firmware if bundle is missing

swenson:
Makes test_rom_iccm_ecc_unc and test_rom_dccm_ecc_unc runnable without CPTRA_FIRMWARE_BUNDLE, matching the pattern used by other integration tests. Exposes TestBinaries and build_test_binaries from the test module so the rom submodule can reuse the existing build path.

### DIFF
--- a/tests/integration/src/lib.rs
+++ b/tests/integration/src/lib.rs
@@ -233,13 +233,13 @@ mod test {
         }
     }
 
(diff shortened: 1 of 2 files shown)


parvathib (approved):
LGTM

swenson:
Thanks!

PR #1299, opened 2026-04-23
compute the effective soc images based on the features

parvathib:
Fixes nightly tests


builder/src/all.rs
```diff
@@ -25,6 +25,8 @@ use zip::{
     ZipWriter,
 };
 
+use caliptra_mcu_flash_image::MCU_RT_IDENTIFIER;
```
swenson: nit: move the import to be above with its friends

swenson (approved):
LGTM, one nit

builder/src/all.rs
parvathib: done

PR #1302, opened 2026-04-23
[rom] remove read_fuses and its uses

amit-kumarh:
this function allocates many OTP fields onto the stack, using several kilobytes and causing overflows in several test cases

### DIFF
--- a/romtime/src/otp.rs
+++ b/romtime/src/otp.rs
@@ -4,7 +4,7 @@ use crate::{HexBytes, HexWord, StaticRef};
 use core::fmt::Write;
 use mcu_error::{McuError, McuResult};
 use registers_generated::fuses;
-use registers_generated::fuses::{FuseEntryInfo, Fuses};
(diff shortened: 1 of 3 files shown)


romtime/src/otp.rs
```diff
@@ -705,6 +705,16 @@ impl Otp {
         )
     }
 
+    /// Read a specific HEK seed partition from OTP.
+    /// index must be between 0 and 7.
+    pub fn read_hek_seed(&self, index: usize, data: &mut [u8; 48]) -> McuResult<()> {
+        if index > 7 {
+            return Err(McuError::ROM_OTP_INVALID_DATA_ERROR);
+        }
```
clundin25, line 713: Technically the spec allows up 16, but our reference has 8.

I guess integrators need to re-implement this code if they tweak the fuse map?

Either way it's the same as before, I just realized now

clundin25 (approved):

PR #1303, opened 2026-04-23
Revise Calitptra MCTP VDM commands

parvathib:
Revise and categorize Caliptra MCTP VDM commands

### DIFF
--- a/docs/src/external_mctp_vdm_cmds.md
+++ b/docs/src/external_mctp_vdm_cmds.md
@@ -74,13 +74,46 @@ The following table describes the commands defined under this specification. The
 | Device Capabilities           | 02h     | R   | Retrieve device capabilities.                       |
 | Device ID                     | 03h     | R   | Retrieve device ID.                                 |
 | Device Information            | 04h     | R   | Retrieve device information.                        |
-| Export CSR                    | 05h     | R   | Export CSR for device keys.                         |
-| Import Certificate            | 06h     | R   | Import CA-signed certificate.                       |
-| Get Certificate State         | 07h     | R   | Check the state of the signed certificate chain.    |
-| Get Log                       | 08h     | R   | Retrieve debug log or attestation log.              |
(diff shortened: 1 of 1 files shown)


helloxiling (approved):
Looks good to me. 
I also suggest checking if any existing cmd opcode in our code base should be modified to align with the spec.

parvathib:
> Looks good to me. I also suggest checking if any existing cmd opcode in our code base should be modified to align with the spec.

I've created an issue to correct the command codes in code once we finalize the VDM commands: https://github.com/chipsalliance/caliptra-mcu-sw/issues/1304

fdamato (approved):
Looks Good!

swenson (approved):

mlvisaya:
qq:  Does this doc apply to both MCTP VDM and SPDM VDM? I think we can't support Debug Unlock on MCTP VDM due to the MCTP buffer size. Instead, we're going to support it using SPDM VDM.

parvathib:
> qq: Does this doc apply to both MCTP VDM and SPDM VDM? I think we can't support Debug Unlock on MCTP VDM due to the MCTP buffer size. Instead, we're going to support it using SPDM VDM.

I understand the confusion. These commands listed here are generic to Caliptra and can be defined as Caliptra VDM commands that use SPDM as the transport, making them applicable to all integrators.
However, this PR is currently limited in scope and only revises the existing Caliptra MCTP VDM commands. I have not yet added these commands as generic Caliptra VDM commands.
For out-of-band (OOB) use cases, these commands would be issued as SPDM VDM commands over MCTP. For in-band (IB) use cases, the commands can use the mailbox transport.
I plan to add these details later, but for now, the scope of this PR is strictly to revise and consolidate the existing Caliptra MCTP VDM commands.

mlvisaya:
ok LGTM. I think the ideas of this PR was just to categorize the MCTP VDM commands.

mlvisaya (approved):
LGTM

Issue #959, opened 2026-02-25
[OCP-LOCK]: MCU ROM Documentation


clundin25:
Create documentation on how integrators should modify the ROM to work with OCP LOCK.

Things include:
- Setting the correct key release address.
- Writing HEK Seed fuses
- Reporting the HEK Seed state

clundin25:
Completed in https://github.com/chipsalliance/caliptra-mcu-sw/pull/1326

PR #1326, opened 2026-04-29
[OCP-LOCK]: Add Integration Guide

clundin25:
This closes https://github.com/chipsalliance/caliptra-mcu-sw/issues/959

### DIFF
--- a/docs/src/ocp_lock_integrator_guide.md
+++ b/docs/src/ocp_lock_integrator_guide.md
@@ -1,0 +1,101 @@
+# OCP LOCK Integrator Guide
+
+OCP LOCK (Layered Open-source Cryptographic Key management) is a specification for secure key management for Data-At-Rest protection in self-encrypting storage devices. This guide provides technical details for integrators to implement and customize the OCP LOCK functionality within the MCU project while maintaining compliance with the OCP LOCK specification.
+
+## Overview
+
+OCP LOCK acts as a Key Management Block (KMB), serving as the only entity capable of reading Media Encryption Keys (MEKs) and programming them into a vendor-implemented encryption engine. 
+
+### Core Components
+
+- **`romtime::ocp_lock`**: Contains platform-agnostic logic and types (defined in `romtime/src/ocp_lock.rs`).
+- **`Platform` Trait**: An interface for platform-specific behavior (e.g., slot count, state transitions).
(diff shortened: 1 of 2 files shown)


docs/src/ocp_lock_integrator_guide.md
```diff
@@ -0,0 +1,102 @@
...
+        otp: &crate::otp::Otp,
+        perma_bit: &PermaBitStatus,
+        seeds: &HekSeeds,
+    ) -> Result<usize, Error>;
+}
+```
+
+### HEK Seed States
+
+HEK Seed states are defined by the EKP spec. The MCU runtime uses this spec for attestation and state management. KMB uses a simplified state for communicating to KMB. The simplified states are defined in the OCP LOCK specification and are compatible with the expanded states defined in EKP.
```
zhalvorsen: Is this supposed to refer to the MCU?


```suggestion
HEK Seed states are defined by the EKP spec. The MCU runtime uses this spec for attestation and state management. KMB uses a simplified state for communicating to MCU. The simplified states are defined in the OCP LOCK specification and are compatible with the expanded states defined in EKP.
```

docs/src/ocp_lock_integrator_guide.md
```diff
@@ -0,0 +1,102 @@
...
+    Note right of CAL: HEK State registered for attestation
+
+    Note over ROM: Handoff Phase
+    ROM->>DCCM: Write HekState into HandoffData (Always)
+    Note right of DCCM: Persistent state for MCU Runtime
+```
+
+## Handoff Mechanism
+
+The OCP LOCK state is a critical part of the system's security posture. To ensure consistency, the ROM **always** writes the OCP LOCK state into the handoff table, even if OCP LOCK is not explicitly enabled or no valid HEK is found.
```
zhalvorsen: This isn't true anymore right `the ROM **always** writes the OCP LOCK state into the handoff table, even if OCP LOCK is not explicitly enabled`?

zhalvorsen (approved):

docs/src/ocp_lock_integrator_guide.md
clundin25: Whoops, thanks!

docs/src/ocp_lock_integrator_guide.md
clundin25: Cleaned up a little bit, will likely need to flesh this out a bit more when the runtime is further along.

Issue #998, opened 2026-03-03
recovery flow: squash CheckFwActivation and ActivateCheckRecoveryStatus states


swenson:
In the recovery flow in ROM, we check for recovery status 2 and then immediately wait for recovery status 1. This introduces a race condition, which can be avoided if we just skip the check for recovery status 2 altogether, or check for recovery status != 4.

PR #1327, opened 2026-04-29
recovery: Remove possible race condition (#998)

ternst-nv:
Remove the possibility of a race condition in the recovery state machine by expecting the `BootingImage` status prior to declaring done, or moving on to the next image.

Instead determine if the Caliptra OCP infrastructure is requesting another image, if so re-enter the image loop.  If the loading is complete move to done.


swenson (approved):
LGTM

PR #1328, opened 2026-04-29
image_loading: introduce DmaTransfer trait for FlashImageLoader

mlvisaya:
- Add DmaTransfer trait in dma_transfer.rs that abstracts source-to- destination DMA transfers, with DMAMapping as supertrait
- Update FlashImageLoader to be generic over DmaTransfer instead of DMAMapping, and simplify flash_load_image to delegate to the trait
- Add FlashReaderDma implementation in the emulator platform that preserves the existing buffered behavior (flash -> stack buffer -> DMA -> destination)
- Platforms with direct flash-to-AXI DMA can implement DmaTransfer with larger transfer sizes and no intermediate copy


swenson (approved):
LGTM

mlvisaya:
thanks!

PR #1333, opened 2026-04-29
fix(emulator): fix MCI reset reason handling for hitless firmware update

mlvisaya:
Track a snapshot of reset_reason at cycle completion (reason_at_cycle_complete) to detect when Caliptra writes new FW update bits via DMA between reset cycles. Previously, after the first boot completed, the reset_cycle_complete flag caused the emulator to treat freshly-written FW_HITLESS_UPD bits as stale, overwriting them with WARM_RESET. This caused the hitless update boot path to be skipped, hanging the firmware update test.


parvathib (approved):

mlvisaya:
thanks

PR #1338, opened 2026-04-30
[FPGA-Wrapper]: Make encryption engine RDY bit in ctrl register sticky

clundin25:


hw/fpga/src/caliptra_fpga_realtime_regs.rdl
```diff
@@ -260,7 +260,8 @@ regfile interface_regs {
     } ocp_lock_auxiliary_data_reg[8] @ 0x260;
 
     reg {
-        field { sw=rw; hw=na; desc = "OCP LOCK Control"; } ctrl[32] = 32'h0;
+        field { sw=r; hw=rw; desc = "OCP LOCK Ready bit. Sticky bit indicating the engine has processed the MEK."; } rdy[0:0] = 1'b0;
```
timothytrippel, line 263: As written, IIUC, the software will only be able to read this bit, and hardware will need to set/clear this bit. Is that intended?

timothytrippel (approved):

hw/fpga/src/caliptra_fpga_realtime_regs.rdl
clundin25: I think this is right. @gwangbae can you confirm?

PR #1339, opened 2026-04-30
Implement SPDM CHUNK_SEND handling

### LINKED ISSUE #1355: [SPDM_CERT_PROVISIONING] Add support for CHUNK_SEND for large requests

ArthurHeymans:
Fixes #1355

## Summary
- Add SPDM CHUNK_SEND / CHUNK_SEND_ACK request-response codes and dispatch
- Add responder-side large request context and CHUNK_SEND validation/reassembly
- Return CHUNK_SEND_ACK with EarlyErrorDetected and embedded ERROR for invalid chunk streams
- Process the reconstructed large request on the final chunk
- Advertise a larger max_spdm_msg_size while keeping data_transfer_size transport-limited

## Testing
- cargo test -p caliptra-mcu-spdm-lib
- cargo check -p caliptra-mcu-spdm-lib
- cargo check -p user-app


runtime/userspace/api/spdm-lib/src/context.rs
```diff
@@ -120,24 +122,66 @@ impl<'a> SpdmContext<'a> {
         Ok(())
     }
 
-    async fn handle_request(&mut self, buf: &mut MessageBuf<'a>) -> CommandResult<()> {
-        let req = buf;
+    pub(crate) async fn handle_request(&mut self, buf: &mut MessageBuf<'a>) -> CommandResult<()> {
```
parvathib: I think this scope change is unnecessary.

runtime/userspace/api/spdm-lib/src/context.rs
```diff
@@ -120,24 +122,66 @@ impl<'a> SpdmContext<'a> {
...
+            // Reset large request context if the next request is not a CHUNK_SEND.
+            self.large_req_context.reset();
+        }
+
         // Check for requests prohibited within session
         self.validate_request_in_session_context(req_code, req)?;
+        Ok((req_msg_header, req_code))
+    }
 
+    async fn handle_non_chunk_send_request(
```
parvathib: Could we name this `dispatch_request` or something similar?

runtime/userspace/api/spdm-lib/src/commands/chunk_send_ack_rsp.rs
```diff
@@ -0,0 +1,656 @@
...
+
+    if matches!(
+        request_code,
+        Some(code) if code == ReqRespCode::ChunkSend.into() || code == ReqRespCode::ChunkGet.into()
+    ) {
+        ctx.large_req_context.reset();
+        return encode_error_response_to_slice(version, ErrorCode::InvalidRequest, response);
+    }
+
+    if ctx.large_req_context.copy_message_to(req).is_err() {
```
parvathib, line 270: Assembled large request cannot be copied to request buffer. Request buffer is the transport buffer and would be small, and reassembled message would be always bigger in this scenario and copy will fail.

parvathib:
The protocol implementation looks good. 
Rather than staging the large request buffer inside the library, it's recommended to have the user space provide the buffer (similar to how PR #1351 refactors LargeResponseCtx to hold an app-provided &'a mut [u8].) 
We should handle the large request context the same way - the user app allocates the large request buffer and passes it in when constructing SpdmContext. 
This removes the hardcoded `MAX_MCTP_SPDM_MSG_SIZE` constant from the library, allowing the user/platform to manage the sizing and avoids the copy altogether (the reassembled large request can be dispatched directly from the referenced large request buffer without needing to copy into any intermediate buffer.)
we can address this as a separate follow up PR, applying the same app-provided buffer pattern in #1351 for LargeResponseCtx.

runtime/userspace/api/spdm-lib/src/commands/chunk_send_ack_rsp.rs
parvathib: lets address this in follow up PR. I created an issue for that: https://github.com/chipsalliance/caliptra-mcu-sw/issues/1357

parvathib (approved):

PR #1345, opened 2026-05-01
[OCP-LOCK]: Signing interface

clundin25:
Used to push signing details to be platform specific.


clundin25:
As a quick note, this interface is pretty inefficient (Duplicates the signature buffer, we should fix before adding ML-DSA support).

For now I want to punt on this, added https://github.com/chipsalliance/caliptra-mcu-sw/issues/1731 to track this as a p2

runtime/userspace/api/caliptra-api/src/ocp_lock.rs
```diff
@@ -65,6 +70,12 @@ pub struct Certificate<'a> {
     pub signature: BitStringRef<'a>,
 }
 
+#[async_trait]
+pub trait OcpLockSigner {
+    async fn sign(&self, label: &[u8], data: &[u8], signature: &mut [u8]) -> CaliptraApiResult<()>;
```
zhalvorsen, line 75: Can we have this return a DPE `Signature` type so the size and byte buffer are less arbitrary?

runtime/userspace/api/caliptra-api/src/signer.rs
```diff
@@ -0,0 +1,65 @@
...
+use async_trait::async_trait;
+use core::mem::size_of;
+use dpe::commands::{Command, SignP384Cmd};
+use dpe::response::SignP384Resp;
+#[async_trait]
+pub trait DpeTransport: Send + Sync {
+    async fn invoke(&self, cmd: &Command, resp_buf: &mut [u8]) -> CaliptraApiResult<usize>;
+}
+use crate::ocp_lock::OcpLockSigner;
+use zerocopy::TryFromBytes;
```
zhalvorsen: uber nit: but can you move this to below the other `use` statements?


```suggestion
use crate::ocp_lock::OcpLockSigner;
use zerocopy::TryFromBytes;

#[async_trait]
pub trait DpeTransport: Send + Sync {
    async fn invoke(&self, cmd: &Command, resp_buf: &mut [u8]) -> CaliptraApiResult<usize>;
}
```

runtime/userspace/api/caliptra-api/src/signer.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        let mut resp_buf = [0u8; size_of::<SignP384Resp>()];
+        let len = self.transport.invoke(&command, &mut resp_buf).await?;
+
+        let dpe_resp = SignP384Resp::try_ref_from_bytes(&resp_buf[..len])
+            .map_err(|_| CaliptraApiError::InvalidResponse)?;
+
+        if signature.len() < 96 {
+            return Err(CaliptraApiError::InvalidArgBufferTooSmall);
+        }
```
zhalvorsen: It would be good to check this at the beginning of the function

zhalvorsen (approved):

runtime/userspace/api/caliptra-api/src/ocp_lock.rs
clundin25: Going to do this in a follow up. I need to change DPE & that includes some API changes since the active one.

runtime/userspace/api/caliptra-api/src/signer.rs
clundin25: Done!

runtime/userspace/api/caliptra-api/src/signer.rs
clundin25: Done!

zhalvorsen (approved):

PR #1348, opened 2026-05-01
[auth-cmds] Add a command to revoke a vendor FW verification key

### LINKED ISSUE #1259: Add runtime command to revoke vendor firmware verification keys

zhalvorsen:
This checks the key to be revoked was not used for boot. This is a form of proof of possession check to ensure other PKs are valid before revocation.

Tests will be added in a follow-on change.

### DIFF
--- a/romtime/src/otp.rs
+++ b/romtime/src/otp.rs
@@ -512,7 +512,8 @@ impl Otp {
 
     /// Read vendor public key hash valid.
     pub fn read_vendor_pk_hash_valid(&self) -> McuResult<u32> {
-        self.read_entry(fuses::VENDOR_PK_HASH_VALID)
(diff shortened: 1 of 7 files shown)


romtime/src/otp.rs
```diff
@@ -567,6 +568,24 @@ impl Otp {
         self.read_entry(entry)
     }
 
+    /// Read vendor ECC revocation (4 bytes).
+    pub fn write_vendor_ecc_revocation(&self, index: usize, value: u32) -> McuResult<()> {
```
clundin25: nit: Docstring says read, but the function seems to be writing

clundin25 (approved):

romtime/src/otp.rs
zhalvorsen: Good catch, thanks!

Issue #872, opened 2026-02-11
[provisioning] develop reference fuse configuration MCU programs


timothytrippel:
To write fuses for the first time during manufacturing we should develop reference programs that can be loaded into MCU SRAM that can write fuses using the fuse controller. We should develop two such programs to:
- [ ] run in the `test_unlocked0` LC state to write the test tokens
- [ ] run in the `manuf` LC state to write the remaining fuses

See this [table](https://chipsalliance.github.io/caliptra-mcu-sw/provisioning.html#lifecycle-state-constraints-for-provisioning-fuses) for which fuses should be written in which program.

Each program should be created by writing a single wrapper binary that iterates over a hash map of the fuse field/value pairs---contained in a static library that is autogenerated with tooling described in #871---and writes each value to it's respective fuse field.

Each binary should be buildable using the xtask flows for building firmware. Here is a good example of building a hello-world standalone MCU binary to get started: https://github.com/cfrantz/caliptra-mcu-sw/commit/b211ac20f46576007042b0d5803b3362ec5dcfca

PR #1350, opened 2026-05-01
[doc] IdevID cert provisioning

mlvisaya:


docs/src/idevid-cert-provisioning.md
```diff
@@ -0,0 +1,224 @@
...
+## Certificate Storage Model
+
+### Fuse Storage Options
+
+
+The IDevID certificate is reconstructed at boot using a **template-based approach**, as recommended by the [Caliptra 2.0 specification](https://chipsalliance.github.io/Caliptra/2.0/specification/HEAD/#idevid-certificate):
+
+> *"It is recommended that the vendor define an IDevID certificate template such that the SoC at runtime can reconstruct the same certificate that the pCA endorsed. The SoC is recommended to store the IDevID certificate signature in fuses and the IDevID certificate template in the firmware image."*
+
+The certificate **signature** must reside in immutable storage (OTP fuses) to prevent a compromised firmware update from substituting a forged certificate. The **TBS template** is stored in the firmware binary image, and **device-specific fields** (public key, serial number, Subject Key Identifier) are derived at boot from the IDevID public key retrieved via Caliptra mailbox commands.
```
timothytrippel, line 52: if we are going to be recommending storing cert templates in fuses, then I think our upstream reference fuse map should be updated to accommodate this.

we could expand: https://github.com/chipsalliance/caliptra-ss/issues/982?

docs/src/idevid-cert-provisioning.md
```diff
@@ -0,0 +1,224 @@
...
+
+| Algorithm | Fuse Storage | Binary Image | Caliptra Command (boot) |
+|-----------|-------------|-------------|------------------------|
+| **ECC P384** | Signature (R ‖ S) = **96 bytes** | TBS template | `GET_IDEV_ECC384_INFO` → public key |
+| **MLDSA-87** | Signature = **4,627 bytes** (see note) | TBS template | `GET_IDEV_MLDSA87_INFO` → public key |
+
+> **Note**: It is the **integrator's responsibility** to allocate sufficient fuse space to store the signature — whether by expanding the OTP controller partitions, provisioning external SoC fuses, or using another immutable storage mechanism appropriate for their platform (e.g. EPROM).
```
timothytrippel, line 59: do we have a guide that explains how an integrator can expand the reference fuse map? i.e. a script that can be run after updating https://github.com/chipsalliance/caliptra-ss/blob/main/src/fuse_ctrl/data/otp_ctrl_mmap.hjson ?

@ekarabu does this exist?

docs/src/idevid-cert-provisioning.md
```diff
@@ -0,0 +1,224 @@
...
+
+
+| Certificate Field | Value | Source | When Available |
+|---|---|---|---|
+| Version | v3 (2) | Static | Template |
+| Serial Number (20B) | SHA256(DER(pubkey)) last 20B, masked | Derived from public key | Boot (computed) |
+| Issuer Name | pCA's Subject Name | Static per vendor | Template |
+| Validity notBefore | Vendor-defined | Static per vendor | Template |
+| Validity notAfter | 99991231235959Z | Static | Template |
+| Subject Name CN | "Caliptra 1.0 IDevID" | Static | Template |
```
timothytrippel, line 71: should this be 2.x?

docs/src/idevid-cert-provisioning.md
```diff
@@ -0,0 +1,224 @@
...
+| Subject Name serialNumber | Hex(SHA256(DER(pubkey))) | Derived from public key | Boot (computed) |
```
timothytrippel, line 72: what is the order of the hexstring? BE or LE?

docs/src/idevid-cert-provisioning.md
```diff
@@ -0,0 +1,224 @@
...
+
+> **Note**: It is the **integrator's responsibility** to allocate sufficient fuse space to store the signature — whether by expanding the OTP controller partitions, provisioning external SoC fuses, or using another immutable storage mechanism appropriate for their platform (e.g. EPROM).
+
+### Reconstruction Approach
+
+
+| Certificate Field | Value | Source | When Available |
+|---|---|---|---|
+| Version | v3 (2) | Static | Template |
+| Serial Number (20B) | SHA256(DER(pubkey)) last 20B, masked | Derived from public key | Boot (computed) |
```
timothytrippel, line 67: What does "masked" mean here? is this your way of signifying the you will set the MSB of the value to ensure this is fixed size? (since this field is technically an ASN.1 integer which can be variable sized)

docs/src/idevid-cert-provisioning.md
```diff
@@ -0,0 +1,224 @@
...
+|-----------|-------------|-------------|------------------------|
+| **ECC P384** | Signature (R ‖ S) = **96 bytes** | TBS template | `GET_IDEV_ECC384_INFO` → public key |
+| **MLDSA-87** | Signature = **4,627 bytes** (see note) | TBS template | `GET_IDEV_MLDSA87_INFO` → public key |
+
+> **Note**: It is the **integrator's responsibility** to allocate sufficient fuse space to store the signature — whether by expanding the OTP controller partitions, provisioning external SoC fuses, or using another immutable storage mechanism appropriate for their platform (e.g. EPROM).
+
+### Reconstruction Approach
+
+
+| Certificate Field | Value | Source | When Available |
```
timothytrippel, line 64: this just shows the ECDSA cert. is there an ML-DSA cert too?

docs/src/idevid-cert-provisioning.md
timothytrippel: I.e., this: https://github.com/chipsalliance/caliptra-sw/blob/a38c99b60c97df406201e098df1b89e9301a26d1/common/src/x509.rs#L125?

docs/src/idevid-cert-provisioning.md
```diff
@@ -0,0 +1,224 @@
...
+
+---
+
+## Manufacturing Provisioning Flow
+
+### Prerequisites
+
+- Device lifecycle is in **Manufacturing mode** (IDevID CSR generation only works in this state)
+- UDS and IDevID certificate attribute fuses have been programmed
+- Caliptra firmware is loaded and running
```
timothytrippel, line 91: Why does the Caliptra Core FW need to be loaded? Can't the Caliptra Core ROM support producing the CSRs (https://github.com/chipsalliance/caliptra-sw/blob/a38c99b60c97df406201e098df1b89e9301a26d1/rom/dev/src/flow/cold_reset/idev_id.rs#L467)?

We should avoid loading any firmware to enable provisioning into Caliptra Core or MCU SRAMs over OCP recovery as this is not an ATE-friendly protocol.

docs/src/idevid-cert-provisioning.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Sequence Diagram
+
+
+
+```mermaid
+sequenceDiagram
```
timothytrippel, line 98: we need to completely update this flow to remove the MCU runtime firmware from the picture to be ATE friendly.

My suggestion would be to follow the same flow used for manufacturing fuse provisioning (https://github.com/chipsalliance/caliptra-mcu-sw/issues/872), i.e., write a dedicated bare metal binary that can be loaded into MCU SRAM over JTAG before any other firmware is loaded via OCP recovery to perform the bulk of the CSR provisioning, and waits just before the IDevID CSR is read out of the Caliptra Core mailbox, then the CSR can be read out over JTAG as well (ATE friendly). Lastly, the bare metal binary can wait for the signature to come back (over JTAG) to be written to fuses (which could be an optional step, depending on the integrator).

Wdty?

PR #1351, opened 2026-05-01
spdm-lib: Refactor large VDM responses and Caliptra VDM command handler

parvathib:
- Use app-provided static buffer to stage buffered responses served in chunks (streaming responses like certificates use a separate path)
- Rename crate external-cmds-common to caliptra-mcu-common-commands and define and use OCP error codes for common Caliptra commands
- Rename trait UnifiedCommandHandler to CaliptraCmdHandler


mlvisaya (approved):
LGTM

PR #1353, opened 2026-05-04
spdm: Refactor SpdmCertStore for Cert slot management

parvathib:
- Adds trait methods for setting and erasing certificate in a slot
- Make the parameter ordering consistent across all methods
- Add documentation for Certificate store

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -31,5 +31,6 @@
 - [Caliptra Utility Host Library](./caliptra_util_host_library.md)
 - [Network Boot Recovery](./network_boot.md)
 - [Integrator's Guide](./integrator-guide.md)
+    - [SPDM Integration Guide](./spdm-integrator-guide.md)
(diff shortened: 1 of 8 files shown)


docs/src/spdm.md
```diff
@@ -315,4 +315,61 @@ pub trait SpdmSecureSessionManager {
     /// - `Result<(), SpdmError>`: Returns `Ok(())` if the secure message is encoded successfully, or an error code.
     async fn encode_secure_message(&self, response: &mut [u8]) -> Result<(), SpdmError>;
 }
+```
+
+## Integration guide
```
swenson: I wonder if we should move this into the integrator's guide, as a separate sub-document?

swenson (approved):
LGTM

docs/src/spdm.md
parvathib: good idea, I'll put this under integrator's guide as sub document.

PR #1359, opened 2026-05-05
[caliptra-spdm-vdm-client] Add host-side SPDM VDM test framework

parvathib:
- `caliptra-vdm`: Caliptra VDM command protocol and common payload validation
- `caliptra-spdm-requester`: SPDM requester context and CaliptraVdmTransport using SPDM-Utils libspdm rust wrappers
- `caliptra-spdm-vdm-client`: client lib + caliptra-spdm-validator binary
- Integration test for `ExportAttestedCsr` over SPDM/MCTP in the workflow


mlvisaya:
ideally we should implement the spdm_vdm as a new transport under caliptra-util-host/transport/src/transports, and the new caliptra-vdm commands will be added under command/src/api/ .  The main idea was same command can be dispatched to multiple transports. For example, get_device_id can be sent via MCU mailbox, or via SPDM VDM, or MCTP VDM. In the upper layer user will just call the API get_device_id and then specify a transport to use. I think the contents of caliptra-vdm need to be moved to the folder i listed above.

parvathib:
> ideally we should implement the spdm_vdm as a new transport under caliptra-util-host/transport/src/transports, and the new caliptra-vdm commands will be added under command/src/api/ . The main idea was same command can be dispatched to multiple transports. For example, get_device_id can be sent via MCU mailbox, or via SPDM VDM, or MCTP VDM. In the upper layer user will just call the API get_device_id and then specify a transport to use. I think the contents of caliptra-vdm need to be moved to the folder i listed above.

Agreed! I've moved the SPDM VDM transport implementation under `caliptra-util-host/transport/src/transports/spdm_vdm/` and the shared command APIs under `commands/src/api/`. The dispatch now works
as you described: same API (e.g., get_device_id, export_attested_csr) routes through whichever transport the user selects at session creation.

One clarification on transport scope: the common Caliptra command set will be shared between Mailbox and SPDM VDM only (no duplication across two OOB transports). For existing MCTP VDM
device-info commands, the path forward depends on standardization: if a command is accepted as a common Caliptra command, it moves to SPDM VDM; if it remains vendor-specific, it stays
in MCTP VDM. Each transport's dispatch table declares what it supports, so unsupported commands fail cleanly at the transport layer.

caliptra-util-host/run-validator-tests.sh
mlvisaya, line 1: please add this to caliptra-util-host/xtask. instead of creating a bash script. 

there's already a builder script there.

caliptra-mcu-sw/caliptra-util-host$ cargo xtask --help

Usage: caliptra-util-host-xtask <COMMAND>

Commands:
  build       Build the caliptra-util-host library and its components
  test        Run tests for the caliptra-util-host library
  clean       Clean build artifacts
  fmt         Format code
  clippy      Run clippy lints
  cbindings   Generate C bindings
  check       Run comprehensive checks (build, test, format, clippy)
  precheckin  Run pre-check-in checks (format, clippy, build)
  server      Run the mailbox server
  validator   Run the mailbox client validator
  help        Print this message or the help of the given subcommand(s)

Options:
  -h, --help
          Print help (see a summary with '-h')

  -V, --version
          Print version

and the validator to run.

parvathib: shell script was a mistake, i'll remove. I've added the prerequisite steps to workflow file already. I can add xtask commands in the follow up PR if needed.

mlvisaya (approved):
lgtm

PR #1363, opened 2026-05-05
testing: Fix non_blocking race condition

ternst-nv:
The i3c socket allowed both sender and receiver to modify the non-blocking nature of the tcp file descriptor.  Since these run in different threads, the behavior of the socket could change unexpectedly since the non_blocking nature was not locked.

Switch to instead support a mechanism to read/write exact while remaining in non-blocking mode to avoid the race condition.


amit-kumarh (approved):
lgtm, thanks for the fix!

PR #1364, opened 2026-05-06
Add revoke vendor PK hash mcu-mbox command

embediver:
Adds a MC_FUSE_REVOKE_VENDOR_PK_HASH ("RVKH") command to invalidate a specific vendor PK hash slot.
If the a slot is not provisioned with a hash an error is returned.

Part of #544.

### DIFF
--- a/runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
+++ b/runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
@@ -16,9 +16,10 @@ use caliptra_mcu_mbox_common::messages::{
     CommandId, DeviceCapsReq, DeviceCapsResp, DeviceIdReq, DeviceIdResp, DeviceInfoReq,
     DeviceInfoResp, ExportAttestedCsrReq, ExportAttestedCsrResp, FirmwareVersionReq,
     FirmwareVersionResp, FuseIncreaseCaliptraMinSvnReq, FuseIncreaseCaliptraMinSvnResp,
-    FuseRevokeVendorPubKeyReq, FuseRevokeVendorPubKeyResp, FuseWriteResp, GetAuthCmdChallengeReq,
-    GetAuthCmdChallengeResp, MailboxRespHeader, MailboxRespHeaderVarSize, McuAesDecryptInitReq,
-    McuAesDecryptUpdateReq, McuAesEncryptInitReq, McuAesEncryptUpdateReq, McuAesGcmDecryptFinalReq,
+    FuseRevokeVendorPkHashReq, FuseRevokeVendorPkHashResp, FuseRevokeVendorPubKeyReq,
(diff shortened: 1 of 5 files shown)


runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -915,6 +920,50 @@ impl<'a> CmdInterface<'a> {
...
+    ) -> Result<(&'r mut [u8], MbxCmdStatus), MsgHandlerError> {
+        // Decode the request
+        let req = FuseRevokeVendorPkHashReq::ref_from_bytes(req)
+            .map_err(|_| MsgHandlerError::InvalidParams)?;
+        let (resp, _) = FuseRevokeVendorPkHashResp::mut_from_prefix(resp_buf)
+            .map_err(|_| MsgHandlerError::InvalidParams)?;
+
+        // Check the given slot has a valid PK hash provisioned
+        let otp = otp::Otp::<DefaultSyscalls>::new();
+        if !otp.valid_vendor_pk_hash_slot(req.vendor_pk_hash_slot) {
```
zhalvorsen: This function only checks if that PK hash isn't revoked, should we have a separate function to check if a slot is not all zeros? I guess that check is also already being done in syscall function so we could also drop this check from here

runtime/userspace/syscall/src/otp.rs
```diff
@@ -117,6 +117,30 @@ impl<S: Syscalls> Otp<S> {
...
+
+        // Check if the slot is provisioned to not burn an empty slot
+        let pk_hash = self.read_vendor_pk_hash(vendor_pk_hash_slot)?;
+        if pk_hash.iter().eq(repeat(&0)) {
+            Err(ErrorCode::Invalid)?
+        }
+
+        let valid_mask = self.read(reg::VENDOR_PK_HASH_VALID, 0)?;
+
+        let valid_mask = valid_mask ^ (1 << vendor_pk_hash_slot);
```
zhalvorsen: nit: this might just be personal preference, but I think its easier to model mentally with a bitwise OR here. Feel free to push back if you disagree:


```suggestion
        let valid_mask = valid_mask | (1 << vendor_pk_hash_slot);
```

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
embediver: Good catch!
Updated the doc comment of the `valid_vendor_pk_hash_slot()` function since it returns false for invalid slots, as well as revoked ones.
Dropping it completely since its done in the syscall and would cause an error for revoking a already revoked slot.

runtime/userspace/syscall/src/otp.rs
embediver: Sure, OR is simpler here, no idea why I put an XOR.

zhalvorsen (approved):

PR #1370, opened 2026-05-07
Add ExportIdevidCsr Caliptra VDM command

parvathib:
Add a new Caliptra VDM command (ExportIdevidCsr) that exports the IDevID CSR over the SPDM VDM transport, using the large response chunking infrastructure.

Manufacturing mode: returns IDevID CSR for requested algorithm.
Production mode: rejects with error (negative test).

### DIFF
--- a/caliptra-util-host/apps/spdm/test-config.toml
+++ b/caliptra-util-host/apps/spdm/test-config.toml
@@ -2,6 +2,8 @@
 
 # SPDM Client Test Configuration
 
+mode = "production"
+
 [network]
 # Server address for the SPDM-over-MCTP bridge (emulator I3C controller socket)
 server_address = "127.0.0.1:63333"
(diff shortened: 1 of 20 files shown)


caliptra-util-host/apps/spdm/test-config.toml
```diff
@@ -16,5 +18,9 @@ verbose_output = false
 
 [export_attested_csr]
 # ECC384 = 1, MlDsa87 = 2
-key_ids = [0, 1, 2]
+key_ids = [1, 2, 3]
```
swenson, line 21: Does the comment match what these key IDs are? I thought the comment was for the algorithms (as below)?

swenson (approved):
LGTM, one question about a comment

caliptra-util-host/apps/spdm/test-config.toml
parvathib: that is correct. Let me update the position of comment.

parvathib: done.

PR #1371, opened 2026-05-07
spdm-lib: Unify large request/response into shared LargeMessageCtx

parvathib:
- Combine LargeRequestCtx and LargeResponseCtx into a single
      LargeMessageCtx using one app-provided buffer.
- Eliminate the 1KB intermediate buffer in chunk_send_ack response generation
- Reduce the user-app stack reservation (0xAE00 -> 0xA000) to address the Process load error


mlvisaya (approved):
LGTM

PR #1375, opened 2026-05-07
regs: Disable stub warning by default

ternst-nv:
The emulator app disables stubwarnings by default, but tests which directly invoke the emulator default to true based on the constant value.  This can result in large amounts of spurious warnings for registers whose default behavior does not need to change, but are often written or read from.

Adjust the default constant to match the emulator invocation.


swenson (approved):
LGTM

PR #1378, opened 2026-05-07
feat: add ExternalOTP driver for IDevID signature storage

mlvisaya:
Add a partition-based ExternalOTP peripheral driver for storing IDevID certificate signatures (ECC and MLDSA) in external immutable storage.

Architecture:
- HIL trait (runtime/kernel/drivers/external_otp): platform-agnostic ExternalOtp trait with read, write, lock_partition, and is_partition_locked APIs operating on u32 values.
- Capsule (runtime/kernel/capsules): SyscallDriver exposing the HIL to userspace via command-based syscalls.
- Component (runtime/kernel/components): Tock Component wiring.
- Userspace API (runtime/userspace/syscall): synchronous wrapper for userspace applications.

The default emulator implementation uses External SRAM (accessed via DMA) as a reference storage backend. Partition data is laid out sequentially in SRAM with a lock region (one u32 per partition) appended after all partition data.

NOTE: The External SRAM backed implementation is for reference only. Integrators must replace ExtSramBackedExternalOtp with a driver for their platform's actual hardware fuse controller, EPROM, or other immutable storage element.


parvathib (approved):

parvathib:
Suggest a follow-on: back ExternalOTP with flash instead of External SRAM. Flash is already file-backed in the emulator, so the IDevID signature would persist across emulator invocatios, and attestation tests could rely on a pre-provisioned image rather than burning the OTP on every run.

mlvisaya:
thanks! the follow-up PR for flashbacked OTP is noted

Issue #898, opened 2026-02-13
Fuses: Finish runtime command/read/write/lock handlers


swenson:
- Runtime fuse mailbox handlers. No handler exists for `MC_FUSE_READ`, `MC_FUSE_WRITE`, or `MC_FUSE_LOCK_PARTITION` in the runtime command dispatcher (`runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs`).
- Fuse read handler. Needs to: look up `(partition, entry)` → offset/size, read raw fuses via OTP controller, apply FuseLayout to extract the logical value, serialize into `FuseReadResp`.
- Fuse write handler. Needs to: look up `(partition, entry)`, apply FuseLayout encoding (write path), write to OTP controller via DAI.
- Fuse lock handler. Needs to: validate partition number, call OTP controller digest finalization to lock partition.

embediver:
Implementation is done.
Integration tests will be done together with tests for other mcu-mbox tests.
The current implementation for reading and writing don't do any Layout en-/decoding, since they work at a raw address level.
For access to logical fuse values the specific commands that are implemented at this point are better suited, but we could add separate functionality for logical access at a later point if needed.

PR #1380, opened 2026-05-08
[mcu-mbox] fuse read write lock commands

embediver:
Implementation of
- `MC_FUSE_READ`
- `MC_FUSE_WRITE`
- `MC_FUSE_LOCK_PARTITION`

Issue #898.

### DIFF
--- a/romtime/src/mci_mbox_cmd.rs
+++ b/romtime/src/mci_mbox_cmd.rs
@@ -14,25 +14,13 @@ use crate::{HexWord, Mci, Otp};
 use caliptra_api::mailbox::populate_checksum;
 use caliptra_mcu_error::McuError;
 use caliptra_mcu_mbox_common::messages::{
-    verify_checksum, CommandId, FuseLockPartitionReq, FuseReadReq, MailboxReqHeader,
-    MailboxRespHeader, MAX_FUSE_DATA_WORDS,
+    verify_checksum, CommandId, FuseLockPartitionReq, FuseReadReq, FuseWriteReq, MailboxRespHeader,
+    MAX_FUSE_DATA_WORDS,
 };
 use caliptra_mcu_registers_generated::mci;
-use core::cmp::Ordering;
 use core::mem::size_of;
 use tock_registers::interfaces::{ReadWriteable, Readable, Writeable};
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
-
-/// Wire-format header for MC_FUSE_WRITE requests (without the variable-length data).
-#[repr(C)]
-#[derive(FromBytes, KnownLayout, Immutable)]
-struct FuseWriteReqHdr {
-    pub hdr: MailboxReqHeader,
-    pub partition: u32,
-    pub entry: u32,
-    pub start_bit: u32,
-    pub length: u32,
-}
 
 /// Wire-format response for MC_FUSE_READ.
 #[repr(C)]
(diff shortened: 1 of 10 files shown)


runtime/kernel/capsules/src/otp.rs
```diff
@@ -365,6 +370,29 @@ impl Otp {
...
+            if length_bits > 32 {
+                return CommandReturn::failure(ErrorCode::INVAL);
+            }
+            CommandReturn::success()
+        }) {
+            Ok(c) => c,
+            Err(_) => CommandReturn::failure(ErrorCode::FAIL),
+        }
+    }
+    fn lock_otp_partition(&self, partition: u32) -> CommandReturn {
```
zhalvorsen, line 410: Pretty minor, but it might be nice to take the `PartitionId` as the parameter here to make it more explicit

zhalvorsen (approved):

runtime/kernel/capsules/src/otp.rs
embediver: The method essentially just maps the error from the romtime function to the syscall error.
The conversion from u32 to `PartitionId` already happens inside the romtime function.

But matching on the returned error to distinguish between invalid partition id and failure would be good...
There is another case where the handling could be improved, I'll improve it.

embediver: Improved the error handling in that case.
The read syscall has the same problem for verifying that the word address is correct, but there is currently no functionality in the generated fuse definitions for checking that.
Also the `otp.read_word()` method only returns a generic error when reading fails, so there is still a TODO comment in the `write_otp_raw()` method of the syscall implementation to improve the error handling.

romtime/src/mci_mbox_cmd.rs
```diff
@@ -217,75 +218,27 @@ fn handle_fuse_read(buf: &mut [u8], dlen: usize, otp: &Otp) -> Result<usize, Mcu
...
-            return Err(McuError::ROM_OTP_FUSE_INVALID_LENGTH);
-        }
-        Ordering::Equal => {}
-    }
+    crate::println!(
+        "[mci-mbox] IFPW: word_addr={:08X}, data={:08X}, mask={:032b}",
+        req.word_addr,
+        req.data,
+        req.mask,
+    );
```
zhalvorsen, line 238: We probably don't want to print this data in case it is supposed to be kept secret

romtime/src/otp_provision.rs
```diff
@@ -251,149 +251,38 @@ pub fn fuse_read_dai(
...
-                    crate::println!("[otp-provision] DAI write error at word addr {}", word_addr);
-                    return Err(McuError::ROM_OTP_FUSE_DAI_WRITE_ERROR);
-                }
-            }
-        }
-
-        data_bit_consumed += bits_in_word;
+    // Check if the user wants to set a bit from `1` to `0`.
+    // We can do so by inverting `value` to compare every `0` bit in value to the stored one.
+    if (masked_current_word & !masked_value) != 0 {
```
zhalvorsen: Can you add a few unit test vectors for this check?

zhalvorsen (approved):

romtime/src/otp_provision.rs
embediver: Done. Can you take a look and say if it's sufficient?
Also simplified the raw-write syscall implementation to use this function since it was just duplicated functionality at this point.

romtime/src/mci_mbox_cmd.rs
embediver: Done.
Don't know why Github wont show this as outdated...

zhalvorsen (approved):

PR #1382, opened 2026-05-08
docs: add SVN anti-rollback specification

swenson:
Add `docs/src/svn.md`, a specification for Security Version Number anti-rollback across the Caliptra subsystem.

## Model

Each firmware component tracks two SVN values:

- `current_svn` — the SVN of the running image. Used for attestation and enforcement.
- `min_svn` — minimum acceptable SVN, stored in OTP. Images with `current_svn < min_svn` are rejected.

`min_svn` is set independently of `current_svn`, so a release can stage rollback windows before permanently committing a new floor.

## Categories covered

1. **Caliptra Core firmware** — enforced by Caliptra Core ROM (existing fuses).
2. **MCU Runtime firmware** — its `current_svn` is the SoC manifest SVN (the value Caliptra Core authenticates and binds into MCU Runtime's DPE context). Enforced by MCU ROM against a new vendor floor `MCU_RT_MIN_SVN`, independently of Caliptra Core's `CPTRA_CORE_SOC_MANIFEST_SVN`.
3. **MCU Component SVN Manifest** — carries its own `(current_svn, min_svn)` enforced by MCU ROM against a new fuse `MCU_COMPONENT_SVN_MANIFEST_MIN_SVN`.
4. **SoC component images** — Caliptra Core enforces manifest-level SVN; optional per-component enforcement by MCU against `SOC_IMAGE_MIN_SVN[i]`.

## Fuses

Existing Caliptra Core SVN fuses (read by MCU ROM and forwarded to Caliptra; Caliptra Core enforces internally).

New MCU vendor-partition fuses:

- `MCU_RT_MIN_SVN` — compared against the SoC manifest SVN
- `MCU_COMPONENT_SVN_MANIFEST_MIN_SVN` — the Component SVN Manifest's own floor
- `SOC_IMAGE_MIN_SVN[0..M]` — optional per-component floors

All new fuses must use a one-hot encoding (recommended: `OneHotLinearOr` with 3x duplication; plain `OneHot` acceptable when hardware-level redundancy is provided). `CPTRA_CORE_ANTI_ROLLBACK_DISABLE` is reused as the global on/off switch rather than introducing a new MCU-only fuse.

## Image and manifest formats

- `McuImageHeader`: `min_svn: u16 + reserved`. MCU Runtime's `current_svn` is the SoC manifest SVN, not declared independently in the header.
- **MCU Component SVN Manifest** (optional): fixed-size 1024 bytes — 8-byte header (magic, format version, `u8` manifest-self `current_svn`/`min_svn`) plus 127 entries of `(component_id: u32, current_svn: u16, min_svn: u16)`, sized to match Caliptra's `AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT`. All-zero entries are treated as empty slots. The manifest is a header in the MCU runtime image (similar to the firmware-manifest DOT section) identified by its magic; MCU ROM looks for the magic in SRAM after Caliptra Core has loaded the runtime, and skips per-component enforcement when it is absent.

## Enforcement and burning

- ROM-only fuse burning, triggered exclusively by signed image headers and authenticated SVN manifests; runtime never burns SVN fuses.
- Cold boot and hitless update share identical MCU Runtime SVN logic.
- PLDM Verify Component performs SVN checks before bundle activation, so a downgrade attempt never reaches the hitless reset.
- `SocComponentSvn` trait lets platforms extract the running SVN from opaque component bytes, cross-checked against the manifest entry.
- `component_id` (reused from Caliptra's `AuthManifestImageMetadata`) maps to fuse slots via a platform-defined `SVN_FUSE_MAP`; the mapping is many-to-one so components updated together can share a slot.

## Platform configuration

- `vendor_fuses.hjson` example for the new fuses
- `SVN_FUSE_MAP` example
- `ImageVerifier` reference implementation
- `SocComponentSvn` trait definition

## Security considerations

The spec closes with notes on `min_svn` / `current_svn` separation, single-SVN-per-release for MCU Runtime, manifest-format-SVN independence from component SVNs, ROM-only fuse burning, one-way commitment, anti-rollback disable polarity, SVN exhaustion, and ownership transfer.

### DIFF
--- a/docs/src/svn.md
+++ b/docs/src/svn.md
@@ -4,8 +4,8 @@
 
 Each firmware component tracks two SVN values:
 
-- **`current_svn`** — the security version of the running image, declared by
-  the image. Used for enforcement and attestation.
+- **`current_svn`** — the security version of the running image. Used for
+  enforcement and attestation.
 - **`min_svn`** — the minimum acceptable security version, stored in OTP fuses.
   Any image with `current_svn < min_svn` is rejected.
 
@@ -14,12 +14,37 @@ Each firmware component tracks two SVN values:
 the device runs version 10. The deployer chooses when to permanently commit a
 new minimum.
 
-This document covers three categories of components:
+This document covers four categories of components:
 
 1. **Caliptra Core firmware** — enforced by Caliptra Core ROM.
-2. **MCU Runtime firmware** — enforced by MCU ROM.
-3. **SoC component images** — manifest-level SVN enforced by Caliptra Core;
-   optional per-component enforcement by MCU.
+2. **MCU Runtime firmware** — its `current_svn` is the **SoC manifest SVN**
+   (the same value Caliptra Core authenticates and binds into the MCU
+   Runtime's DPE context). Enforced by both Caliptra Core (against
+   `CPTRA_CORE_SOC_MANIFEST_SVN`) and, with a separate vendor-controlled
+   floor, by MCU ROM (against `MCU_RT_MIN_SVN`).
+3. **MCU Component SVN Manifest** — its own SVN, enforced by MCU ROM against
+   `MCU_COMPONENT_SVN_MANIFEST_MIN_SVN`.
+4. **SoC component images** — manifest-level SVN enforced by Caliptra Core;
+   optional per-component enforcement by MCU against `SOC_IMAGE_MIN_SVN[i]`.
+
+### MCU Runtime SVN reuses the SoC manifest SVN
+
+The MCU Runtime image is delivered as part of the SoC manifest and is
+authenticated by Caliptra Core, which then creates the MCU Runtime's DPE
+context using the SoC manifest's SVN. To keep DPE attestation, Caliptra
+Core's anti-rollback, and MCU's anti-rollback consistent, MCU Runtime does
+**not** declare its own `current_svn` independently of the SoC manifest:
+there is exactly one SVN per release for MCU Runtime, and it lives in the
+SoC manifest. MCU ROM obtains the authenticated SoC manifest SVN from
+Caliptra Core (mechanism is platform-defined; see
+[Cold Boot and Hitless Update — MCU Runtime SVN](#cold-boot-and-hitless-update--mcu-runtime-svn)).
+
+`MCU_RT_MIN_SVN` is a separate, MCU-vendor-controlled floor the MCU ROM
+enforces against this same SoC manifest SVN. It is independent of
+`CPTRA_CORE_SOC_MANIFEST_SVN` so that an MCU vendor (or device owner)
+can advance the floor without coordinating a Caliptra Core fuse change,
+and so that ownership-transfer operations that retarget Caliptra fuses do
+not silently disturb the MCU-side floor.
 
 ## Threat Model
 
@@ -63,11 +88,15 @@ Added in a vendor partition (e.g., `VENDOR_NON_SECRET_PROD_PARTITION`):
 
 | Fuse | Size | Recommended Encoding | Purpose |
 |---|---|---|---|
-| `MCU_RT_MIN_SVN` | 16 B | `OneHotLinearOr{bits:N, dupe:3}` | MCU Runtime min SVN |
+| `MCU_RT_MIN_SVN` | 16 B | `OneHotLinearOr{bits:N, dupe:3}` | MCU Runtime min SVN (compared against the SoC manifest SVN) |
+| `MCU_COMPONENT_SVN_MANIFEST_MIN_SVN` | 4 B | `OneHotLinearOr{bits:N, dupe:3}` | Min SVN for the MCU Component SVN Manifest itself |
 | `SOC_IMAGE_MIN_SVN[0..M]` | 4 B each | `OneHotLinearOr{bits:N, dupe:3}` | Per-slot SoC image min SVN (optional) |
 
 The number of `SOC_IMAGE_MIN_SVN` slots (`M`) and the number of one-hot bits
-per slot are integrator-defined.
+per slot are integrator-defined. `MCU_COMPONENT_SVN_MANIFEST_MIN_SVN` exists
+even when the integrator does not provision any `SOC_IMAGE_MIN_SVN` slots,
+because the manifest's own anti-rollback applies regardless of how many
+component slots are mapped.
 
 #### Encoding
 
@@ -87,21 +116,26 @@ See [Fuse Layout Options](fuses.md#fuse-layout-options) for encoding details.
 
 ## MCU Image Header
 
-The MCU Runtime binary starts with an `McuImageHeader`:
+The MCU Runtime binary may be preceded by an `McuImageHeader`. It carries
+only the MCU-side `min_svn` burn request; the runtime's `current_svn` is
+the SoC manifest SVN (see
+[Overview](#mcu-runtime-svn-reuses-the-soc-manifest-svn)).
 
 | Field | Size | Description |
 |---|---|---|
-| `current_svn` | 2 B | Security version of this image |
 | `min_svn` | 2 B | Requested new floor to burn into `MCU_RT_MIN_SVN` (0 = no update) |
-| `reserved` | 4 B | Reserved |
+| `reserved` | 6 B | Reserved (must be zero) |
 
 Constraints (validated by ROM; image is rejected on violation):
 
-- `min_svn ≤ current_svn`
-- Both values must fit within the maximum representable in `MCU_RT_MIN_SVN`'s
-  one-hot encoding.
+- `min_svn ≤ soc_manifest_svn` (the authenticated SoC manifest SVN, obtained
+  from Caliptra Core).
+- `min_svn` must fit within the `MCU_RT_MIN_SVN` one-hot range.
+- All `reserved` bytes must be zero.
 
-The firmware bundler sets these via `--svn <current>` and `--min-svn <min>`.
+The firmware bundler sets the requested floor via `--min-svn <min>`. There
+is no `--svn` flag at MCU Runtime scope; the runtime's `current_svn` is the
+SoC manifest's SVN, set by the SoC manifest signing flow.
 
 ## MCU Component SVN Manifest (Optional)
 
@@ -117,22 +151,28 @@ Caliptra Core) provides rollback protection for SoC images, and the
 
 ### Format
 
-The manifest is fixed-size, sized to match Caliptra's
-`AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT` (127 entries) — the maximum number of
-components a single SoC manifest can describe.
+The manifest is a fixed-size structure. The per-entry section is sized to
+match Caliptra's `AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT` (127 entries) —
+the maximum number of components a single SoC manifest can describe.
 
 | Field | Size | Description |
 |---|---|---|
 | Magic | 4 B | `0x4D435356` (`"MCSV"`) |
-| Version | 4 B | Manifest format version |
+| Format Version | 2 B | Manifest format version |
+| `current_svn` | 1 B | SVN of this manifest (rolled forward when manifest entries change) |
+| `min_svn` | 1 B | Requested new floor to burn into `MCU_COMPONENT_SVN_MANIFEST_MIN_SVN` (0 = no update) |
 | Entries | 8 B × 127 | `(component_id: u32, current_svn: u16, min_svn: u16)` |
 
-Total: 1024 bytes. SVNs are `u16`; even a single 65535-bit one-hot fuse would
-be impractically large.
+Total: 1024 bytes (8-byte header + 1016 bytes of entries).
 
 An entry where all fields are zero (`component_id == current_svn == min_svn == 0`)
 is treated as an empty slot and ignored, allowing manifests to declare fewer
 than 127 entries by zero-padding.
+
+Header constraints (validated; manifest is rejected on violation):
+
+- `min_svn ≤ current_svn`
+- Both must fit within `MCU_COMPONENT_SVN_MANIFEST_MIN_SVN`'s one-hot range.
 
 Per-entry constraints (validated; entry is rejected on violation):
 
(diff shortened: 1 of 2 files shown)


docs/src/svn.md
```diff
@@ -0,0 +1,378 @@
...
+
+`min_svn` is set independently of `current_svn`. A release may carry
+`current_svn = 10` but `min_svn = 7`, allowing rollback to versions 7–9 while
+the device runs version 10. The deployer chooses when to permanently commit a
+new minimum.
+
+This document covers three categories of components:
+
+1. **Caliptra Core firmware** — enforced by Caliptra Core ROM.
+2. **MCU Runtime firmware** — enforced by MCU ROM.
```
parvathib: We should probably discuss this. The proposal is that Caliptra RT enforces the anti-rollback protection during loading and validating the MCU RT. Since MCU ROM is not considered part of the TCB. The SVN also is needed by Caliptra RT to create MCU RT DPE context

mhatrevi (approved):

mlvisaya (approved):
LGTM

docs/src/svn.md
```diff
@@ -0,0 +1,378 @@
...
+defaults to 0 (enforcement on) and should be set only on development or
+manufacturing devices.
+
+### New MCU SVN Fuses
+
+Added in a vendor partition (e.g., `VENDOR_NON_SECRET_PROD_PARTITION`):
+
+| Fuse | Size | Recommended Encoding | Purpose |
+|---|---|---|---|
+| `MCU_RT_MIN_SVN` | 16 B | `OneHotLinearOr{bits:N, dupe:3}` | MCU Runtime min SVN |
```
parvathib: Are any of these fuses accessible by Caliptra Core?

docs/src/svn.md
```diff
@@ -0,0 +1,378 @@
...
+
+### Cold Boot and Hitless Update — MCU Runtime SVN
+
+After Caliptra loads MCU Runtime into MCU SRAM, MCU ROM enforces and
+potentially burns the MCU SVN before jumping to firmware. The same logic
+applies on cold boot and hitless update reset:
+
+1. Read `McuImageHeader` from SRAM.
+2. Read `MCU_RT_MIN_SVN` and `CPTRA_CORE_ANTI_ROLLBACK_DISABLE` from OTP.
+3. If anti-rollback is not disabled and `header.current_svn < fuse_min_svn`:
```
parvathib: How would Caliptra Core update the SVN in the DPE context for MCU?

docs/src/svn.md
```diff
@@ -0,0 +1,378 @@
...
+
+Constraints (validated by ROM; image is rejected on violation):
+
+- `min_svn ≤ current_svn`
+- Both values must fit within the maximum representable in `MCU_RT_MIN_SVN`'s
+  one-hot encoding.
+
+The firmware bundler sets these via `--svn <current>` and `--min-svn <min>`.
+
+## MCU Component SVN Manifest (Optional)
```
parvathib, line 140: how is the anti-rollback for MCU Component SVN manifest enforced?

docs/src/svn.md
```diff
@@ -0,0 +1,378 @@
...
+If the manifest is absent, per-component enforcement is skipped; only the
+Caliptra-enforced SoC manifest SVN applies.
+
+## SVN Fuse Burning
+
+`min_svn` fuses are **only burned by MCU ROM**. Runtime never burns SVN fuses,
+ensuring fuse programming runs in the most trusted execution context before
+mutable firmware has control.
+
+Burns are triggered exclusively by authenticated firmware images: the
```
parvathib, line 416: Does this mean burning fuse will be integrated into the recovery boot or hitless update flow, or handled through a separate external command? Also, could you add details on the per-component flow?

docs/src/svn.md
```diff
@@ -0,0 +1,378 @@
...
+[Component SVN Fuse Map](#component-svn-fuse-map)).
+
+If this manifest is omitted, only the SoC manifest-level SVN (enforced by
+Caliptra Core) provides rollback protection for SoC images, and the
+`SOC_IMAGE_MIN_SVN` fuses are unused.
+
+### Format
+
+The manifest is fixed-size, sized to match Caliptra's
+`AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT` (127 entries) — the maximum number of
```
parvathib: So this count excludes the `MCU RT` and the `MCU component SVN manifest` SoC components from AUTH_MANIFEST_IMAGE_METADATA_MAX_COUNT, right?

docs/src/svn.md
swenson: We can do that as well, but currently Caliptra RT doesn't do that, and doesn't have any way to know MCU's SVN, right?

docs/src/svn.md
swenson: Not really -- theoretically Caliptra core could read them, but it would have no way to determine which fuse to read, and the fuse map is integration-specific. Normally MCU ROM reads the fuses and sends the fuses to the Caliptra fuse registers, but there is no fuse register for this.

docs/src/svn.md
swenson: I don't think Caliptra core has any knowledge of MCU's SVN, so it's not relevant. If in the future we add the ability for Calitpra core to know MCU's SVN, then we would want to provide a way to update it, I assume.

docs/src/svn.md
swenson: Good question. We would have to have a way to know the SVN for a particular component. I'm not sure the right approach there. Any ideas?

docs/src/svn.md
swenson: Yeah, I think if we get an updated FW image, we would want to handle this as part of update reset.

I'll add some details on per-component flows.

docs/src/svn.md
swenson: Yeah, I think so.

docs/src/svn.md
swenson: I added some details about this about SoC components.

docs/src/svn.md
swenson: Added.

docs/src/svn.md
parvathib: One option is to reuse SoC manifest SVN field (and corresponding fuses) for MCU RT as well, treating MCU as a special case.

docs/src/svn.md
parvathib: I’m specifically interested in the MCU RT SVN. If this can be made accessible to the Core, it could enforce anti-rollback in addition to verifying the digest against the SoC manifest for MCU RT. And would use the SVN to create/update DPE context.

docs/src/svn.md
parvathib: To clarify my question: during recovery boot, when Caliptra RT creates a DPE context for the MCU RT component, it needs both `<digest, svn>` as inputs. Similarly, during hitless updates, the SVN would likely need to be updated along with the digest, journey digest in the MCU DPE context.

Since SVN is a key information, it should be included in attestation claims for MCU RT component. This information comes from the DPE context and cannot come from the MCU ROM, as the MCU ROM is not part of the TCB.

parvathib:
Suggestions for how this works with the Caliptra Core side:

**Summary:** Anti-rollback enforcement for SoC components (including the SVN manifest) stays with MCU, while Caliptra Core handles manifest-level protection and attestation.

MCU (ROM + RT):
- Enforces anti-rollback (fuse burning MCU ROM + `current_svn >= fuse_min_svn` checks in MCU RT) for all SoC components, including the SVN manifest itself
- Passes `current_svn` per SoC component to Caliptra Core via `AuthorizeAndStashReq.svn`

Caliptra Core:
- Enforces manifest-level SVN via `CPTRA_CORE_SOC_MANIFEST_SVN` fuse (existing)
- Creates MCU RT DPE context node with SVN for attestation certificates (source for MCU RT SVN — see options below)
- SoC component DPE context creation is optional; when created, uses `current_svn` from `AuthorizeAndStashReq.svn` passed by MCU RT

**1. MCU RT SVN in DPE context:**
Today `AuthorizeAndStashReq.svn` defaults to 0 for MCU RT (fw_id=2), so its DPE context node has no SVN and attestation certificates omit the security version for MCU RT. Caliptra RT needs a source for this SVN.

**Open question:** Should MCU RT have its own SVN independent of the manifest SVN, or is tying it to the manifest SVN acceptable?

- **Option 1 (Preamble SVN):** Reuse the auth manifest preamble SVN. Simple, already validated against fuses. But couples MCU RT SVN to the entire manifest's SVN.
- **Option 2 (Parse McuImageHeader):** Caliptra RT reads `McuImageHeader.current_svn` from the authenticated MCU image in SRAM via DMA. MCU RT-specific SVN, single source of truth in the image. Adds DMA read code in Caliptra RT.
- **Option 3 (MCU metadata entry flags):** Add a per-component SVN field to `AuthManifestImageMetadata` (e.g., flags[31:16]). Caliptra Core reads MCU RT's SVN directly from the signed manifest for the DPE context — no MCU image parsing needed. This value must match `McuImageHeader.current_svn` that MCU ROM uses for fuse enforcement. Both are set by the same vendor/operator build pipeline. The trade-off is SVN in two places that must stay in sync.

**2. SVN Manifest anti-rollback:**
The SVN manifest goes through `AUTHORIZE_AND_STASH` like any other SoC component. Adding `current_svn` and `min_svn` to its header allows it to follow the same anti-rollback pattern as all other components:

| Field | Size | Description |
|---|---|---|
| Magic | 4 B | `0x4D435356` ("MCSV") |
| Version | 4 B | Manifest format version |
| current_svn | 2 B | Security version of this manifest |
| min_svn | 2 B | Requested new floor to burn into fuse |
| Entries | 8 B × 127 | `(component_id: u32, current_svn: u16, min_svn: u16)` |

MCU RT checks `current_svn >= fuse_min_svn` using the SVN manifest's own `SOC_IMAGE_MIN_SVN[i]` fuse slot. MCU ROM burns the fuse from `min_svn`. No special cases — uniform treatment across all SoC components. The auth manifest preamble SVN still provides umbrella protection on top.

**3. Per-SoC-component SVN in AuthorizeAndStashReq:**
MCU RT passes `current_svn` from the authenticated SVN manifest in `AuthorizeAndStashReq.svn` for each SoC component. Caliptra Core uses it for the DPE context node (if created) or stashes it along with Digest to PCR31. No changes to `AuthManifestImageMetadata` needed — the SVN manifest is the single source of per-component SVN.

**Caveat:** Caliptra Core trusts the SoC component SVN value MCU RT passes. An alternative hardening step could be to embed `current_svn` in auth manifest metadata (e.g., flags[31:16]) for cross-verification.

**SVN responsibility matrix:**

| Component | SVN Check | Who | What | Fuse |
|---|---|---|---|---|
| Auth Manifest | Anti-rollback | Caliptra RT | `preamble.svn >= soc_manifest_fuse_svn` | `CPTRA_CORE_SOC_MANIFEST_SVN` (existing) |
| MCU RT | DPE context SVN | Caliptra RT | Source TBD — see options 1-3 above | Depends on option chosen |
| MCU RT | min_svn enforcement | MCU ROM | `header.current_svn >= MCU_RT_MIN_SVN` | `MCU_RT_MIN_SVN` (new OTP) |
| MCU RT | Fuse burning | MCU ROM | Burns fuse if `header.min_svn > fuse_min_svn` | `MCU_RT_MIN_SVN` (new OTP) |
| SVN Manifest | Anti-rollback | MCU RT | `current_svn >= fuse_min_svn` | `SOC_IMAGE_MIN_SVN[i]` (new OTP) |
| SVN Manifest | Fuse burning | MCU ROM | Burns fuse from SVN manifest header `min_svn` | `SOC_IMAGE_MIN_SVN[i]` (new OTP) |
| SoC Components | min_svn enforcement | MCU RT | `current_svn >= fuse_min_svn` per component | `SOC_IMAGE_MIN_SVN[i]` (new OTP) |
| SoC Components | Fuse burning | MCU ROM | Burns fuse from SVN manifest `min_svn` entries | `SOC_IMAGE_MIN_SVN[i]` (new OTP) |

---

docs/src/svn.md
```diff
@@ -0,0 +1,552 @@
...
+
+This document covers four categories of components:
+
+1. **Caliptra Core firmware** — enforced by Caliptra Core ROM.
+2. **MCU Runtime firmware** — its `current_svn` is the **SoC manifest SVN**
+   (the same value Caliptra Core authenticates and binds into the MCU
+   Runtime's DPE context). Enforced by both Caliptra Core (against
+   `CPTRA_CORE_SOC_MANIFEST_SVN`) and, with a separate vendor-controlled
+   floor, by MCU ROM (against `MCU_RT_MIN_SVN`).
+3. **MCU Component SVN Manifest** — its own SVN, enforced by MCU ROM against
```
parvathib, line 25: Anti-rollback for SVN manifest is enforced by MCU RT (just like other SoC components) I guess. MCU ROM would only writes the fuses associated with it.

docs/src/svn.md
```diff
@@ -0,0 +1,552 @@
...
+The map is **many-to-one**: multiple `component_id` values may share the same
+`SOC_IMAGE_MIN_SVN[i]` slot, in which case those components share a `min_svn`
+floor. This is appropriate for components that always update together as a
+unit and conserves fuse space. Sharing components must agree on `min_svn` per
+release; the build system should validate this.
+
+### Loading and Authentication
+
+The manifest is delivered as a SoC image (with its own `component_id` and
+digest in the SoC manifest), via the recovery interface or a PLDM firmware
```
parvathib: I may be misunderstanding this, but I was assuming the SVN manifest would be delivered similar to other SoC components.
When we say “recovery” here, do we mean it would be delivered alongside the MCU RT and SoC manifest, or are we referring to the PLDM image loading sequence instead?

parvathib (approved):
LGTM.  Just need a couple of minor clarifications. Thanks for addressing the comments aptly.

docs/src/svn.md
swenson: Not sure about that -- if MCU ROM read the manifest, it can refuse to boot if the SVN manifest has too low of a version number, right? And probably should? It can be enforced by MCU RT during a firmware update though.

docs/src/svn.md
swenson: Oh, I think it is part of the MCU RT. I missed this when reviewing. Let me fix this.

swenson:
Thanks!

PR #1383, opened 2026-05-08
2.1 Cherry pick swenson phase 1

swenson:
fuses: Remove unused generated Fuses struct (#974)

And use new partition info in OTP digest calculation.

(cherry picked from commit 1e800d7e3c0c7896b41d8df0354e2c39323cbb30)

rom: Add OTP SW digest to driver (#973)

We refactor the OTP digest code (PRESENT cipher) so that it is no_std
and able to be used by the emulator or by the firmware. We also added an
iterator-based digest method so that the entire OTP partition doesn't
need to be copied to the stack, but can be read a dword at a time.

(cherry picked from commit 776c8717f001cc3157f07fb5021cb044cb484e22)

rom: implement DOT challenge/response recovery via MCI mbox0 (#918)

Add a challenge/response recovery mechanism to the MCU ROM for Device Ownership Transfer (DOT). When the DOT blob is corrupted or missing and the fuse array is in ODD (locked) state, the ROM can recover by communicating with an external agent (e.g., BMC) over MCI mailbox 0.

Recovery protocol:

Two-transaction flow: DOT_RECOVERY_REQUEST (agent sends ECC P-384 + MLDSA-87 public keys and new DOT blob data) followed by DOT_CHALLENGE_RESPONSE (agent signs a ROM-generated random challenge).
ROM verifies public keys against the vendor recovery PK hash burned in OTP fuses, then verifies both ECDSA P-384 and MLDSA-87 signatures using Caliptra mailbox commands.
All public keys, signatures, and hashes use standard byte order on the wire (SEC1 for ECC, FIPS 204 for MLDSA). The ROM performs any byte-swapping needed internally for Caliptra.
Both commands include a Caliptra-standard checksum as the first u32 field.
Key implementation details:

Add Mbox0RecoveryTransport in rom/src/dot_recovery.rs implementing the RecoveryTransport trait. Uses zero-copy references into MCI mbox0 SRAM for large MLDSA keys/signatures to stay within 16KB ROM RAM.
Add exec_mailbox_req_u32_parts() to romtime for streaming multi-part &[u32] payloads to the Caliptra mailbox without stack copies.
Add cm_ecdsa384_verify, cm_mldsa87_verify, and cm_sha384 crypto helpers to device_ownership_transfer.rs.
Refactor DotBlob to separate DotBlobFields for reuse in recovery flow.
Add attempt_dot_recovery() to cold_boot.rs: tries challenge/response transport first, then falls back to backup blob handler. Only triggered when fuse array is in ODD state.
Add RecoveryTransport trait and supporting types (RecoveryRequest, ChallengeResponse, EccP384PublicKey, DotBlobData).
Add dot_recovery_transport and dot_recovery_handler fields to RomParameters.
Wire up Mbox0RecoveryTransport in emulator ROM entry point.
Testing:

Add integration tests with random ECC/MLDSA key generation per run.
Tests exercise: successful recovery, wrong PK hash rejection, and fuse state guards (EVEN state skips DOT, ODD state triggers recovery).
Host-side test code performs signing via MCI mbox0 protocol.
Documentation:

Update docs/src/dot.md with DOT_RECOVERY_REQUEST and DOT_CHALLENGE_RESPONSE command format tables following the standard mailbox command documentation format.

(cherry picked from commit a6784b633b2400d91c271929eb2e62584b04e685)


mhatrevi (approved):

swenson:
Thanks!

PR #1388, opened 2026-05-09
Add the support for getlog/clearlog commands

helloxiling:


mlvisaya (approved):
LGTM. nit: more_data and data_size are encoded as u32 in MCTP VDM but encoded as u8 in SPDM VDM. If this is intentional, then it should be good.

helloxiling:
> LGTM. nit: more_data and data_size are encoded as u32 in MCTP VDM but encoded as u8 in SPDM VDM. If this is intentional, then it should be good.

Yes, it is intentional — each transport uses its protocol's natural completion-code width: MCTP VDM = u32.  `more_data` and `data_size` follows the same width for alignment. Thanks for the careful look!

PR #1389, opened 2026-05-09
Add PLDM firmware update restart capability

mlvisaya:
- Allow PLDM tasks to be restarted. The use-case is to do streaming boot and then firmware update. In this case PLDM task needs to be restarted to give it a new set of FdOps.
- Fix MCI stale firmware update reason detection (reason_at_cycle_complete)

### DIFF
--- a/runtime/userspace/api/pldm-lib/src/daemon.rs
+++ b/runtime/userspace/api/pldm-lib/src/daemon.rs
@@ -90,6 +90,9 @@ impl<'a> PldmService<'a> {
             return Err(PldmServiceError::StartError);
         }
 
+        // Reset the stopped signal to prevent stale Signaled state from a
+        // previous cycle causing wait_until_stopped() to return immediately.
+        self.stopped_signal.reset();
         self.running.store(true, Ordering::SeqCst);
 
         let cmd_interface: &'static CmdInterface<'static> =
--- a/runtime/userspace/api/pldm-lib/src/daemon.rs
+++ b/runtime/userspace/api/pldm-lib/src/daemon.rs
@@ -52,6 +52,20 @@ pub struct PldmService<'a> {
     cmd_interface: CmdInterface<'a>,
     running: &'static AtomicBool,
     initiator_signal: &'static Signal<CriticalSectionRawMutex, ()>,
+    stopped_signal: &'static Signal<CriticalSectionRawMutex, ()>,
+}
+
+static RUNNING: AtomicBool = AtomicBool::new(false);
+static INITIATOR_SIGNAL: Signal<CriticalSectionRawMutex, ()> = Signal::new();
+static STOPPED_SIGNAL: Signal<CriticalSectionRawMutex, ()> = Signal::new();
+
+/// Wait until the PLDM service has fully stopped (responder task exited).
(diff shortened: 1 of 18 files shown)


runtime/userspace/api/pldm-lib/src/daemon.rs
```diff
@@ -52,6 +52,20 @@ pub struct PldmService<'a> {
...
+static INITIATOR_SIGNAL: Signal<CriticalSectionRawMutex, ()> = Signal::new();
+static STOPPED_SIGNAL: Signal<CriticalSectionRawMutex, ()> = Signal::new();
+
+/// Wait until the PLDM service has fully stopped (responder task exited).
+/// Can be called without a PldmService instance.
+pub async fn wait_until_stopped() {
+    if !RUNNING.load(Ordering::SeqCst) {
+        return;
+    }
+    STOPPED_SIGNAL.wait().await;
```
helloxiling, line 68: We may have potential issue with STOPPED_SIGNAL leak across cycles and can return prematurely. 
embassy_sync::Signal::signal() leaves the slot in Signaled; only wait() (or the new reset()) clears it. Across the typical shutdown order in pldm_responder/pldm_responder_task:
  1. running.store(false) (call it A)
  2. stopped_signal.signal(()) (call it B)

If a caller invokes wait_until_stopped() after B, the early-return path is taken and the Signaled value is never consumed. On the next cycle, when a caller actually needs to block (RUNNING=true at check
 time), STOPPED_SIGNAL.wait().await returns immediately from the leftover and the caller proceeds while the new responder is still running.  The fix could be reset this signal at start() .

emulator/periph/src/mci.rs
```diff
@@ -613,9 +618,17 @@ impl MciPeripheral for Mci {
...
             };
 
-            if has_fw_update_bits && !self.reset_cycle_complete {
-                // Caliptra wrote FW update bits for this reset cycle
-                // (first reset after cold boot). Preserve them.
+            // Detect whether FW update bits are fresh: either this is the
+            // first reset cycle (!reset_cycle_complete), or Caliptra wrote
+            // new FW bits via DMA since the last cycle completed (the value
+            // changed from what we saved at cycle completion).
+            let fw_bits_are_fresh = has_fw_update_bits
```
helloxiling, line 625: If Caliptra writes the same FW-update reason value across two consecutive cycles, the comparison `reason == self.reason_at_cycle_complete` is true even though the bits were freshly written, and the path falls into the "stale" branch and clears them. Can you please review if this is a concern?

runtime/userspace/api/pldm-lib/src/daemon.rs
mlvisaya: Good catch! Added `self.stopped_signal.reset()` at the beginning of `start()` before setting `RUNNING` to `true`. This clears any stale `Signaled` state left from a previous cycle, preventing `wait_until_stopped()` from returning immediately on the next cycle.

emulator/periph/src/mci.rs
mlvisaya: this MCI change has been abandoned for this PR.

helloxiling (approved):
Looks good to me.

PR #1390, opened 2026-05-09
Add streaming boot flash write-back support

mlvisaya:
- Add test-streaming-boot-flash-write-back feature and test
- Add SpiFlashStagingMemory for flash-backed staging area
- Add skip_activation and verify_running_image_match to firmware update
- Add caliptra-image-types dependency for ImageManifest verification
- Wire up feature gates across emulator, builder, and platform crates


platforms/emulator/runtime/userspace/apps/user/src/firmware_update/mod.rs
```diff
@@ -71,6 +74,29 @@ pub async fn firmware_update<D: DMAMapping>(dma_mapping: &D) -> Result<(), Error
...
 
+    #[cfg(feature = "test-streaming-boot-flash-write-back")]
+    {
+        let fw_params = PldmFirmwareDeviceParams {
+            descriptors: &config::fw_update_consts::DESCRIPTOR.get()[..],
+            fw_params: config::fw_update_consts::FIRMWARE_PARAMS.get(),
+        };
+        let staging_memory = flash_staging::SpiFlashStagingMemory::new();
+        let staging_memory: &'static flash_staging::SpiFlashStagingMemory =
+            unsafe { core::mem::transmute(&staging_memory) };
```
helloxiling: `SpiFlashStagingMemory` is stack-allocated but transmuted to &'static and stored in DOWNLOAD_CTX for spawned PLDM tasks. With skip_activation, start() can return while that static context still points to the stack value, leaving a dangling reference. 
Can we consider making the staging memory truly static, or clear the PLDM context and wait for the service to stop before returning?

mlvisaya: Good catch — fixed in 9d490b12. The staging memory is now a `LazyLock` static, so the reference is genuinely `'static` and the unsafe transmute is eliminated.

helloxiling (approved):
Looks good to me.

PR #1398, opened 2026-05-11
Add state machine and timeout to support chunked mailbox requests

mlvisaya:
Enforce ordering for the chunked mailbox flow (initiate -> send chunks -> execute) via a MailboxState enum. Requests issued out of order are rejected with INVAL. A configurable timeout resets the mailbox to idle if the user stalls between steps, preventing one app from blocking others indefinitely.

Also adds integration tests for the chunked path and timeout behavior.

### DIFF
--- a/runtime/kernel/capsules/src/mailbox.rs
+++ b/runtime/kernel/capsules/src/mailbox.rs
@@ -92,6 +92,17 @@ mod rw_allow {
     pub const COUNT: u8 = 1;
 }
 
+/// State machine for the chunked mailbox command flow (commands 2/3/4).
+#[derive(Clone, Copy, PartialEq)]
+enum MailboxState {
+    /// No mailbox operation in progress.
+    Idle,
+    /// After initiate_request (command 2), waiting for data chunks and execute.
+    Initiated,
+    /// After execute (command 4) or enqueue_command (command 1), polling for response.
+    Executing,
+}
+
 #[derive(Default)]
 pub struct App {}
 
@@ -108,6 +119,11 @@ pub struct Mailbox<'a, A: Alarm<'a>> {
     >,
     // Which app is currently using the storage.
     current_app: OptionalCell<ProcessId>,
+    // Current state of the mailbox state machine.
+    state: Cell<MailboxState>,
+    // Timeout ticks for the initiated state before auto-resetting to idle.
+    // If None, no timeout is enforced.
+    timeout_ticks: Option<u32>,
     resp_min_size: Cell<usize>,
     resp_size: Cell<usize>,
 }
@@ -122,12 +138,15 @@ impl<'a, A: Alarm<'a>> Mailbox<'a, A> {
             AllowRwCount<{ rw_allow::COUNT }>,
         >,
         driver: &'static mut CaliptraSoC,
+        timeout_ticks: Option<u32>,
     ) -> Mailbox<'a, A> {
         Mailbox {
             alarm,
             driver: TakeCell::new(driver),
             apps: grant,
             current_app: OptionalCell::empty(),
+            state: Cell::new(MailboxState::Idle),
+            timeout_ticks,
             resp_min_size: Cell::new(0),
             resp_size: Cell::new(0),
         }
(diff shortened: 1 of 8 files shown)


runtime/kernel/capsules/src/mailbox.rs
```diff
@@ -152,23 +174,32 @@ impl<'a, A: Alarm<'a>> Mailbox<'a, A> {
...
+                },
+            )
             .ok_or(ErrorCode::RESERVE)?
     }
 
     fn send_next_chunk(&self, processid: ProcessId) -> Result<(), ErrorCode> {
-        // Check if we're already executing a mailbox command.
-        if self.current_app.is_none() {
-            return Err(ErrorCode::CANCEL);
+        // Only allowed after initiate_request (command 2).
```
mhatrevi, line 255: Verify that the caller is current_app

runtime/kernel/capsules/src/mailbox.rs
```diff
@@ -220,17 +253,22 @@ impl<'a, A: Alarm<'a>> Mailbox<'a, A> {
     }
 
     fn execute(&self) -> Result<(), ErrorCode> {
```
mhatrevi: This needs to accept process id and verify the ownership.

runtime/kernel/capsules/src/mailbox.rs
```diff
@@ -310,27 +348,50 @@ impl<'a, A: Alarm<'a>> Mailbox<'a, A> {
...
-        let reschedule = self
-            .driver
-            .map(|driver| {
-                if driver.is_mailbox_busy() {
-                    true
+        match self.state.get() {
+            MailboxState::Initiated => {
+                // Timeout: user didn't complete the chunked send in time.
+                debug!("Mailbox initiate timeout: resetting to idle");
+                let _ = self.alarm.disarm();
```
mhatrevi, line 446: The mailbox lock doesn't seem to be getting cleared here. Need to reset the execute bit on soc_mbox.

mhatrevi: Similar issue in romtime/src/soc_manager.rs.

runtime/userspace/syscall/src/mailbox.rs
```diff
@@ -123,49 +123,57 @@ impl<S: Syscalls> Mailbox<S> {
         }
     }
 
-    pub async fn execute_with_payload_stream(
+    /// Initiates a chunked mailbox request.
+    ///
+    /// Call this first, then send data with [`send_chunk`](Self::send_chunk),
+    /// and finally call [`execute_chunked_request`](Self::execute_chunked_request).
+    pub async fn start_chunked_request(
```
mhatrevi, line 137: This method is public, but it does not hold the global mailbox mutex. How is concurrency handled?

mhatrevi (changes requested):

runtime/kernel/capsules/src/mailbox.rs
mlvisaya: Done. Added a check that `self.current_app.get() == Some(processid)`, returning `ErrorCode::INVAL` if the caller isn't the process that initiated the request.

runtime/kernel/capsules/src/mailbox.rs
mlvisaya: Done. Changed signature to `fn execute(&self, processid: ProcessId)` with the same ownership check. Updated command 4 dispatch to pass `processid`.

runtime/kernel/capsules/src/mailbox.rs
mlvisaya: Done. Added `driver.abort_request()` in the `Initiated` timeout arm of `alarm()` to release the HW mailbox lock before resetting state. Also added `pub fn abort_request(&mut self)` to `CaliptraSoC` in `romtime/src/soc_manager.rs` which clears the execute bit via `self.soc_mbox().execute().write(|w| w.execute(false))`.

runtime/userspace/syscall/src/mailbox.rs
mlvisaya: The kernel enforces process ID verification and state machine ordering, so different processes cannot interleave chunked flows. Within a single process with multiple async tasks, callers should use `execute_with_payload_stream` (which holds the global mailbox mutex) or acquire `MAILBOX_MUTEX` externally. Added documentation on `start_chunked_request` explaining this concurrency model.

mhatrevi (approved):

PR #1402, opened 2026-05-11
rom: replace `{:?}` error prints with HexWord codes to shrink ROM

ternst-nv:
`Result::unwrap` and `"{:?}"` formatting pull in each error enum's `Debug` impl plus its variant-name strings and the shared `core::fmt` machinery. Converting the 13 error-print sites in `cold_boot.rs` and `device_ownership_transfer.rs` to `HexWord(err.into())` / `HexWord(crate::err_code(&err))` / `HexWord(usize::from(err) as u32)` lets the linker drop `<CaliptraApiError as Debug>::fmt` (~644 B), `<McuError as Debug>::fmt`, `<FlashDrvError as Debug>::fmt`, and all their variant-name strings.

`mcu-rom-emulator --features hw-2-1` .text: 72,364 → 70,704 B (saved 1,660 B / 1.62 KB).

### DIFF
--- a/rom/src/cold_boot.rs
+++ b/rom/src/cold_boot.rs
@@ -590,8 +590,8 @@ impl ColdBoot {
         if let Err(err) = lc.transition(LifecycleControllerState::Scrap, &LifecycleToken([0u8; 16]))
         {
             romtime::println!(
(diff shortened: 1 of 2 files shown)


swenson (approved):
LGTM

PR #1403, opened 2026-05-11
fuses: emit each fuse entry as its own const

ternst-nv:
Replace the single FUSE_ENTRY_TABLE array with one standalone &FuseEntryInfo const per entry so the linker can drop unreferenced entries. Saves ~916 bytes of .text in the emulator ROM (hw-2-1).


swenson (approved):
LGTM; I really thought I had done this already :)

PR #1408, opened 2026-05-11
spdm-lib: Share large message buffer via static pool

parvathib:
Replace per-task inline buffer with a shared static pool using the LargeMsgBufProvider trait. MCTP and DOE responders acquire the buffer on demand and release when idle. Saves ~4 KB .bss when both transports are enabled.


swenson (approved):
LGTM

PR #1412, opened 2026-05-12
spdm-lib: Remove measurement staging buffer, use shared large msg buf

parvathib:
Remove the 4 KB MeasurementRecord::data inline buffer from SpdmContext. Measurements are now fetched directly into large_msg_ctx.buf, which is already acquired for GetMeasurements. Also acquire the buffer for Challenge and KeyExchange (measurement summary hash).

.bss savings: 4,072 B per responder task.
Increase user-app stack from 0xa000 to 0xae00, enabled by the savings.


ArthurHeymans (approved):

swenson (approved):
LGTM

PR #1414, opened 2026-05-12
rom: Remove dot blob debug output

ternst-nv:
Remove the dot blob debug statement, saving ~3Kb in ROM imem.  The raw blob is still output and could be decoded, so information is not lost.


pjhades (approved):

swenson (approved):
LGTM

PR #1415, opened 2026-05-12
ocp: demo ocp lib utilization

ternst-nv:
Integrate the OCP Software Library into ROM, under an optional flag demonstrating how to use a USB interface to achieve an OCP Recovery boot.  This includes creating a manager interface to allow the integrator to determine what the boot order of the chip should be.  Appropriate integration tests were also added.


platforms/fpga/rom/src/riscv.rs
```diff
@@ -210,13 +210,23 @@ pub extern "C" fn rom_entry() -> ! {
 
     let hooks = LoggingRomHooks;
 
+    use mcu_rom_common::recovery::flash::FlashImageProvider;
```
swenson: nit: move imports to the top

swenson (approved):
LGTM

common/ocp/src/cms/slice_indirect.rs
```diff
@@ -94,15 +105,15 @@ impl IndirectCmsRegion for SliceIndirectRegion<'_> {
...
-        let available = size - offset;
-        let copy_len = buf.len().min(available);
-        buf[..copy_len].copy_from_slice(&self.buf[offset..offset + copy_len]);
+        let copy_len = self.read_at(self.imo as usize, buf);
         self.advance_imo(copy_len);
         Ok(copy_len)
     }
 
+    fn device_read(&self, offset: u32, buf: &mut [u8]) -> usize {
+        self.read_at((offset & !0x3) as usize, buf)
```
pjhades: nit: define the alignment as a constant.

pjhades (approved):
LGTM!

builder/src/all.rs
```diff
@@ -637,6 +639,24 @@ pub fn all_build(args: AllBuildArgs) -> Result<()> {
         }
     }
 
+    // Build ROM-only test features (e.g. test-usb-ocp-recovery) that aren't runtime
+    // features but need their own ROM binary. Each is built with the base rom_features
+    // combined with the test feature so the resulting binary matches what tests expect.
+    let rom_test_feature_list: Vec<&str> = match rom_test_features {
```
timothytrippel, line 645: What is the difference between this and the "ROM_ONLY_TEST_FEATURES" above? Why doe we need two test feature arrays?

Issue #1190, opened 2026-04-07
Checksum Validation Failure in MCU Mailbox Crypto Passthrough


Behemoth1:
Checksum validation fails in handle_crypto_passthrough when forwarding cryptographic operation responses from Caliptra to the MCU mailbox, causing "Transport error" even when the underlying crypto operation succeeds (208 bytes response).

The handle_crypto_passthrough function acts as a bridge between:
FW → sends MCU mailbox commands (e.g., MC_SHA_INIT, MC_HMAC)
MCU Runtime → translates to Caliptra commands (e.g., CM_SHA_INIT, CM_HMAC)
Caliptra Runtime → processes crypto operations and returns responses
The Issue: Each mailbox protocol calculates checksums differently:

Caliptra response: checksum = calc_checksum(CALIPTRA_CMD_ID, payload)
MCU mailbox expects: checksum = calc_checksum(MCU_CMD_ID, payload)

When handle_crypto_passthrough forwards Caliptra's response directly to msg_buf without recalculating the checksum, the transport layer's verification fails:
// transport.rs:72
if !verify_checksum(hdr.chksum, self.cmd_opcode, payload) {
    return Err(TransportError::ChkSumMismatch);  // ← Fails here
}

Added checksum recalculation in handle_crypto_passthrough after copying Caliptra's response:

// Copy response from Caliptra to msg_buf
msg_buf[..resp_len].copy_from_slice(&resp_buf[..resp_len]);

// Recalculate checksum for MCU mailbox protocol
use mcu_mbox_common::messages::calc_checksum;
use core::mem::size_of;

if resp_len >= size_of::<MailboxRespHeader>() {
    // Calculate checksum with MCU command ID
    let payload = &msg_buf[size_of::<u32>()..resp_len];
    let new_checksum = calc_checksum(mcu_cmd_code, payload);
    
    // Update checksum in response header
    msg_buf[0..4].copy_from_slice(&new_checksum.to_le_bytes());
}

Added mcu_cmd_code parameter to distinguish between MCU and Caliptra command IDs:

pub async fn handle_crypto_passthrough<T: Default + IntoBytes + FromBytes>(
    &self,
    msg_buf: &mut [u8],
    req_len: usize,
    mcu_cmd_code: u32,           // MCU mailbox command ID (for checksum)
    caliptra_cmd_code: u32,      // Caliptra command ID (for mailbox API)
    resp_buf: &mut [u8],
[...]

PR #1419, opened 2026-05-12
mcu-mbox-lib: recalc response checksum with MCU cmd ID in passthrough

sviatlana-yarmolik:
The Caliptra mailbox response returned by execute_mailbox_cmd() carries a checksum computed over the Caliptra command ID, but the MCU mailbox peer verifies the response checksum against the MCU command ID it sent. Without recalculation the peer rejects every crypto passthrough response with ChkSumMismatch.

Thread the MCU command ID into handle_crypto_passthrough and recompute the response checksum before returning.

Fixes https://github.com/chipsalliance/caliptra-mcu-sw/issues/1190

### DIFF
--- a/runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
+++ b/runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
@@ -128,6 +128,7 @@ impl<'a> CmdInterface<'a> {
                 self.handle_crypto_passthrough::<McuFipsSelfTestStartReq>(
                     msg_buf,
                     req_len,
+                    cmd,
                     CaliptraCommandId::SELF_TEST_START.into(),
                     &mut resp_bytes,
                 )
@@ -138,6 +139,7 @@ impl<'a> CmdInterface<'a> {
                 self.handle_crypto_passthrough::<McuFipsSelfTestGetResultsReq>(
                     msg_buf,
                     req_len,
+                    cmd,
                     CaliptraCommandId::SELF_TEST_GET_RESULTS.into(),
                     &mut resp_bytes,
                 )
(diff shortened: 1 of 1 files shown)


RaunakGu (approved):

RaunakGu:
hi @swenson, @clundin25, Can you please check and if changes look good, help to approve, thank you

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -659,6 +689,13 @@ impl<'a> CmdInterface<'a> {
         match status {
             Ok(resp_len) => {
                 msg_buf[..resp_len].copy_from_slice(&resp_buf[..resp_len]);
+                // Recalculate checksum for MCU mailbox protocol.
+                // Caliptra response has checksum for Caliptra command, but we need checksum for MCU command.
+                if resp_len >= core::mem::size_of::<MailboxRespHeader>() {
+                    let payload = &msg_buf[core::mem::size_of::<u32>()..resp_len];
+                    let new_checksum = mcu_mbox_common::messages::calc_checksum(mcu_cmd_code, payload);
+                    msg_buf[0..4].copy_from_slice(&new_checksum.to_le_bytes());
```
clundin25: nit: Use `size_of` like line 695

sviatlana-yarmolik: Fixed in 14dc039d

clundin25 (approved):

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -659,6 +689,15 @@ impl<'a> CmdInterface<'a> {
         match status {
             Ok(resp_len) => {
                 msg_buf[..resp_len].copy_from_slice(&resp_buf[..resp_len]);
+                // Recalculate checksum for MCU mailbox protocol.
+                // Caliptra response has checksum for Caliptra command, but we need checksum for MCU command.
+                if resp_len >= core::mem::size_of::<MailboxRespHeader>() {
```
zhalvorsen, line 694: There shouldn't be any responses without a response header. So it would probably be better to return an error here even though it should never use that code path. This could help us catch ourselves as early as possible where the header may have been forgotten.

PR #1424, opened 2026-05-13
spdm: Convert VCA-only command handlers to sync

parvathib:
- Convert handle_get_version, handle_get_capabilities,
  handle_negotiate_algorithms, and handle_end_session to sync functions.
  These only use VCA transcript appends (synchronous buffer copies).
- Add sync append_message_to_vca_transcript() on SpdmContext.
- Replace write!() formatting in version_to_str with manual decimal
  encoding to avoid pulling in core::fmt machinery.

Saves ~2.8 KB .text in user-app.


swenson (approved):
LGTM, love it

PR #1428, opened 2026-05-13
[bugfix] fix default FW_SRAM_EXEC_REGION_SIZE calculation in ROM

amit-kumarh:
The calculation for `FW_SRAM_EXEC_REGION_SIZE` was using byte sizes instead of 4KB block counts, which is not accurate.

https://chipsalliance.github.io/caliptra-ss/v2_1/regs/?p=soc.mci_top.mci_reg.FW_SRAM_EXEC_REGION_SIZE

### DIFF
--- a/rom/src/cold_boot.rs
+++ b/rom/src/cold_boot.rs
@@ -565,8 +565,8 @@ impl BootFlow for ColdBoot {
         mci.set_flow_checkpoint(McuRomBootStatus::McuMboxAxiUsersConfigured.into());
 
         let size_value = params.mcu_fw_sram_exec_region_size.unwrap_or(
-            unsafe { MCU_MEMORY_MAP.sram_size }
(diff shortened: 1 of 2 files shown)


swenson (approved):
LGTM

karri-next:
A friendly question @amit-kumarh 
How did you find this bug? Are there public FPGA regression results somewhere?

For instance on the "emulator" platform, my experience is that modifying this reg (or the MCU ROM config for the AXI user IDs), has no affect on access to the MCU SRAM.
I was just about to file this as a bug before I noticed your PR :)

amit-kumarh:
Unfortunately, there are currently no additional FPGA tests beyond those that are in the repo and runs in CI. That said, I don't think AXI user ID filtering is implemented on the emulator, which is why I had to test the original change entirely on an FPGA.

As far as this particular bug, it was caught by @tonywangxy https://github.com/chipsalliance/caliptra-mcu-sw/pull/1261#issuecomment-4427917930

PR #1430, opened 2026-05-13
ocp: Add tests validating image load

ternst-nv:
Add integration tests validating a complete image load can occur and boot the system into a good image state.  This included a couple small bug patches in regard to debug output.  It also enabled Staging RAM within the emulator to allow the INDIRECT CMS concept to be validated.

Finally a number of utilities were added to simplify test writing.


pjhades (approved):
LGTM

clundin25 (approved):

PR #1431, opened 2026-05-13
mctp: Reduce MCTP_MAX_MESSAGE_SIZE from 2048 to 1024

parvathib:
SPDM responder uses MAX_SPDM_RESPONDER_BUF_SIZE = 1024 for message buffers. The kernel MCTP driver allocated 2048 per buffer, double what is needed. Each MCTP driver instance has 3 message buffers (rx, tx, buffered_rx), and there are 3 driver instances (SPDM, PLDM, Caliptra).

Saves ~8 KB kernel .sram (9 buffers × 1024 B reduction).


swenson (approved):
LGTM

PR #1432, opened 2026-05-13
ocp: no panic ocp

ternst-nv:
Introduce a number of small changes to remove the possibility of the ROM panicking with the ocp feature enabled.

Then add the ocp feature to the list of features validated as panic free within the precheckin.

### DIFF
--- a/platforms/emulator/rom/usb/src/lib.rs
+++ b/platforms/emulator/rom/usb/src/lib.rs
@@ -14,7 +14,7 @@ use registers_generated::usbdev::bits::*;
 use romtime::StaticRef;
 use tock_registers::interfaces::{Readable, Writeable};
 use tock_registers::LocalRegisterCopy;
-use zerocopy::{FromBytes, IntoBytes};
+use zerocopy::IntoBytes;
 
 type RxEntry = LocalRegisterCopy<u32, Rxfifo::Register>;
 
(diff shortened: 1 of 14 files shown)


xtask/src/test.rs
```diff
@@ -151,31 +151,27 @@ pub(crate) fn test_hello_c_emulator() -> Result<()> {
...
+        let rom_elf = PROJECT_ROOT
+            .join("target")
+            .join(TARGET)
+            .join("release")
+            .join("mcu-rom-emulator");
+        let rom_elf = std::fs::read(rom_elf)?;
+        let symbols = elf_symbols(&rom_elf)?;
+        if symbols.iter().any(|s| s.contains("panic_is_possible")) {
+            bail!(
+                "The MCU ROM contains the panic_is_possible symbol, which is not allowed. \
```
swenson, line 171: nit: I think we should add which feature this was compiled with to make it easier to debug when we do get a panic.

platforms/emulator/rom/usb/src/lib.rs
```diff
@@ -51,32 +51,43 @@ impl ExamplarUsbDriver {
...
+        let w0 = self.regs.buffer.get(base).map_or(0, |r| r.get());
+        let w1 = self.regs.buffer.get(base + 1).map_or(0, |r| r.get());
+        let w0b = w0.to_le_bytes();
+        let w1b = w1.to_le_bytes();
+        let raw_bytes: [u8; SETUP_PACKET_LEN] = [
+            w0b[0], w0b[1], w0b[2], w0b[3], w1b[0], w1b[1], w1b[2], w1b[3],
+        ];
+        // Safety: SetupPacket is repr(C, packed) + FromBytes (any bit pattern valid),
+        // and SETUP_PACKET_LEN == size_of::<SetupPacket>() (compile-time assert).
+        unsafe { core::mem::transmute::<[u8; SETUP_PACKET_LEN], SetupPacket>(raw_bytes) }
```
swenson: Does the `packed` prevent us from using `zerocopy`? Generally I prefer `zerocopy`, which avoids unsafe.

swenson (approved):
LGTM

PR #1433, opened 2026-05-13
[cherry-pick] remaining PK hash selection commits

amit-kumarh:
* Lock PK Hash slots after selection
* Enable use of a strap to skip one PK Hash slot for rotation


swenson (approved):
LGTM

PR #1438, opened 2026-05-13
docs: categorize ROM fuses and document recovery PK hash format

swenson:
Reorganize docs/src/rom-fuses.md as the authoritative list of every OTP fuse the reference MCU ROM consumes today or will consume shortly:

- Add an up-front Fuse Categories section splitting fuses into Mandatory (Caliptra core fuses + iTRNG straps), Optional (could be hardcoded or strapped, e.g., SoC stepping ID, IDevID attrs, owner PK hash, prod debug unlock PKs, encrypted-firmware AES key), and Recommended but not required (feature-specific: DOT, MCU SVN anti-rollback, OCP LOCK), with logical sizes for each entry.
- Scope is limited to Caliptra core and Caliptra MCU fuses; broader SoC-level provisioning (PLL/clock trim, AST, pinmux, watchdog, SKU, etc.) is out of scope and noted at the top with a reference to OpenTitan's CREATOR_SW_CFG partition as prior art.
- Move CPTRA_CORE_IDEVID_CERT_IDEVID_ATTR into Mandatory and document the actual usage: Caliptra ROM consumes only the first 16 of the 24 u32 entries (Flags, ECC/MLDSA SubjectKeyId, UeidType, ManufacturerSerialNumber), so 64 of the 96 bytes need provisioning.
- Add an optional fw_encryption_key fuse for encrypted firmware deployments (Single{bits:K}, K in {128, 192, 256}).
- Add vendor_recovery_pk_hash to the Fuse Field Reference and OTP Encoding Recommendations, noting that integrators needing rotation should add vendor_recovery_pk_hash_{0..N} plus a vendor_recovery_pk_hash_valid revocation bitmask.

Document the vendor recovery PK hash format in docs/src/dot.md: the format is intentionally aligned with Caliptra's CPTRA_OWNER_PK_HASH and the production debug unlock PK hashes, so an integrator can reuse the same (ECC P-384, MLDSA-87) key pair across all three roles. Hash is SHA-384 over dword-reversed ECC X || dword-reversed ECC Y || raw FIPS 204 MLDSA bytes (2688 bytes total), and the 48-byte digest is stored in OTP in dword-reversed form. Add a fully worked example with hex output that integrators can reproduce with python3 + hashlib. Flag in an implementation-status callout that aligning the current DOT code and integration test with this format is tracked separately.

Remove MCU_RT_MIN_SVN from docs/src/svn.md. The MCU Runtime SVN floor is enforced by Caliptra Core against CPTRA_CORE_SOC_MANIFEST_SVN; MCU ROM only performs the OTP write on Caliptra Core's behalf. Removing the redundant MCU-side floor simplifies the design and removes the McuImageHeader.min_svn field and the ImageVerifier section. The MCU Component SVN Manifest and SOC_IMAGE_MIN_SVN[i] continue to exist unchanged.


docs/src/provisioning.md
```diff
@@ -389,14 +389,14 @@ The fuse partition below corresponds to the Caliptra Subsystem 2.1.1 [reference
...
-|     CPTRA_SS_LOCK_HEK_PROD_0      |    CPTRA_SS_LOCK_HEK_PROD_0_RATCHET_SEED    |      48      | PROD                |
-|     CPTRA_SS_LOCK_HEK_PROD_1      |    CPTRA_SS_LOCK_HEK_PROD_1_RATCHET_SEED    |      48      | PROD                |
-|     CPTRA_SS_LOCK_HEK_PROD_2      |    CPTRA_SS_LOCK_HEK_PROD_2_RATCHET_SEED    |      48      | PROD                |
-|     CPTRA_SS_LOCK_HEK_PROD_3      |    CPTRA_SS_LOCK_HEK_PROD_3_RATCHET_SEED    |      48      | PROD                |
-|     CPTRA_SS_LOCK_HEK_PROD_4      |    CPTRA_SS_LOCK_HEK_PROD_4_RATCHET_SEED    |      48      | PROD                |
-|     CPTRA_SS_LOCK_HEK_PROD_5      |    CPTRA_SS_LOCK_HEK_PROD_5_RATCHET_SEED    |      48      | PROD                |
-|     CPTRA_SS_LOCK_HEK_PROD_6      |    CPTRA_SS_LOCK_HEK_PROD_6_RATCHET_SEED    |      48      | PROD                |
-|     CPTRA_SS_LOCK_HEK_PROD_7      |    CPTRA_SS_LOCK_HEK_PROD_7_RATCHET_SEED    |      48      | PROD                |
+|     CPTRA_SS_LOCK_HEK_PROD_0      |    CPTRA_SS_LOCK_HEK_PROD_0_RATCHET_SEED    |      32      | PROD                |
```
timothytrippel, line 392: The fields are 48 bytes wide aren't they? https://github.com/chipsalliance/caliptra-ss/blob/main/src/fuse_ctrl/data/otp_ctrl.hjson#L2472

swenson: The partition is 48 bytes, including 8 byte digest and 8 byte zeroization. The fields are 32 bytes.

swenson: The fuse registers are also 8x4 = 32 bytes: https://chipsalliance.github.io/caliptra-rtl/main/internal-regs/?p=clp.soc_ifc_reg.fuse_hek_seed%5B0%5D

timothytrippel: ahh good catch.

timothytrippel (approved):

swenson:
Thanks!

PR #1439, opened 2026-05-13
kernel: Add capsule logging and optimize mctp to reduce .text

parvathib:
- Add capsule_error!/capsule_debug! with feature-gated debug prints
- Replace {:?} with hex error codes across MCTP, DOE, MBOX capsules
- Add CaliptraApiError match with hex codes in mailbox capsule
- Table-driven MCTP version support entries

Saves ~5 KB kernel .text.


mlvisaya (approved):
LGTM!

PR #1440, opened 2026-05-14
pldm-lib: replace FdInternal async Mutex with RefCell

helloxiling:
Switch `FdInternal::inner` from `Mutex<NoopRawMutex, FdInternalInner>` to `RefCell<FdInternalInner>`, drop `async` from its 38 accessors, and remove 91 `.await` call sites in `FdContext`. It is still safe because the two PLDM tasks (responder, initiator) run on the same single-thread cooperative `TockExecutor` (embassy `arch-riscv32`). Every original critical section was `lock().await; <one sync stmt>; drop` — never held across an `.await` — so the cooperative scheduler, not the mutex, was serialising those accesses. `RefCell::borrow*()` covers the same single- statement sections; no overlapping borrow is reachable. Cross-task races remain handled by `CancellationFlag` (`AtomicBool`).

User-app built with pldm-lib memory footprints reduction vs origin/main:
  feature                                 code size     .bss
  test-pldm-discovery             -7,618 B    -1,152 B
  test-pldm-fw-update           -7,616 B    -1,152 B
  test-pldm-fw-update-e2e    -7,622 B    -1,152 B
  test-pldm-streaming-boot   -7,686 B   -1,152 B


mlvisaya (approved):
LGTM

PR #1447, opened 2026-05-14
spdm: flatten error codes into 2 u32 values

parvathib:
- Add error_code()/ext_code() u32s on SpdmError replacing Debug-based console prints.
- Introduce stable wrapper-type IDs (error_type_id, 0xE0..0xFF) so a given inner error type uses the same byte value at every nesting depth.
- Group related variants into sub-enums so SpdmError stays simple.
- Replace &'static str-carrying variants across the SPDM error types with named variants.

Saves ~22 KB user-app .text.


swenson (approved):
LGTM

PR #1448, opened 2026-05-14
rom: align DOT recovery PK hash with caliptra-sw fuse layout

swenson:
So that a single 48-byte hash can be burned into cptra_ss_owner_pk_hash, debug-unlock vendor PK hash, and vendor_recovery_pk_hash, match the caliptra-sw conventions in the DOT path:

* OTP storage: each 4-byte word of the digest is stored byte-reversed (caliptra-sw `to_hw_format` decodes ECC via `u32::from_be_bytes`). `recovery_pk_hash_from_otp_bytes` undoes that on load.

* SHA input: the ECC public key is per-dword-reversed before hashing (`to_hw_format`), the MLDSA public key is hashed in natural order (the crypto layer uses `u32::from_le_bytes` for MLDSA). `cm_owner_pk_hash_sha384` swap_bytes-es each ECC u32 word before calling `cm_sha384`. Used at all three PK-hash check sites.

Tests cover the documented example 48afdb07...54988bd7 -> fuse bytes 07dbaf48...d7988b54, the SHA input swap convention, and a new integration test that builds OTP via the owner-PK-hash storage path and runs the full DOT_OVERRIDE challenge/response.

### DIFF
--- a/rom/src/device_ownership_transfer.rs
+++ b/rom/src/device_ownership_transfer.rs
@@ -35,6 +35,29 @@ pub trait OwnerPolicy {}
 #[derive(Clone, FromBytes, IntoBytes, Immutable, KnownLayout)]
 pub struct RecoveryPkHash(pub [u32; 12]);
 
+/// Convert a 48-byte recovery PK hash payload read from OTP into a
+/// [`RecoveryPkHash`] by reversing the bytes within each 4-byte word.
+///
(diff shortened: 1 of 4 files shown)


mlvisaya (approved):
LGTM

mhatrevi (approved):

swenson:
Thanks!

PR #1450, opened 2026-05-14
rom: send ACTIVATE_FIRMWARE INITIAL_ACTIVATE after encrypted decrypt

swenson:
On the encrypted-boot path, after MCU ROM decrypts firmware in MCU SRAM via CM_AES_GCM_DECRYPT_DMA, send ACTIVATE_FIRMWARE with the new INITIAL_ACTIVATE flag so Caliptra runtime publishes FW_EXEC_CTRL[MCU]. Without this, MCI's BOOT_RST_MCU state would hold MCU in reset forever after the warm reset, since mcu_sram_fw_exec_region_lock gates reset release.

Bumps the caliptra-sw git pin to a6e87753 (current main) to pick up the new `flags` field on `ActivateFirmwareReq` and the `ActivateFirmwareFlags::INITIAL_ACTIVATE` constant added in chipsalliance/caliptra-sw#3720, and sets `flags: 0` on the existing `ActivateFirmwareReq` initializers in image_loader and firmware_update.

This will enable us to re-enable the encrypted FW test in caliptra-sw.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -330,7 +330,7 @@ checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
 [[package]]
 name = "caliptra-api"
 version = "0.1.0"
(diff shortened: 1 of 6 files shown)


mlvisaya (approved):
LGTM

swenson:
Thanks!

Issue #1178, opened 2026-04-06
Add CFI protections in firmware


swenson:
Priority is for ROM

Runtime would want to be pragmatic.

PR #1451, opened 2026-05-14
rom: add preliminary CFI hardening (#1178)

pjhades:
This commit attempts to harden the CFI of the MCU ROM with the help of the caliptra-cfi crate.

To summarize the changes:
- A handler is added for CFI panics
- CFI-related error types are added
- Enhance CFI on cold boot, DOT and fuse writing function


romtime/src/register.rs
```diff
@@ -0,0 +1,19 @@
...
+use tock_registers::interfaces::Writeable;
+use tock_registers::{registers::ReadWrite, RegisterLongName};
+
+pub fn cfi_set_register<R: RegisterLongName>(reg: &ReadWrite<u32, R>, val: u32) {
+    let val_copy = cfi_launder(val);
+    reg.set(val);
+    CfiCounter::delay();
+    reg.set(val_copy);
+    cfi_assert_eq(val_copy, val);
+}
```
pjhades: This one is open for discussion. I was considering reading the written value back and then comparing them. But that depends on if the register can be reliably read back. I think that could 1) make it harder to maintain because every time when we need to harden a register write we have to make sure the register supports read-back; and 2) introduce the inconsistency between how we harden different register writes.

romtime/Cargo.toml
```diff
@@ -20,7 +22,8 @@ caliptra-ureg.workspace = true
 zerocopy.workspace = true
 
 [features]
-default = []
+default = ["cfi"]
```
pjhades, line 25: I followed the same pattern used in caliptra-sw, where:
1. The `cfi` feature is enabled by default.
2. The attribute macros `cfi_impl_fn` and `cfi_mod_fn` are gated by the `cfi` feature.
3. Other CFI APIs and macros like `cfi_launder` are not gated by the `cfi` feature.

rom/src/rom.rs
```diff
@@ -961,6 +1010,21 @@ pub fn rom_start(params: RomParameters) {
...
+        digest.update(cycle1);
+        digest.update(cycle2);
+        let digest: [u8; 32] = digest.finalize().into();
+        let words: [u32; 8] = transmute!(digest);
+        Ok((words[0], words[1], words[2], words[3]))
+    };
+
+    CfiCounter::reset(&mut entropy_gen);
+    CfiCounter::reset(&mut entropy_gen);
+    CfiCounter::reset(&mut entropy_gen);
```
pjhades: We need to initialize the CFI state with some entropy source for the PRNG, before we use the CFI counter. But at this stage, we don't have a good one available. So the intention here is to initialize it with some "low-quality" source and re-initialize it once the Caliptra mailbox is available.

I'd like to have some input here as for what other sources can be used besides the cycle counts.

error/src/lib.rs
```diff
@@ -542,6 +542,7 @@ impl McuError {
             0x5_0017,
             "DCCM uncorrectable ECC error"
         ),
+        (ROM_CFI_PANIC, 0x6_0000, "ROM CFI panic"),
```
pjhades, line 645: We could define the CFI-related errors here, but `cfi_panic_handler` expects a `u32` and the `CfiPanicinfo` type information is lost. Converting the `u32` back to `CfiPanicInfo` requires hardcoding the `u32` value of enum variants. So I think a better way would be to first explicitly specify the `u32` values for such variants from caliptra-cfi upstream, and implement `From<32>` for `CfiPanicInfo` there.

rom/src/cold_boot.rs
```diff
@@ -1140,24 +1148,42 @@ impl BootFlow for ColdBoot {
...
+        cfi_assert!(mci.is_ss_config_done_sticky());
+        cfi_assert!(mci.is_ss_config_done());
 
         // Verify PK hashes haven't been tampered with after locking
         romtime::println!("[mcu-rom] Verifying production debug unlock PK hashes");
-        if let Err(err) = verify_prod_debug_unlock_pk_hash(mci, otp) {
-            romtime::println!("[mcu-rom] PK hash verification failed");
-            fatal_error(err);
+        let result = verify_prod_debug_unlock_pk_hash(mci, otp);
+        if cfi_launder(result.is_ok()) {
```
ternst-nv, line 1172: Taking a look at the base library, why do we need to use both `launder` and `assert`?

rom/src/cold_boot.rs
```diff
@@ -1140,24 +1148,42 @@ impl BootFlow for ColdBoot {
         mci.set_flow_checkpoint(McuRomBootStatus::SsConfigDoneSet.into());
 
         // Verify that SS_CONFIG_DONE_STICKY and SS_CONFIG_DONE are actually set
-        if !mci.is_ss_config_done_sticky() || !mci.is_ss_config_done() {
+        if cfi_launder(!mci.is_ss_config_done_sticky()) {
             romtime::println!("[mcu-rom] SS_CONFIG_DONE verification failed");
             fatal_error(McuError::ROM_SOC_SS_CONFIG_DONE_VERIFY_FAILED);
         }
+        if cfi_launder(!mci.is_ss_config_done()) {
```
ternst-nv, line 1162: Others can comment here, but I think these register lifecycle checks would be lower priority than validating digests or registers controlling debug/configuration parameters.

rom/src/cold_boot.rs
```diff
@@ -1140,24 +1148,42 @@ impl BootFlow for ColdBoot {
...
+            }
         }
         mci.set_flow_checkpoint(McuRomBootStatus::PkHashVerified.into());
 
         // Verify MCU mailbox AXI users haven't been tampered with after locking
         romtime::println!("[mcu-rom] Verifying MCU mailbox AXI users");
-        if let Err(err) = verify_mcu_mbox_axi_users(mci, &mcu_mbox_config) {
-            romtime::println!("[mcu-rom] MCU mailbox AXI user verification failed");
-            fatal_error(err);
+        let result = verify_mcu_mbox_axi_users(mci, &mcu_mbox_config);
```
ternst-nv, line 1182: 👍 , this looks like a good place to check

rom/src/rom.rs
ternst-nv: Hmm, that is a good question, and I don't know if cycle count will work.  @swenson are you aware of any hardware integrations with an RNG source which could be used here?

romtime/src/register.rs
ternst-nv: Others can comment here as well, but I would be less concerned with register writes then register reads.

rom/src/cold_boot.rs
pjhades: My understanding is that we'd like to achieve multiple things:
- We use `launder` to prevent the compiler from optimizing away the if-else block.
- Once the execution lands in either branch, we double check the result for some added redundancy.
- We also want to print the error message when the result is `Err`.

What's the alternative pattern were you thinking about? I might be overthinking when writing this.

RaunakGu:
@zhalvorsen - Can we please add this in a milestone, thank you

zhalvorsen:
> @zhalvorsen - Can we please add this in a milestone, thank you

It is added to the 2.1 ROM release using the tracking issue https://github.com/chipsalliance/caliptra-mcu-sw/issues/1178

RaunakGu:
@pjhades, When are we targeting to close this PR?

pjhades:
@RaunakGu I just turned this PR to ready-for-review. I will ping people to review it and I hope we can close it this week. Please note that this PR isn't supposed to harden everything but only serves as the beginning work of CFI protection.

I think the remaining uncertain thing is still the availability of an entropy source that can be used for the CFI counters before Caliptra mailbox is online.

romtime/src/register.rs
pjhades: I dropped this change. We can consider adding it back in a future PR if necessary.

rom/src/cold_boot.rs
ternst-nv: NVM, this looks like the correct pattern.

rom/src/cold_boot.rs
```diff
@@ -1149,24 +1155,44 @@ impl BootFlow for ColdBoot {
...
         // Verify PK hashes haven't been tampered with after locking
         caliptra_mcu_romtime::println!("[mcu-rom] Verifying production debug unlock PK hashes");
-        if let Err(err) = verify_prod_debug_unlock_pk_hash(mci, otp) {
-            caliptra_mcu_romtime::println!("[mcu-rom] PK hash verification failed");
-            fatal_error(err);
+        let result = verify_prod_debug_unlock_pk_hash(mci, otp);
+        if cfi_launder(result.is_ok()) {
+            cfi_assert!(result.is_ok());
+        } else {
+            cfi_assert!(result.is_err());
```
ternst-nv: Hmm, I think we want to remove the `cfi_assert` here as we'd like to fail into the normal `fatal_error`. Ditto for subsequent error calls.

rom/src/lib.rs
```diff
@@ -153,3 +153,14 @@ pub fn err_code(err: &CaliptraApiError) -> u32 {
         _ => 0xdead_ffff,
     }
 }
+
+#[allow(unused)]
```
ternst-nv: I think we can remove this if `#[used]` is also present?

ternst-nv (commented):
Looks like a good start, I think the big remaining question for me is the initial entropy source.

rom/src/cold_boot.rs
swenson: Generally I think that is correct -- security critical things should be guarded by CFI. However, if someone managed to glitch past the stocky SS_CONFIG_DONE, we would probably want to stop booting, so this one in particular might be good to check with CFI.

rom/src/rom.rs
swenson: Cycle count will be pretty deterministic in real hardware -- our FPGA it depends on the host timing, but the start cycle count will almost always fall in a particular range.

I don't think we have a real good option other than requesting that integrators provide *some* randomness source -- it doesn't have to be FIPS since we only use this for CFI.

swenson (approved):
LGTM

rom/src/rom.rs
```diff
@@ -973,13 +990,35 @@ pub struct RomParameters<'a> {
...
+    /// mailbox is available. If `None` and CFI is enabled, initialization will error out.
+    pub cfi_entropy_source: Option<&'a mut dyn CfiEntropySource>,
 }
 
-pub fn rom_start(params: RomParameters) {
+fn initialize_cfi_state(params: &mut RomParameters) {
+    match &mut params.cfi_entropy_source {
+        None => {
+            if cfg!(feature = "cfi") {
+                caliptra_mcu_romtime::println!("[mcu-rom] Invalid reset reason: multiple bits set");
```
ternst-nv: [nit] Seems like this error statement should be updated.

ternst-nv (approved):
LGTM

rom/src/rom.rs
pjhades: Oops..copy pasta. updated.

PR #1457, opened 2026-05-15
Put whole idevID Cert in ExternalOTP

mlvisaya:
Change the ExternalOTP partition to store the whole idevID certificate, initialize it with an example cert


platforms/emulator/runtime/userspace/apps/example/src/test_external_otp.rs
```diff
@@ -20,11 +20,11 @@ pub(crate) fn test_external_otp() {
     assert!(count >= 2, "expected at least 2 partitions, got {}", count);
     println!("  partition_count: {}", count);
 
-    // Partition 0x01: ECC signature (96 bytes).
+    // Partition 0x01: ECC Cert (547 bytes).
     let ecc_size = otp.partition_size(0x01).unwrap();
     assert_eq!(
-        ecc_size, 96,
-        "expected ECC partition size 96, got {}",
+        ecc_size, 547,
```
parvathib, line 64: can we compare the data matches too

mlvisaya: added comparison

parvathib (approved):
LGTM

PR #1461, opened 2026-05-15
[fuse] Clarify VENDOR_PK_HASH_VALID usage

amit-kumarh:
Not a functional change - update fuse description and add clarification for VENDOR_PK_HASH_VALID that burnt bits indicate an invalidated PK hash entry


swenson (approved):
LGTM

PR #1466, opened 2026-05-16
Fix UpdateComponentResponse::decode() to handle failure responses per…

mlvisaya:
… PLDM spec

The PLDM Firmware Update specification (DSP0267) states that when a response has a non-success completion code, only the message header and completion_code fields are present in the payload (4 bytes total).

Root cause: When the FD (Firmware Device) receives an UpdateComponent request while not in the ReadyXfer state (e.g. due to a timing/race condition in the MCTP transport), it returns an InvalidStateForCommand (0x84) completion code. Per the PLDM spec, failure responses only contain 4 bytes (header + completion_code). The decoder expected 12+ bytes for ALL responses, failing to parse valid failure responses.

This caused the UA to drop the packet as "Unhandled", never transition out of the ReadyXfer state, and ultimately hang the firmware update protocol.

The fix checks the completion_code immediately after reading it. If it is not PldmBaseCompletionCode::Success, the decoder returns early with the header and completion_code populated and remaining fields set to their default values. This allows the UA to properly process failure responses and handle error recovery.

### DIFF
--- a/common/pldm/src/message/firmware_update/update_component.rs
+++ b/common/pldm/src/message/firmware_update/update_component.rs
@@ -2,7 +2,8 @@
 
 use crate::codec::{PldmCodec, PldmCodecError};
 use crate::protocol::base::{
-    InstanceId, PldmMsgHeader, PldmMsgType, PldmSupportedType, PLDM_MSG_HEADER_LEN,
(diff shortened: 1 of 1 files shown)


helloxiling (approved):
Looks good to me.
Thanks for addressing it, Marco!

PR #1468, opened 2026-05-17
emulator: clean up direct-read flash and linker reservation

helloxiling:
The log capsule now goes through the flash controller HIL (PR #1445),
so the 64 MB CPU-side mirror at 0x3800_0000 is no longer accessed by
firmware.  Remove the dead pieces of the old direct-read path:

  * Emulator: drop the DIRECT_READ_FLASH_ORG/SIZE constants, the read
    dispatch in McuRootBus, and the direct_read_flash_offset/size
    fields.  Also drop the host-side RAM cache that used to mirror the
    flash file backend; page reads now seek + read_exact against the
    file directly, which is the source of truth.
  * Firmware bundler: drop the templated 'flash' MEMORY region, the
    FLASH_OFFSET/FLASH_LENGTH kernel substitution, and the unused
    Platform::flash field/method.
  * Kernel linker template: drop the .storage section and the
    _sstorage / _estorage symbols.
  * Emulator runtime board: drop the externs for _sstorage/_estorage
    and the corresponding 'Logging flash region' PMP entry.
  * Manifests: drop the [platform.flash] block from the emulator
    example/user apps, hw-model test-fw, and bare-metal runtime.


mlvisaya (approved):
LGTM

PR #1472, opened 2026-05-18
rom: add ROM hooks for OCP LOCK fuses, stable owner key, encrypted FW…

swenson:
… decrypt

Extends the RomHooks trait (#1270) with hook pairs for three 2.1-specific cold-boot flows so integrators can observe them without patching the common ROM:

- pre/post_set_ocp_lock_fuses: OCP LOCK HEK fuse programming
                                          (cfg ocp-lock)
- pre/post_stable_owner_key_derivation: stable owner key derivation
                                          (cfg stable-owner-key)
- pre/post_encrypted_firmware_decrypt: CM_AES_GCM_DECRYPT_DMA on the
                                          encrypted-boot path
                                          (cfg core_test + recovery wire)

All three hook pairs live on code paths that are dead-code-eliminated when their corresponding feature is off, so they contribute zero bytes to ROMs that don't enable the feature:

- OCP LOCK and stable-owner-key hooks are inside `#[cfg(feature = "...")]` blocks.
- Encrypted FW hooks are inside the `if encrypted_boot { ... }` block; and a runtime wire check, so when `core_test` is off the whole block (including the hook calls) is DCE'd.

The hooks are wired in:
- `Soc::populate_fuses` call site in `cold_boot.rs` (around the OCP LOCK fuse block, gated by `cfg(feature = "ocp-lock")`).
- Around `derive_stable_owner_key` call in `cold_boot.rs` (gated by `cfg(feature = "stable-owner-key")`).
- Around `decrypt_firmware` call inside the encrypted boot branch in `cold_boot.rs` (only reachable on the encrypted-firmware path).

All hook methods are default no-ops on the RomHooks trait so existing integrators don't need to update their impls. The example LoggingRomHooks in platforms/{emulator,fpga}/rom record bits 14\u201319 in mci_reg_fw_extended_error_info[0] alongside the existing hooks.

Limitation: hooks for the other 2.1 flows (DOT, FIPS zeroization, I3C services) were intentionally dropped because each `call_hook` site costs ~12-40 bytes of .text (Option check + vtable dispatch) on always-compiled code paths and the `test-usb-ocp-recovery` ROM is already near the 64 KiB ROM size budget. Integrators that want fine-grained DOT/FIPS/I3C observability can still use the existing `McuRomBootStatus` flow checkpoints written to MCI registers.

### DIFF
--- a/platforms/emulator/rom/src/riscv.rs
+++ b/platforms/emulator/rom/src/riscv.rs
@@ -136,6 +136,30 @@ impl RomHooks for LoggingRomHooks {
         caliptra_mcu_romtime::println!("[mcu-rom-hook] post_load_firmware");
         record_hook_bit(13);
     }
+    fn pre_set_ocp_lock_fuses(&self) {
+        caliptra_mcu_romtime::println!("[mcu-rom-hook] pre_set_ocp_lock_fuses");
+        record_hook_bit(14);
+    }
+    fn post_set_ocp_lock_fuses(&self) {
+        caliptra_mcu_romtime::println!("[mcu-rom-hook] post_set_ocp_lock_fuses");
+        record_hook_bit(15);
+    }
+    fn pre_stable_owner_key_derivation(&self) {
+        caliptra_mcu_romtime::println!("[mcu-rom-hook] pre_stable_owner_key_derivation");
+        record_hook_bit(16);
+    }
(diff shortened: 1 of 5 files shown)


mlvisaya (approved):
LGTM

swenson:
Thanks!

Issue #754, opened 2026-01-14
Setup CI for size / usage analysis


swenson:
* ROM instruction size
* Kernel instruction size
* App instruction size
* Kernel Stack size
* App Stack size

PR #1473, opened 2026-05-18
xtask: add `sizes` subcommand and enforce ROM size budgets in precheckin

swenson:
`cargo xtask sizes` builds every ROM variant in `caliptra-mcu-builder::features::ROM_VARIANTS` and prints a table (or CSV with `--csv`) comparing each variant's `.text` against its platform budget.
    
`cargo xtask precheckin` builds all variants too; the linker and `append_rom_digest` together catch any over-budget regression.
    
`sizes`, `precheckin`, and `clippy` accept `--variants` to override the default list. Format: `platform:f1+f2,platform2:f3`.
    
Per-variant clippy and panic-symbol checks are scaffolded but disabled (with TODOs); enabling them is left to a follow-up PR that fixes the pre-existing lints / panic paths they surface.

Related to #754

### DIFF
--- a/xtask/src/clippy.rs
+++ b/xtask/src/clippy.rs
@@ -1,16 +1,27 @@
 // Licensed under the Apache-2.0 license
 
 use anyhow::{bail, Result};
+use caliptra_mcu_builder::features::RomVariant;
 use caliptra_mcu_builder::PROJECT_ROOT;
 use std::process::Command;
 
-pub(crate) fn clippy() -> Result<()> {
+pub(crate) fn clippy(_rom_variants: &[RomVariant]) -> Result<()> {
     clippy_all()?;
+    // TODO: extend coverage with `clippy_rom_variants(_rom_variants)`
+    // once each ROM variant is clippy-clean.
+    // Tracked in https://github.com/chipsalliance/caliptra-mcu-sw/issues/1478
     Ok(())
 }
 
+/// Public entry point for `cargo xtask clippy-rom-variants`, the
+/// on-demand per-variant pass. Lets users (and the TODO follow-up
+/// PR) exercise per-variant clippy without rewiring `precheckin`.
+pub(crate) fn rom_variants(variants: &[RomVariant]) -> Result<()> {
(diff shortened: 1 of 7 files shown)


xtask/src/clippy.rs
```diff
@@ -1,16 +1,21 @@
...
 
 use anyhow::{bail, Result};
+use caliptra_mcu_builder::features::RomVariant;
 use caliptra_mcu_builder::PROJECT_ROOT;
 use std::process::Command;
 
-pub(crate) fn clippy() -> Result<()> {
+pub(crate) fn clippy(_rom_variants: &[RomVariant]) -> Result<()> {
     clippy_all()?;
+    // TODO: extend coverage with `clippy_rom_variants(_rom_variants)`
```
ternst-nv, line 10: [nit] Could we file a P1 issue for this for the 2.1 ROM?  While it isn't necessary it would be good to burn those down in time for a full release.

xtask/src/clippy.rs
```diff
@@ -20,7 +25,47 @@ fn clippy_all() -> Result<()> {
...
+    }
+    Ok(())
+}
+
+/// Run clippy against every ROM variant so feature-gated code paths
+/// in `mcu-rom-{emulator,fpga}` are linted with their actual feature
+/// set (the workspace clippy pass only exercises default features).
+///
+/// Not wired into `clippy()` by default — see the TODO there.
+#[allow(dead_code)]
```
ternst-nv: Could we invoke a subset of our targets for this?  Or are they currently all broken?

ternst-nv (approved):
LGTM with a couple nits

xtask/src/clippy.rs
swenson: Added https://github.com/chipsalliance/caliptra-mcu-sw/issues/1478

xtask/src/clippy.rs
swenson: Added it as an option, but I think they are all currently failing the clippy check.

swenson:
Thanks!

PR #1475, opened 2026-05-19
fpga ci: surface test failures and skip Dev-init OTP wedgers

swenson:
run_test_suite was silently discarding the result of the remote nextest
invocation, so missing test archives, extraction errors, and nextest
non-zero exits never failed the CI step. Capture and return the result
after copying the JUnit log.

Also wire xtask up to the mounted squashfs CI publishes (TEST_BIN /
CPTRA_TEST_BIN_DIR), let CI override the cd target via
CPTRA_TEST_WORKSPACE, build the test archive with --release, and emit
--status-level=all so a hung test is visible in the streamed log.

Once tests actually ran, constructing the model with
LifecycleControllerState::Dev wedged the FPGA at "Taking subsystem out
of reset" (reproduced locally with test_lc_dev_to_prod). Disable the
Dev-init tests in nightly-ci until the bitstream / hw-model regression
is root-caused.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -22,23 +22,20 @@ store-failure-output = true
 
 [profile.nightly-ci]
 inherits = "nightly"
+# Tests that program OTP with LifecycleControllerState::Dev as the initial
+# state hang the FPGA at "Taking subsystem out of reset" (confirmed locally
+# with test_lc_dev_to_prod). Dev-init tests are disabled until that's
+# root-caused.
 default-filter = """
 (package(caliptra-mcu-hw-model)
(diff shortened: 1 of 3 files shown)


helloxiling (approved):
Looks good to me.

clundin25 (approved):

clundin25:
@swenson I think the intent of using xtask was to make sure the flows continue to work. That would be a good invariant to keep, IMO

swenson:
@clundin25 yeah, I'll try to keep that. I'm currently debugging why this is failing. I'll ping you when this is ready for a re-review.

PR #1480, opened 2026-05-19
Wire up MCU mailbox 1 interrupts and fix MCI global interrupt enable gating

chrisherczeg:
Wire up MCU mailbox 1 interrupts and fix MCI global interrupt enable gating

Two related fixes that surfaced when running a SOC<->MCU mailbox test that
exercises both MCU mailboxes via interrupts:

1. `McuMailbox0Internal` was being reused for the mailbox 1 instance, but
   `IrqEventToMcu` only had `Mbox0*` variants and `Mci::poll` only drained
   IRQ events from `mcu_mailbox0`. As a result, the mailbox 1 instance
   never raised an interrupt to the MCU, even though its register accesses
   were already plumbed.

   - Add `Mbox1CmdAvailable` / `Mbox1TargetDone` variants to `IrqEventToMcu`.
   - Add `mbox_index: u8` on `MciMailboxImpl` (plus `set_mbox_index`) so the
     same implementation type can act as either mailbox 0 or mailbox 1.
   - In `Mci::new`, tag the `mcu_mailbox1` instance with index 1.
   - In `Mci::poll`, drain IRQ events from `mcu_mailbox1` and map them to
     the `NotifMbox1{CmdAvailSts,TargetDoneSts}` bits (gated by their
     enables), mirroring the existing mailbox 0 path.

2. `update_mci_irq` gated the entire MCI interrupt line on bit 0 of
   `intr_block_rf_global_intr_en_r`. Per the generated register definition
   bit 0 is `ErrorEn` and bit 1 is `NotifEn`; firmware that enables only
   notifications never saw an interrupt because the helper short-circuited
   on the wrong bit.

   - Extract `compute_mci_irq_level` as a pure helper so the gating logic
     is testable in isolation.
   - Gate error0 events on `ErrorEn` (bit 0) and notif0 events on
     `NotifEn` (bit 1), independently.

3. Convert the `panic!` calls in the mailbox CSR write paths (SRAM,
   target user, target user valid, execute) into warn-and-ignore behavior.
   These writes are gated by `is_locked()`; a buggy firmware write should
   not abort the entire emulator process, it should be ignored the way
   real RTL would drop a write whose enable is deasserted.

Tests added:

- `test_mailbox1_interrupt_handling` mirrors the existing mailbox 0
  interrupt test for mailbox 1, asserting that the matching mailbox 0
  status bit stays clear (catches future cross-wiring regressions) and
  that `Mci::new` correctly tags `mcu_mailbox1` with index 1.
- `test_compute_mci_irq_level_gating` covers `ErrorEn` / `NotifEn`
  independence, including the specific regression of enabling only
  `ErrorEn` while a notification is pending.

### DIFF
--- a/emulator/periph/src/mcu_mbox0.rs
+++ b/emulator/periph/src/mcu_mbox0.rs
@@ -38,8 +38,12 @@ pub struct McuMailbox0Internal {
 
 impl McuMailbox0Internal {
     pub fn new(clock: &Clock) -> Self {
+        Self::new_with_mbox_index(clock, 0)
+    }
+
+    pub fn new_with_mbox_index(clock: &Clock, mbox_index: u8) -> Self {
         Self {
-            regs: Arc::new(Mutex::new(MciMailboxImpl::new(clock))),
+            regs: Arc::new(Mutex::new(MciMailboxImpl::new(clock, mbox_index))),
         }
     }
 
@@ -179,7 +183,7 @@ impl MciMailboxImpl {
     const CMD_STATUS_VAL: u32 = 0x0;
     const HW_STATUS_VAL: u32 = 0x0;
 
-    pub fn new(clock: &Clock) -> Self {
+    pub fn new(clock: &Clock, mbox_index: u8) -> Self {
         Self {
             sram: MciMailboxRam::new(),
             lock: ReadOnlyRegister::new(Self::LOCK_VAL),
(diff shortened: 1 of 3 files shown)


emulator/periph/src/mcu_mbox0.rs
```diff
@@ -192,9 +198,17 @@ impl MciMailboxImpl {
...
             max_dlen_in_lock_session: 0,
             test_mcu_mbox_driver: false,
+            mbox_index: 0,
         }
     }
 
+    /// Set which physical mailbox (0 or 1) this instance represents. The MCI
+    /// uses this to select the correct IRQ event variant when forwarding
+    /// CMD_AVAILABLE / TARGET_DONE notifications to the MCU.
+    pub fn set_mbox_index(&mut self, idx: u8) {
```
swenson: Why don't we just add this as a parameter to the constructor rather than setting it after making it? Seems like something we don't want to change after we make the mailbox.

swenson (approved):
LGTM; one thing I would change

swenson (approved):

PR #1481, opened 2026-05-19
[bare-metal] move away from runtime paradigm to support FPGA

### LINKED ISSUE #1446: [provisioning] enable bare metal binaries on FPGA
requires some enhancements to the bare metal binary code to pull things out of the binary config and into the firmware bundler

Also, we want to be able to build binaries outside of the template one in `runtime/bare-metal`. need to figure out exactly how that will happen

amit-kumarh:
The bare-metal binaries are no longer treated as mini runtime firmwares that are launched by the ROM into standalone binaries that are included in all_fw.zip to better match the provisioning use case of sideloading into SRAM over JTAG.


amit-kumarh:
the next follow on will allow for more easily building binaries aside from just the template in `runtime/bare-metal`, since we'll likely need at least 3 binaries for provisioning

runtime/bare-metal/src/main.rs
```diff
@@ -6,38 +6,50 @@
...
-#[no_mangle]
-pub extern "C" fn main() {
-    // Write the message to the console.
-    const UART0: *mut u8 = 0x1000_1041 as *mut u8;
-    unsafe {
-        for byte in MSG {
-            core::ptr::write_volatile(UART0, *byte);
+    unsafe fn write_byte(byte: u8) {
+        if cfg!(feature = "fpga") {
+            core::ptr::write_volatile(FPGA_UART_OUTPUT, byte as u32 | 0x100);
```
zhalvorsen: Can we use something like `platforms/fpga/rom/src/io.rs` to let each platform set these values without it living in this file directly? This makes it easier for an integrator to do these operations as they will also be platform specific

tests/integration/src/jtag/test_bare_metal_jtag.rs
```diff
@@ -0,0 +1,100 @@
...
+        // Pull bare-metal bytes from prebuilt bundle environment.
+        let binaries = FirmwareBinaries::from_env().expect("Firmware bundle not found");
+        let bare_metal_bytes = &binaries.mcu_bare_metal;
+        assert!(
+            !bare_metal_bytes.is_empty(),
+            "mcu_bare_metal binary is empty"
+        );
+
+        // Sideload and execute bare metal binary
+        sideload_bare_metal(&mut *mcu_tap, bare_metal_bytes)
```
zhalvorsen, line 74: Does the MCU TAP have access to this region on cold reset? https://github.com/chipsalliance/caliptra-ss/blob/main/docs/CaliptraSSIntegrationSpecification.md#mcu-fw-boot-update makes it sound like Caliptra has access to the region until it sets `FW_EXEC_CTRL[2]`. This might be masked because we have all AXI users set to the same value in the hw-model currently

amit-kumarh: I believe this works because debug is enabled due to the LC state (the manuf and prod debug unlock tests use the same tap as well) - and we need debug mode to be able to halt and rewrite the PC.

```sv
// When debug enabled allow full access to the SRAM
assign axi_debug_req_qual = cif_resp_if.dv & debug_en;
...
    if (exec_region_req) begin
        if (fw_exec_region_mcu_access) begin
            exec_region_filter_success = (axi_mcu_lsu_req | axi_mcu_ifu_req | axi_debug_req_qual);
            exec_region_filter_error   = ~exec_region_filter_success;
        end 
        else begin
            exec_region_filter_success = axi_mcu_sram_config_req | axi_debug_req_qual;
            exec_region_filter_error   = ~exec_region_filter_success;
        end
    end
```

in all SRAM regions debug mode grants access.

confirmed locally that even with different IDs we are still able to write to the SRAM.

zhalvorsen: Okay, cool. Thanks for checking!

runtime/bare-metal/src/main.rs
amit-kumarh: Good idea - pulled that logic out to a separate crate for now, which can be swapped out (or have a module added) for another platform. At a later time, perhaps we can revisit merging it with the ROM IO crates? But I didn't want to touch that boot flow at the moment.

zhalvorsen: Maybe it would be good to add an issue to track making this more generic so integrators can use it more directly. As-is I'm not sure how an integrator would replace that module with their platform's implementation

zhalvorsen (approved):

runtime/bare-metal/src/main.rs
amit-kumarh: Right now, the thinking is that caliptra-mcu-bare-metal-io can be swapped out/modified itself (which seems to be how an integrator would also create a ROM for their platform?). I can file an issue to make this more generic with traits and such - I think the question would then be do we want to replace the IO implementations in the platform ROM and runtimes with this.

zhalvorsen: I think the difference is that the ROM does these steps and then from their `main()` calls into the `rom-common` crate to do all of the generic stuff. Is that to model you are planning to go with as well?

amit-kumarh: Ah ok, I see what you're saying now. Filed an issue for it at https://github.com/chipsalliance/caliptra-mcu-sw/issues/1508 and I'll have a deeper think on how to ensure we can build for out of tree platforms; not sure if that same model is the best one here.

zhalvorsen: Sounds good, thanks!

PR #1482, opened 2026-05-19
[idevid provisioning] Use flash as the external otp storage

mlvisaya:
The external OTP provides the interface to access idevid certificate and
signatures.

For testing purposes, the OTP storage is currently using the staging
SRAM. However, FPGA doesn't have any mmio exposed in the linux kernel
module to directly access the staging sram and initialize it with the
content of the idevid certificate. The staging SRAM also loses its
contents on reset which we don't want to happen when we do a reset
during the test.

Due to this reasons, we are moving the implementation of the external
OTP to flash storage. A flash partition is created in the primary flash
controller for this purpose.

Since flash is async we need to propagate async throughout the
external_otp driver as well. The syscall and drivers have been updated
to async.

The integration test needs to run the imaginary_flash_controller_service
to emulate the flash, so it's been added to the tests.


platforms/emulator/runtime/kernel/drivers/external_otp/src/ext_flash_otp.rs
```diff
@@ -0,0 +1,342 @@
...
+
+        let buffer = self.buffer.take().ok_or(ExternalOtpError::Busy)?;
+        // Pre-fill with 0xFF for partial reads (< 4 bytes).
+        buffer[..4].copy_from_slice(&[0xFF; 4]);
+
+        self.state.set(State::Reading);
+        match self.flash.read(buffer, flash_addr, readable) {
+            Ok(()) => Ok(()),
+            Err(_) => {
+                self.state.set(State::Idle);
```
mhatrevi, line 157: buffer seems to be leaked here?

mhatrevi: Please see other places as well for this pattern.

mhatrevi (approved):

mlvisaya:
Addressed the buffer leak feedback: Changed `FlashStorage::read/write` trait to return the buffer on error (`Result<(), (ErrorCode, &'static mut [u8])>`), following standard Tock HIL conventions. Updated `FlashStorageToPages`, `ExtFlashBackedExternalOtp` (all 5 call sites), and `FlashPartition` to properly recover the buffer on all error paths.

PR #1484, opened 2026-05-20
rom: fix clippy warnings and remove panic paths

swenson:
Fix all per-variant clippy lints and eliminate panic symbols from the FPGA flash driver. Enable per-variant clippy and panic-check for all 11 ROM variants in precheckin.

Add `cargo xtask panic-check` subcommand for ad-hoc use.


ternst-nv (approved):
LGTM

swenson:
Thanks!

PR #1485, opened 2026-05-20
[rom] default unprovisioned max_svn to 128

### LINKED ISSUE #1486: [rom] Set default max_svn to 128
Currently, an unprovisioned (0) max_svn means that all images must have an SVN of 0, or boot will fail. Once a non-zero value is provisioned, it would then be unable to be changed. By treating the unprovisioned field as a max SVN of 128, the max SVN can then be restricted later on without needing to provision the field at manufacturing

amit-kumarh:
So that an unprovisioned MAX_SVN isn't unbootable with an SVN > 0 and allows for the max_svn to be limited later in product life cycle

### DIFF
--- a/rom/src/rom.rs
+++ b/rom/src/rom.rs
@@ -301,7 +301,9 @@ impl Soc {
         let word = otp
             .read_u32_at(fuses::OTP_CPTRA_CORE_SOC_MANIFEST_MAX_SVN.byte_offset)
             .unwrap_or_else(|_| fatal_error(McuError::ROM_OTP_READ_ERROR));
-        self.registers.fuse_soc_manifest_max_svn.set(word);
+        self.registers
+            .fuse_soc_manifest_max_svn
+            .set(if word == 0 { 128 } else { word });
 
         // Manuf Debug Unlock Token.
         for i in 0..self.registers.fuse_manuf_dbg_unlock_token.len() {
--- a/tests/integration/src/lib.rs
+++ b/tests/integration/src/lib.rs
@@ -559,6 +559,7 @@ mod test {
         caliptra_mcu_hw_model::new(InitParams {
             fuses: Fuses {
                 fuse_pqc_key_type: params.vendor_pqc_type.map(|t| t as u32).unwrap_or(0),
+                soc_manifest_max_svn: 0,
                 vendor_pk_hash,
                 ..Default::default()
             },
--- a/tests/integration/src/rom/mod.rs
+++ b/tests/integration/src/rom/mod.rs
@@ -7,6 +7,7 @@ mod test_lc_ctrl;
 mod test_otp_blank_check;
 mod test_otp_scramble_check;
 mod test_rom_hooks;
+mod test_svn;
 mod test_sw_digest_lock;
 mod test_vendor_key_policy;
 mod test_warm_reset;
--- a/tests/integration/src/rom/test_svn.rs
+++ b/tests/integration/src/rom/test_svn.rs
@@ -1,0 +1,27 @@
+// Licensed under the Apache-2.0 license
+
+#[cfg(test)]
+mod test {
+    use caliptra_api::SocManager;
+    use caliptra_mcu_hw_model::McuHwModel;
+
+    use crate::test::{start_runtime_hw_model, TestParams};
+
+    #[test]
+    fn test_unprovisioned_max_svn() {
+        let mut model = start_runtime_hw_model(TestParams {
+            rom_only: true,
+            ..Default::default()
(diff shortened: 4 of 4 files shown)


RaunakGu:
@amit-kumarh , why do we need to set max_svn for un-provisioned case,
Isn’t Caliptra already skips svn verification for un-provisioned state?

cc @mhatrevi

amit-kumarh:
Caliptra skips the svn verification when the device lifecycle is in Unprovisioned - in this case, "unprovisioned" refers to that specific fuse field being unprovisioned (0).

However, there's some utility to bypassing specifically the MAX_SVN check in manufacturing as well (say if the program that provisions the  fuse field in manufacturing itself has an SVN > 0). So the (eventual) change proposed in #1509 which supersedes this issue is to set the max_svn that is passed to Calipra if the field is unprovisioned and we are in a non-prod state.

RaunakGu:
@amit-kumarh , Thanks for reply,

Caliptra check for security state to say if un-provisioned and skip the check, what ever we populate from MCU ROM will not make any difference, 

Do you mean Caliptra is going to have the check irrespective of state?

amit-kumarh:
Looking at https://chipsalliance.github.io/caliptra-web/docs/2.1/subsystem/ss_hardware_spec.html#lcc-interpretation-for-caliptra-core-security-states

The SVN check will be skipped when the Core state is Unprovisioned, which corresponds to the LCC TEST_UNLOCKED state. However, this doesn't cover the LCC MANUF state, where we may not yet have provisioned the MAX_SVN field yet. (see https://chipsalliance.github.io/caliptra-web/docs/2.1/mcu/provisioning.html#lifecycle-state-constraints-for-provisioning-fuses, where the SVN partition can be provisioned in the field).

> Do you mean Caliptra is going to have the check irrespective of state?

so there would be no change to Caliptra Core. Additionally, #1509 only covers the MAX_SVN - it doesn't affect the minimum SVN, which the Manifest SVN would still need to be higher than (when the Core security state isn't Unprovisioned, so Caliptra will run the SVN check).

RaunakGu:
@amit-kumarh 
MANUF is not an un-provisioned state, it is a non-debug state by Caliptra design, where the drive is expected to be provisioned. If provisioning is not done, there are larger implications, and even secure boot requirements would fail. Given this, I feel we may be overhandling this case.

SVN_MAX provisioning when LC=MANUF is not blocked by design, it has to be provisioned by Manufacturer any way

image

amit-kumarh:
Some of the fuse fields do have need to be provisioned by certain life cycle stages, but as far as I can tell the SVN partition has no such restrictions. Per the provisioning guide, the only fields that need to be provisioned in TEST_UNLOCKED are the LC transition tokens, so it's plausible that the MAX_SVN may be not yet burned in MANUF (along with many of the other fuse fields)

Additionally, the change as proposed will still respect the MAX_SVN if a value has been burned there. We want to accomodate for a scenario that allows for an image with an SVN > 0 in that state prior to initially setting the MAX_SVN (if said image was, say, responsible for burning that field). We could potentially gate this on debug being unlocked as well, if defaulting the MAX_SVN to nonzero prior to burning it is a concern?

RaunakGu:
Hi @amit-kumarh, 

I understand the intent,

But **MANUF** is **non-debug** state, and non-debug **SHALL fail** if right value is not provisioned, that is per definition of non-debug state. We shall not provide a bypass in non-debug security states.

**MANUF** = Non-debug is very close to PROD state, except PROD has some additional restriction, I would recommend not to equate MANUF with un-provisioned state.

amit-kumarh:
Yes - so what if in addition to being in a nonprod lifecycle, we gate on debug being unlocked as well? this would still allow the provisioning scenario above without compromising in a nondebug state

RaunakGu:
hi @amit-kumarh, It seems you may be missing the explanation.

Let me add more points.

we already have a dedicated lifecycle state — MANUF-DEBUG-UNLOCK.
If we start introducing special handling for fields that are not provisioned in MANUF, it sets a precedent for similar changes across multiple fields. MAX_SVN is not unique in this regard, as there are several other fields with similar conditions.
Extending such behaviour broadly would effectively make the MANUF state behave like a debug state, which goes against the intended lifecycle design.

Do you also plan to introduce default values for other eFuse registers in the MANUF state?

amit-kumarh:
Ah, thanks for the additional details, I see the rationale now.

Looking at the rest of the prerequisites to booting firmware, this anyway wouldn't enable an in-band fuse provisioning scenario, in which case we might as well leave it as it is.

PR #1491, opened 2026-05-21
docs: add vendor PK hash management firmware header proposal

swenson:
Similar to the SVN header, we want to offer the ability for integrators to manage vendor PK hashes through an authenticated firmware manifest.

### DIFF
--- a/docs/src/vendor-pk-hash-management.md
+++ b/docs/src/vendor-pk-hash-management.md
@@ -1,0 +1,397 @@
+# Vendor PK Hash Management Firmware Header
+
+## Overview
+
+The MCU Vendor PK Hash Management Manifest is an optional header in the
+MCU runtime firmware image that allows a firmware release to provision
+new vendor public key hashes into OTP fuse slots, revoke individual
+signing keys within a slot, or revoke an entire vendor PK hash slot.
+
+Like the [MCU Component SVN Manifest](svn.md#mcu-component-svn-manifest-optional)
+and the [Firmware Manifest DOT Section](firmware_format.md#firmware-manifest-dot-section),
+this header is authenticated as part of the SoC manifest signature
+verification performed by Caliptra Core. MCU ROM processes the manifest
+after the firmware image has been loaded and authenticated, before
+jumping to the runtime.
+
+Three operations are supported:
+
+1. **Provision** — write a new vendor PK hash and PQC key type to an
+   unprovisioned fuse slot.
+2. **Revoke Key** — burn additional bits in a slot's ECC, LMS, or MLDSA
+   revocation bitmask to revoke individual signing keys within that slot.
+3. **Revoke Slot** — burn a bit in `CPTRA_CORE_VENDOR_PK_HASH_VALID` to
+   permanently disable an entire vendor PK hash slot.
+
+All operations are **one-way** (fuses only burn 0→1) and
+**idempotent** (re-executing a command that has already been applied
+is a safe no-op). This makes the manifest safe to leave in place
+across reboots and hitless updates.
+
+## Threat Model
+
+Vendor PK hash management via a firmware header addresses several
+threat scenarios:
+
+- **Key compromise response.** If a code-signing key is compromised,
+  the firmware signer can ship an update that revokes the compromised
+  key (via REVOKE_KEY) or the entire vendor PK hash slot (via
+  REVOKE_SLOT) before the attacker can exploit it to sign malicious
+  firmware. The revocation is permanent once burned.
+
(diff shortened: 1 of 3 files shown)


docs/src/vendor-pk-hash-management.md
```diff
@@ -0,0 +1,397 @@
...
+
+### Command Execution
+
+Commands are executed in entry order. Each command re-reads the
+relevant fuse state from OTP before acting, ensuring idempotency.
+
+#### PROVISION
+
+1. Read `CPTRA_CORE_VENDOR_PK_HASH_VALID` from OTP.
+2. If the slot's bit is set (slot is revoked): fatal error — a
```
amit-kumarh, line 195: Just to raise the question - should it be a fatal error if the PK operation fails due to a malformed manifest, or just log the failure and continue boot

docs/src/vendor-pk-hash-management.md
```diff
@@ -0,0 +1,397 @@
...
+#### REVOKE_SLOT
+
+1. Read `CPTRA_CORE_VENDOR_PK_HASH_VALID` from OTP.
+2. If the slot's bit is already set (already revoked): no-op.
+3. Otherwise: burn the slot's bit to 1.
+4. Read back to verify.
+
+### Interaction with Fuse Population
+
+The ROM processes the vendor PK hash management manifest **before**
```
amit-kumarh, line 233: I don't think this is possible - we need Caliptra up to download the firmware into SRAM, so we can't pass those fuse values to Caliptra on this boot, right?

docs/src/vendor-pk-hash-management.md
```diff
@@ -0,0 +1,397 @@
...
+   operation is always safe to attempt).
+
+Rejecting the bundle at verify time prevents applying a firmware
+image that would cause a fatal error during ROM boot.
+
+## Platform Configuration
+
+### ROM Feature Gate
+
+The vendor PK hash management manifest is gated by:
```
amit-kumarh, line 311: How does this work - we need both the feature enabled and the RomParam set for this to be enabled? Is there a reason we need both? I suppose we could save code space by flagging it out if unused

docs/src/vendor-pk-hash-management.md
```diff
@@ -0,0 +1,397 @@
...
+
+**ECC-protected hash fuses.** Vendor PK hash fuses use ECC protection
+and are write-once. The ROM must read the slot's current data before
+writing and skip the write if data is already present (matching the
+manifest entry). Attempting to write a second time to an ECC-protected
+field — even with the same data — may or may not succeed depending on
+the OTP controller implementation; the ROM avoids this by checking
+first.
+
+**Power-loss safety.** For PROVISION, the hash and key type are
```
amit-kumarh, line 379: This is something I overlooked. I think there might be a bricking risk then if a power loss corrupts a PK hash slot? (not specific to this proposal)

docs/src/vendor-pk-hash-management.md
```diff
@@ -0,0 +1,397 @@
...
+
+Each entry is 60 bytes. Fields unused by a given command must be zero.
+
+| Field | Offset | Size | Description |
+|---|---|---|---|
+| Command | `0x00` | 1 B | Operation type (see [Commands](#commands)) |
+| Slot Index | `0x01` | 1 B | Target vendor PK hash slot (0..15) |
+| PQC Key Type | `0x02` | 1 B | PROVISION only: `1` = MLDSA, `2` = LMS. Zero otherwise. |
+| ECC Revocation | `0x03` | 1 B | REVOKE_KEY only: bits to OR into `CPTRA_CORE_ECC_REVOCATION_{slot}` (low 4 bits). |
+| LMS Revocation | `0x04` | 4 B | REVOKE_KEY only: bits to OR into `CPTRA_CORE_LMS_REVOCATION_{slot}` (up to 32 bits). |
```
amit-kumarh, line 110: nit: were we only allowing 16 LMS slots for Subsystem due to needing redundancy in the 32 bits?

docs/src/vendor-pk-hash-management.md
amit-kumarh: For in-slot revocation, we could perhaps cold-reboot after so the new fuse values are in place.

But for PK Hash slot rotation, how would this work? Would it need to be a two-stage deployment? ex. Device currently has Slot 0 provisioned. One operation (signed with key A) provisions key B. Then another operation (signed with Key B) can revoke slot A if it wishes (and this manifest can contain the B-signed MCU RT FW)

PR #1492, opened 2026-05-21
fix: stabilize FPGA mcu-mailbox and mctp-vdm tests

helloxiling:
Three fixes that unstick `test_mcu_mbox_cmds` and `test_mctp_vdm_cmds` on FPGA.

- **mcu-mbox-capsule**: zero-pad the request tail word so
  `copy_to_slice` doesn't panic when the tail isn't 4-byte aligned.
- **MC_GET_LOG/MC_CLEAR_LOG**: gated off on FPGA — logging-flash uses
  `mcu_mbox0`, which conflicts with the host's inbound request.
- **MCTP VDM test**: sync with the user-space responder before the
  first request, spawn `ImaginaryFlashController`, and pre-seed its
  flash file with the `LOGGING_PARTITION` fixture (FPGA flash path
  routes via `mcu_mbox0`, not emulator's `primary_flash`).

## Validation

`vck190-subsystem` self-hosted runner, both tests pass:
- `test_mcu_mbox_cmds`: 266.5s
- `test_mctp_vdm_cmds`: 444.6s

https://github.com/chipsalliance/caliptra-mcu-sw/actions/runs/26235166896

### DIFF
--- a/runtime/kernel/capsules/src/mailbox.rs
+++ b/runtime/kernel/capsules/src/mailbox.rs
@@ -199,13 +199,17 @@ impl<'a, A: Alarm<'a>> Mailbox<'a, A> {
         self.current_app.set(processid);
         self.state.set(MailboxState::Executing);
 
-        // App buffer contains the full payload
+        // App buffer contains the full payload. The mailbox is 32-bit wide and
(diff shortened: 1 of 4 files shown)


mlvisaya (approved):
LGTM~

PR #1495, opened 2026-05-21
ocp: negative tests

ternst-nv:
Add a collection of negative tests which validate the OCP Recovery library.


nquarton (approved):
Looks good to me but you'll want to have someone else with MCU SW experience review as well.

swenson (approved):
LGTM

PR #1496, opened 2026-05-21
Implement FE provisioning commands in caliptra-util-host lib

mlvisaya:
Field Entropy (FE_PROG) is an authorized command that programs device-unique entropy into a specific OTP partition. Because it mutates fuses, the protocol requires cryptographic authorization to prevent unauthorized writes.

The provisioning flow works as follows:

1. Get Auth Challenge — The requester sends a GetAuthCmdChallenge command to the device. The device generates a random 32-byte challenge nonce and stores it internally via the CommandAuthorizer.
2. Sign the Challenge — The requester computes an HMAC-SHA384 over the challenge and command-specific payload (command ID + parameters) using a pre-shared key. This produces a MAC that proves the requester is authorized to execute the command.
3. Send FE_PROG with MAC — The requester sends the ProgramFieldEntropy command along with the partition index and the computed MAC. The device verifies the MAC against the stored challenge using CommandAuthorizer::verify_mac. If verification succeeds, the device programs the field entropy into the specified OTP partition. If it fails, the command is rejected.

Changes:
- Add caliptra-mcu-auth-challenge-signer crate with CommandAuthChallengeSigner trait. This will be used to sign the challenge. This is integrator specific.
- Add FE_PROG (Field Entropy Programming) and GetAuthChallenge command support to caliptra-util-host: command-types, commands, transport (mailbox + SPDM VDM)
- Update SPDM VDM validator to use CommandAuthorizer trait for FE_PROG auth
- Update Mailbox validator to use CommandAuthorizer trait for FE_PROG auth
- Add FeProg config with auth_key to SPDM VDM test-config.toml


caliptra-util-host/apps/mailbox/client/src/lib.rs
```diff
@@ -17,6 +17,11 @@ pub use caliptra_mcu_debug_unlock_signer::{
     DebugUnlockKeys, DebugUnlockSigner, LocalDebugUnlockSigner,
 };
```
mhatrevi, line 18:
```markdown
I’m requesting changes because there are a few correctness/security issues that should be addressed before merge.

## Blockers

1. **SPDM FE_PROG flow appears out of sync with the runtime responder**
   - The host-side SPDM VDM protocol adds `GetAuthCmdChallenge = 0x12`, but the runtime SPDM VDM responder currently appears to define/accept commands only through `DeviceOwnershipTransfer = 0x11`.
   - Since the SPDM FE_PROG validator depends on `client.get_auth_challenge()`, this likely fails against the current responder unless responder-side support is added in the same PR or already exists elsewhere.
   - Please either add matching responder-side support for `GetAuthCmdChallenge` and authorized `ProgramFieldEntropy`, or disable/skip the SPDM FE_PROG validation until responder support lands.

2. **FE_PROG MAC input is inconsistent between mailbox and SPDM paths**
   - Mailbox validation signs with the external MCU mailbox command ID `MC_FE_PROG` (`0x4D43_4650`).
   - SPDM validation signs with the internal util-host command ID `CaliptraCommandId::FeProg` (`0x8011`).
   - These produce different HMAC inputs. The runtime authorization path appears to use the external `CommandId`, so one of these paths is likely wrong.
   - Please introduce a single canonical FE_PROG authorization command ID/helper and use it consistently across mailbox and SPDM. Also add a deterministic test for the exact MAC input bytes: `cmd_id(BE) || partition(LE) || challenge`.

3. **FE_PROG request/auth trailer semantics are unclear**
   - The docs describe the MAC as appended after the command body, but `FeProgRequest` models the MAC as a normal struct field.
   - Existing runtime authorization appears to parse the base request (`McuFeProgReq`) first and then read the MAC from bytes appended after that request.
   - Please either separate the command body from the authorization trailer in the host types, or add tests proving the serialized bytes/checksum exactly match the runtime expectation: `McuFeProgReq || mac`.

4. **Committed FE_PROG auth key in default SPDM test config**
   - `caliptra-util-host/apps/spdm/test-config.toml` includes a full hex HMAC key.
   - Even if this is test-only, it looks like provisioning authorization material and should not be in the default config.
   - Please move it to an explicit test-only/sample config, load it via environment/config override, or replace it with a placeholder.

5. **Unconditional mailbox UDP payload logging**
   - `UdpTransportDriver::send_command()` now prints full payloads and responses unconditionally.
   - FE_PROG payloads include the MAC, and other commands may carry sensitive material.
   - Please remove these prints or gate them behind debug/verbose logging, and avoid dumping full payloads for security-sensitive commands.

6. **Dependency source changed to a personal fork**
   - `Cargo.toml` / lockfiles change many Caliptra dependencies from `chipsalliance/caliptra-sw` to `mlvisaya/caliptra-sw`.
   - If this was for development, please revert to the official upstream dependency before merge or point to an upstreamed `chipsalliance/caliptra-sw` revision.

## Additional suggestions

- Mailbox FE_PROG validation currently hardcodes `partition = 0`; consider using config like the SPDM path does.
- If `get_auth_challenge()` fails, the mailbox FE_PROG validator currently treats that as a pass. For an end-to-end FE_PROG validation, this should probably be a skip or fail unless it is explicitly an expected-negative lifecycle test.
- Please add tests for:
  - valid challenge + MAC accepted,
  - missing challenge rejected,
  - wrong MAC rejected,
  - replayed challenge rejected,
  - FE_PROG mailbox serialization/checksum over `partition || mac`.


```

mlvisaya: Thank you for the detailed review! Here is how each blocker has been addressed in the updated commits:

**Blocker 1 — SPDM FE_PROG flow out of sync with runtime responder:**
Resolved in PR #1497. The firmware-side VDM handler now implements a single `AuthorizedCommand = 0x12` (renaming/collapsing the old `GetAuthCmdChallenge = 0x12` and `ProgramFieldEntropy = 0x10` entries) with a u32 sub-command dispatch:
- `0x4D41_4343` → `get_auth_challenge` handler
- `0x4D43_4650` → `program_field_entropy` handler
The host and firmware are now in sync on both the VDM command code and the sub-command ID encoding. See `runtime/userspace/api/spdm-lib/src/vdm_handler/iana/ocp/caliptra_vdm/`.

**Blocker 2 — FE_PROG MAC input inconsistent between mailbox and SPDM paths:**
Resolved. Introduced `MC_FE_PROG_CANONICAL_CMD_ID = 0x4D43_4650` and `MC_GET_AUTH_CMD_CHALLENGE_CANONICAL_CMD_ID = 0x4D41_4343` in `caliptra-util-host/command-types/src/fuse.rs` as the single source of truth. The SPDM validator now uses `MC_FE_PROG_CANONICAL_CMD_ID` (matching the MCU mailbox `CommandId::MC_FE_PROG` FOURCC). Since sub-command IDs equal the mailbox command IDs, `HMAC(key, cmd_id_be32 || partition_le32 || challenge)` is identical across both transports.

**Blocker 3 — FE_PROG request/auth trailer semantics unclear:**
Resolved by making the wire format explicit in the docs (`caliptra_common_commands.md`): the VDM sub-payload for FE_PROG is `sub_cmd_id(4 LE) || partition(4 LE) || mac(48)` = 56 bytes. The MAC is a normal struct field on the host side; the serialized layout matches what the firmware `McuFeProgReq` parser expects.

**Blocker 4 — Committed FE_PROG auth key in default SPDM test config:**
No change made here. The key in `test-config.toml` is a randomly-generated test-only HMAC key used only by the integration test harness — it has no relation to any production provisioning material. This matches the pattern of other test keys committed elsewhere in the repo (e.g., DPE attestation test vectors). If the project policy changes to disallow committed test keys, this can be addressed separately.

**Blocker 5 — Unconditional mailbox UDP payload logging:**
Resolved. All `println!` calls have been removed from `UdpTransportDriver::send_command()` in `caliptra-util-host/apps/mailbox/client/src/network_driver.rs`.

**Blocker 6 — Dependency source changed to personal fork:**
Resolved. All `caliptra-sw` dependencies in `Cargo.toml` and both lockfiles have been updated to the official `chipsalliance/caliptra-sw@fb13ba3a86a5eca0b26e134e64c5197bdfe7d0c5`. No references to `mlvisaya/caliptra-sw` remain.

---

Regarding the additional suggestions: the `partition = 0` hardcoding and the challenge-failure-as-pass behaviour are noted and can be addressed in a follow-up PR to keep this one focused. The additional test cases (valid MAC accepted, wrong MAC rejected, replay rejected, mailbox serialisation checksum) are also good additions for a follow-up.

mhatrevi (approved):

PR #1497, opened 2026-05-21
Add MCU handler for GetAuthChallenge and FeProg Caliptra Commands

mlvisaya:
Add the MCU mailbox and SPDM VDM Handlers for GetAuthChallenge and FeProg.
Since these 2 transports shares the CommandAuthorizer, the CommandAuthorizer will be passed to the VDM Handler as well.


mhatrevi:
1. `MC_FUSE_REVOKE_VENDOR_PK_HASH` appears to have been accidentally dropped from `MockCommandAuthorizer::is_authorized`, even though `CmdInterface` still routes that command through `handle_authorized_command`. Please restore the `FuseRevokeVendorPkHashReq` import and match arm:
   ```rust
   CommandId::MC_FUSE_REVOKE_VENDOR_PK_HASH => size_of::<FuseRevokeVendorPkHashReq>(),
   ```

2. `test_fe_prog_debug_bytes` currently only prints debug output and always returns `Ok(())`, so CI can pass even if FE_PROG is broken. Since FE_PROG is central to this PR, please make this a real regression test by asserting the expected outcome, e.g.:
   ```rust
   let result = execute_authorized_req(&mut hw, cmd);
   assert!(result.is_ok(), "FE_PROG authorized request failed: {result:?}");
   ```
   If failure is expected for this lifecycle/configuration, please assert that specific expected failure instead.

3. Please remove or gate the new debug prints in `tests/integration/src/runtime/mod.rs`. The helper currently prints the full command, challenge, HMAC key, and computed MAC. Even though this is test code, dumping key/MAC material into normal test logs is risky and noisy.

4. The `program_field_entropy` trait doc says that over VDM, SPDM session authentication is sufficient and no MCU-level HMAC authorization is required. But the new VDM FE_PROG handler does call `CommandAuthorizer::verify_mac` before invoking `program_field_entropy`. Please update the doc comment to match the implementation, for example: VDM dispatch verifies authorization before invoking this transport-agnostic handler.

5. Minor: the protocol test comment says ProgramFieldEntropy is dispatched as sub-command `0x02`, but the implementation uses `0x4D43_4650` / `MCFP`. Please update the comment to avoid confusion.

mhatrevi (changes requested):
Please see the feedback.

mlvisaya:
@mhatrevi Thanks for the review! All 5 items addressed in the latest push:

1. **Restore `MC_FUSE_REVOKE_VENDOR_PK_HASH`** — Added `FuseRevokeVendorPkHashReq` import and match arm in `MockCommandAuthorizer::is_authorized` so it matches the `CmdInterface` routing.

2. **Make `test_fe_prog_debug_bytes` a real regression test** — Renamed to `test_fe_prog_authorized_req` and replaced debug prints with `assert!` on both the SVN baseline and FE_PROG results.

3. **Remove debug prints** — Removed all `println!` from `sign_auth_cmd_challenge` and `execute_authorized_req` in `tests/integration/src/runtime/mod.rs` that were dumping HMAC key, MAC, and command material.

4. **Update `program_field_entropy` doc comment** — Changed to: "Over both the MCU mailbox and VDM paths, the dispatch layer verifies HMAC authorization via `CommandAuthorizer::verify_mac` before invoking this transport-agnostic handler."

5. **Fix protocol test comment** — Updated from "sub-command 0x02" to "sub-command 0x4D43_4650 (MCFP)" to match the actual implementation.

mhatrevi (approved):

PR #1503, opened 2026-05-22
logging: Add multi-instance logging flash support

xuanq-mrvl:
`LoggingSyscall` is currently hardcoded to a fixed `driver_num`, this PR makes it support multiple instances for the use case of crash dump etc.

Note that `LoggingSyscall::new` is a breaking change, I was following mbox and flash API pattern for `new(driver_num: u32)` and `default()`. Also, to have a test for the multi-instance support, an additional logging partition is added to emulator. Alternatively, I suppose there can be a feature to toggle it too. I'm open for options.


xuanq-mrvl:
Resolves #1564

platforms/emulator/config/src/flash.rs
```diff
@@ -191,9 +191,55 @@ impl StandAloneChecksumCalculator {
...
     size: 32 * 1024,
     driver_num: 0x9001_0000,
 };
+
+pub const LOGGING_PARTITION_1: FlashPartition = FlashPartition {
+    name: "logging1",
+    offset: 0x03FF_0000,
+    size: 32 * 1024,
+    driver_num: 0x9001_0001,
+};
```
helloxiling: **Feature-gate this second partition.** `caliptra-mcu-sw` is positioned as the generic SDK reference that all integrators consume. Hardcoding a second logging partition here reserves 32 KB of flash at `0x03FF_0000` for *every* downstream, even those that have no crash-log use case, and it freezes the generic emulator flash map into a two-partition layout that's hard to undo later.

Suggest:
```rust
#[cfg(feature = "crash-log")]
pub const CRASH_LOG_PARTITION: FlashPartition = FlashPartition {
    name: "crash_log",
    offset: 0x03FF_0000,
    size: 32 * 1024,
    driver_num: 0x9001_0001,
};
```
and declare the feature in `platforms/emulator/config/Cargo.toml` (precedent: `active-i3c1`). Also prefer purpose-based naming (`CRASH_LOG_PARTITION`) over `LOGGING_PARTITION_1` so the intent is obvious to anyone reading the flash map.

helloxiling: Nit: could we put this second partition behind a Cargo feature (e.g. `crash-log`)? Since this crate is the generic SDK reference, reserving 32 KB at `0x03FF_0000` would land in every integrator's build by default.

Also, would a purpose-based name like `CRASH_LOG_PARTITION` read more clearly than `LOGGING_PARTITION_1`? Happy to discuss.

platforms/emulator/config/src/flash.rs
```diff
@@ -191,9 +191,55 @@ impl StandAloneChecksumCalculator {
...
+
+#[macro_export]
+macro_rules! logging_flash_list {
+    ($macro:ident) => {{
+        $macro!(0, logging0, LOGGING_PARTITION_0);
+        $macro!(1, logging1, LOGGING_PARTITION_1);
```
helloxiling: If we gate the partition above, this entry would need the matching `#[cfg(feature = "crash-log")]` so `LOGGING_FLASH_INSTANCE_COUNT` collapses to `1` by default.

platforms/emulator/config/src/flash.rs
```diff
@@ -191,9 +191,55 @@ impl StandAloneChecksumCalculator {
...
+            count += 1;
+        };
+    }
+    crate::logging_flash_list!(__count_logging_flash_entry);
+    count
+};
+
+pub const LOGGING_FLASH_DRIVER_NUM_START: usize = LOGGING_PARTITION_0.driver_num as usize;
+pub const LOGGING_FLASH_DRIVER_NUM_END: usize =
+    LOGGING_FLASH_DRIVER_NUM_START + LOGGING_FLASH_INSTANCE_COUNT - 1;
```
helloxiling: Minor: could we drop the `_START` / `_END` pair? `LOGGING_FLASH_DRIVER_NUMS` already lists the configured driver numbers, and relying on just that avoids an implicit "must be consecutive" constraint for downstream integrators who may want their own numbering scheme.

platforms/emulator/runtime/src/board.rs
```diff
@@ -201,8 +205,16 @@ impl SyscallDriverLookup for VeeR {
...
+                ..=caliptra_mcu_config_emulator::flash::LOGGING_FLASH_DRIVER_NUM_END => {
+                for instance in &self.logging_flash {
+                    if let Some(drv) = instance {
+                        if drv.get_driver_num() == driver_num {
+                            return f(Some(*drv));
+                        }
+                    }
+                }
+                f(None)
             }
```
helloxiling: Pairs with the flash.rs note above — if we drop the range, this arm could match by membership in `LOGGING_FLASH_DRIVER_NUMS` instead, e.g.:
```rust
n if caliptra_mcu_config_emulator::flash::LOGGING_FLASH_DRIVER_NUMS
        .iter()
        .any(|d| *d as usize == n) => { ... }
```
That also keeps the dispatch working for non-contiguous driver numbers.

runtime/kernel/capsules/src/logging/driver.rs
```diff
@@ -101,12 +103,17 @@ impl<'a> LoggingFlashDriver<'a> {
...
         LoggingFlashDriver {
             driver,
+            driver_num,
             apps: grant,
             buffer: TakeCell::new(buffer),
             current_app: OptionalCell::empty(),
         }
     }
 
+    pub fn get_driver_num(&self) -> usize {
```
helloxiling, line 113: Consistency nit: surrounding code uses `u32` for driver numbers (`LoggingSyscall::new(u32)`, `command(u32, ...)`, `LOGGING_FLASH: u32`). Would you mind making the field + return `u32` so we don't need `as` casts at the comparison sites?

platforms/emulator/runtime/userspace/apps/example/src/test_logging_flash.rs
```diff
@@ -98,8 +101,90 @@ pub async fn test_logging_flash_various_entries() {
     assert!(read_after_clear.is_err(), "Log should be empty after clear");
 }
 
+pub async fn test_logging_flash_multiple_instances() {
+    if LOGGING_FLASH_INSTANCE_COUNT < 2 {
+        return;
+    }
```
helloxiling: Could we swap the runtime guard for `#[cfg(feature = "crash-log")]` on the function? That way the test (and its imports of `LOGGING_FLASH_DRIVER_NUMS` / `LOGGING_FLASH_INSTANCE_COUNT`) only ship in builds that actually have a second instance.

platforms/emulator/runtime/userspace/apps/example/src/main.rs
```diff
@@ -203,6 +203,7 @@ pub(crate) async fn async_main<S: Syscalls>() {
     {
         test_logging_flash::test_logging_flash_simple().await;
         test_logging_flash::test_logging_flash_various_entries().await;
+        test_logging_flash::test_logging_flash_multiple_instances().await;
```
helloxiling, line 207: If the test above is feature-gated, this call site would need the matching `#[cfg(feature = "crash-log")]`.

runtime/kernel/capsules/src/logging/driver.rs
xuanq-mrvl: This seems to be the convention on kernel capsule side of code, most use `usize` for `driver_num`, like `MboxSram` and `FlashPartition`, whereas userspace side uses `u32`

helloxiling (approved):
Looks good to me. 
Thanks for the contribution!

PR #1506, opened 2026-05-22
fix(mctp): handle unsupported control commands without leaking TX buffer

### LINKED ISSUE #1138: [Bug] MCTP TX command handler stops responding after receiving unsupported control command
When the MCU receives an MCTP control command it does not handle (e.g., GetVendorDefinedMessageSupport command code 0x06), the TX command handler stops responding to all subsequent commands. The MCU must be reset to recover.

  Steps to reproduce:

   1. Establish MCTP communication with the MCU over I3C
   2. Send a valid MCTP Set Endpoint ID command — MCU responds correctly
   3. Send an unsupported MCTP control command (e.g., GetVendorDefinedMessageSupport 0x06)
   4. MCU does not respond — no completion code or error is returned
   5. Send another MCTP control command (e.g., Get Endpoint ID)
   6. MCU does not respond — all subsequent TX commands are dropped

  Root cause: In runtime/kernel/capsules/src/mctp/mux.rs, the control command dispatch match has a default branch that returns Err(ErrorCode::NOSUPPORT). This error propagates up and causes the TX packet buffer to not be returned to
  the pool. Without a TX buffer, no further responses can be sent for any command.

>   _ => return Err(ErrorCode::NOSUPPORT),

  Current behavior:

   - The unsupported command is received and consumed
   - No response is generated
   - The TX packet buffer is leaked (not returned to the available pool)
   - All subsequent commands are dropped because no TX buffer is available

  Expected behavior:

   - Unsupported control commands should return a response with completion code 0x05 (ErrorNotSupportedCmd) per DSP0236
   - The TX packet buffer should be properly returned regardless of whether the command is supported
   - The TX command handler should continue processing subsequent commands

  Environment:

   - Caliptra MCU software emulator with I3C TCP transport
   - MCTP-over-I3C with PEC

mlvisaya:
When the MCU receives an MCTP control command it does not handle (e.g., GetVendorDefinedMessageSupport 0x06), the default match branch returned Err(ErrorCode::NOSUPPORT) from inside the tx_pkt_buffer.take() closure. This caused the taken TX buffer to be dropped without being returned to the pool, permanently blocking all subsequent MCTP TX responses.

Fix:
- Replace the early-return error with a proper Unsupported handler that writes completion code 0x05 (ErrorNotSupportedCmd) per DSP0236 and returns Ok(()), allowing the normal response path to transmit the reply and manage the buffer correctly.
- Set resp_data_len for Unsupported to 1 to account for the completion code byte.
- Add integration tests that send an unsupported command (0x06) and verify the MCU responds with the correct completion code and continues to handle subsequent commands.

Fixes #1138

### DIFF
--- a/emulator/app/src/tests/mctp_ctrl_cmd.rs
+++ b/emulator/app/src/tests/mctp_ctrl_cmd.rs
@@ -37,6 +37,8 @@ pub(crate) enum MCTPCtrlCmdTests {
     GetMctpVersionSupportUnspecified,
     GetMctpVersionSupportUnsupported,
     GetMsgTypeSupport,
+    UnsupportedCmd,
+    GetEIDAfterUnsupported,
 }
 
 impl MCTPCtrlCmdTests {
@@ -85,6 +87,12 @@ impl MCTPCtrlCmdTests {
                 vec![0x01]
             }
             MCTPCtrlCmdTests::GetMsgTypeSupport => {
+                vec![]
+            }
+            MCTPCtrlCmdTests::UnsupportedCmd => {
+                vec![]
+            }
+            MCTPCtrlCmdTests::GetEIDAfterUnsupported => {
                 vec![]
             }
         };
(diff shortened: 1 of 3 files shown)


swenson (approved):
LGTM

PR #1514, opened 2026-05-26
spdm-lite: add vendor-defined request dispatch

mhatrevi:
Add SPDM-lite codec support for VENDOR_DEFINED request/response envelopes and a VdmHandler dispatch hook. The stack now parses the vendor-defined envelope, matches registered handlers by StandardsBodyId and vendor ID, and delegates the VDM payload without copying it.

Keep the default handler table empty so existing users preserve UnsupportedRequest behavior until a concrete handler is registered.


mhatrevi:
Closing this PR as the work will be included in the OCP VDM PR.

PR #1515, opened 2026-05-26
[mcu-mbox] Fixes for rotating vendor PK hashes

embediver:
- Adds an integration test for rotating a vendor PK hash.
For now the test only provisions a PK hash that is immediately revoked.
For a more thorough test an image that ueses a PK hash different from
the default one still needs to be added.
- Fix a few smaller bugs that emerged
  - When checking a fuse is all `0` the Iter `eq` method wasn't used correctly
  - `write_entry_multi` has to be used for `VENDOR_PK_HASH_VALID` since it's encoded raw representation is > 32 bit

### DIFF
--- a/runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
+++ b/runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
@@ -12,8 +12,6 @@ use caliptra_mcu_libsyscall_caliptra::mcu_mbox::MbxCmdStatus;
 use caliptra_mcu_libsyscall_caliptra::otp::Otp;
 use caliptra_mcu_libsyscall_caliptra::{caliptra, otp};
 use caliptra_mcu_libsyscall_caliptra::{mailbox::Mailbox, DefaultSyscalls};
-use caliptra_mcu_libtock_console::{Console, ConsoleWriter};
-use caliptra_mcu_libtock_platform::DefaultConfig;
 use caliptra_mcu_mbox_common::messages::{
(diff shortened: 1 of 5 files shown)


runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -825,6 +828,8 @@ impl<'a> CmdInterface<'a> {
         req: &[u8],
         resp_buf: &'r mut [u8],
     ) -> Result<(&'r mut [u8], MbxCmdStatus), MsgHandlerError> {
+        let mut wr: ConsoleWriter<DefaultSyscalls> = Console::<_, DefaultConfig>::writer();
+        let _ = writeln!(wr, "[mcu-mbox] revoking vendor PK hash...");
```
zhalvorsen: I think we can drop these debug statements

runtime/userspace/syscall/src/otp.rs
```diff
@@ -152,14 +154,24 @@ impl<S: Syscalls> Otp<S> {
         slot: u32,
         new_hash: &[u8; VENDOR_PK_HASH_SIZE],
     ) -> Result<(), ErrorCode> {
+        let mut wr: ConsoleWriter<DefaultSyscalls> = Console::<_, DefaultConfig>::writer();
         if slot > 15 {
+            let _ = writeln!(
```
zhalvorsen: We can drop these debug statements as well

zhalvorsen (approved):

PR #1517, opened 2026-05-26
[fuses] Add vendor fuses tracking field entropy partition status

### LINKED ISSUE #568: MCU should keep track of field entropy partitions that have been burned by burning additional bits to track it
As an option, MCU should keep track of field entropy partitions that have been burned by having an additional set of fuses.

zhalvorsen:
This adds three 32 bit words per field entropy partition to keep track of the state of each slot. One word is used to indicate it has started burning fuses for a given slot. Another indicates the burning has completed. The last indicates the fuses have been zeroized. This allows us to check the state before attempting to burn fuses as well as attest to their state because hardware does not allow ROM or FW to view these fuses.

These are written as words instead of bits because the vendor partition is ECC protected.

This also adds checks to the field entropy programming and zeroizing APIs in ROM.

### DIFF
--- a/hw/fuses.hjson
+++ b/hw/fuses.hjson
@@ -31,6 +31,7 @@
     {"mcu_component_svn_manifest_min_svn": 4},
     {"soc_image_min_svn_0": 4},
     {"soc_image_min_svn_1": 4},
+    {"field_entropy_state": 48},
     {"fpga_test_uds_seed_lo": 32},
     {"fpga_test_uds_seed_hi": 32},
     {"fpga_test_field_entropy": 32},
@@ -652,6 +653,14 @@
         partition: "VENDOR_HASHES_PROD_PARTITION",
         otp_item: "CPTRA_CORE_PQC_KEY_TYPE_15",
         layout: {type: "OneHotLinearOr", duplication: 3},
+    },
+    {
+        name: "field_entropy_state",
+        bits: 384,
+        description: "Status of field entropy slots. 32 bits per flag (word 3N: started, word 3N+1: finished, word 3N+2: zeroized for slot N).",
+        partition: "VENDOR_NON_SECRET_PROD_PARTITION",
+        otp_item: "CPTRA_SS_VENDOR_SPECIFIC_NON_SECRET_FUSE_7",
+        layout: {type: "Single"},
     },
     {
         name: "fpga_test_uds_seed_lo",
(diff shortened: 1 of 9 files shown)


hw/fuses.hjson
```diff
@@ -23,6 +23,7 @@
     {"cptra_itrng_entropy_config_0": 4},
     {"cptra_itrng_entropy_config_1": 4},
     {"stable_owner_key_personalization_seed": 32}, // 256-bit seed for stable owner key derivation
+    {"field_entropy_state": 2},
```
moidx: You probably need to use a partition without ECC protection enabled, otherwise you will only be able to write once to this word. Alternatively, we can calculate a sequence of values that will work A --> B --> C based on the ECC calculations. This will make this harder to port across integrations. Alternatively, we can ask the vendors to set this partition as: `integrity:    false,`

zhalvorsen: This has been expanded to a word per flag which is very inefficient, but we don't have another option at the time. This should be optimized later.

rom/src/cold_boot.rs
```diff
@@ -168,17 +168,60 @@ impl ColdBoot {
...
+
+            let started = otp.is_field_entropy_started(slot).unwrap_or_else(|_| {
+                fatal_error(McuError::ROM_COLD_BOOT_READ_FIELD_ENTROPY_STATE_ERROR);
+            });
+            let finished = otp.is_field_entropy_finished(slot).unwrap_or_else(|_| {
+                fatal_error(McuError::ROM_COLD_BOOT_READ_FIELD_ENTROPY_STATE_ERROR);
+            });
+            let zeroized = otp.is_field_entropy_zeroized(slot).unwrap_or_else(|_| {
+                fatal_error(McuError::ROM_COLD_BOOT_READ_FIELD_ENTROPY_STATE_ERROR);
+            });
```
clundin25: Does it make sense to wrap these states with an enum?

rom/src/cold_boot.rs
```diff
@@ -168,17 +168,60 @@ impl ColdBoot {
...
+            // If slot is fully programmed, skip it and mark it as successful.
+            if finished {
+                let base = McuRomBootStatus::FieldEntropyProgrammingStarted as u16;
+                let partition_status = if partition < 4 {
+                    base + 1 + partition as u16
+                } else {
+                    mci.flow_checkpoint()
+                };
+                mci.set_flow_checkpoint(partition_status);
+                continue;
```
clundin25: Since this is a pretty big function, it might make sense to break the branches into two separate helper functions, eliminating the continue

clundin25 (approved):

rom/src/cold_boot.rs
zhalvorsen: I added a few helper functions that makes this function less complex

rom/src/cold_boot.rs
zhalvorsen: Thats a good idea, I added an enum

PR #1521, opened 2026-05-26
rom: parse and enforce MCU Component SVN Manifest

swenson:
When `svn_manifest_enabled` is set on RomParameters, ROM searches MCU SRAM for the manifest magic after the static MCU image header (and any DOT section), validates the header, reads
MCU_COMPONENT_SVN_MANIFEST_MIN_SVN from OTP (gated by CPTRA_CORE_ANTI_ROLLBACK_DISABLE), rejects rolled-back manifests, burns the fuse up to `header.min_svn` with readback verification, and advances the firmware entry offset past the 1024-byte manifest.

Compile-gated by the new `svn-manifest` feature on rom-common.

The DOT and SVN per-header processors now live in a new `firmware_headers` module behind a single `process_firmware_headers()` entry point that both `FwBoot` and `FwHitlessUpdate` call, so future optional headers can plug in alongside them.

Integration tests in `tests/integration/src/test_svn_manifest.rs` cover magic-absent skip, well-formed parse, bad header, rolled-back reject, fuse boundary, and burn (with OTP readback).

### DIFF
--- a/builder/src/features.rs
+++ b/builder/src/features.rs
@@ -78,6 +78,7 @@ pub const ROM_ONLY_TEST_FEATURES: &[&str] = &[
     "test-i3c-services",
     "test-fw-manifest-dot",
     "test-fw-manifest-dot-hitless",
+    "test-svn-manifest",
     "test-dot-recovery",
     "test-rom-hooks",
 ];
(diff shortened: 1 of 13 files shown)


mlvisaya (approved):
LGTM.

nit, in the future, we could revert https://github.com/chipsalliance/caliptra-mcu-sw/pull/390 in favor of this PR.

swenson:
Thanks!

PR #1530, opened 2026-05-29
spdm-lite: add SET_CERTIFICATE provisioning

ArthurHeymans:
## Summary
- add an OCP device identity provisioning tool and tests
- add spdm-lite SET_CERTIFICATE codec/handler support
- persist managed SET_CERTIFICATE slots in PAL cert storage
- align SET_CERTIFICATE behavior with SPDM 1.2/1.3 capability and algorithm negotiation


parvathib:
The added integration test is a good `SET_CERTIFICATE` smoke test. It currently provisions Owner slot 2 with KeyPairID 1 and verifies that `GET_CERTIFICATE(slot 2)` returns the same DER chain.

 Once attestation/CSR support is available, we should extend this test toward the certificate-slot-management flow:

 | Step | Description | Current status |
 |---|---|---|
 | 1 | Attest with Vendor slot 0 before provisioning | Not covered |
 | 2 | Request an attested CSR for KeyPairID 1, validate the CSR/EAT, and use it to build/issue the Owner cert chain | Not covered |
 | 3 | Provision Owner slot 2 normally with `SET_CERTIFICATE(SlotID: 2, KeyPairID: 1, CertChain)` | Covered |
 | 4 | Verify Owner slot 2 with `GET_CERTIFICATE(SlotID: 2)` and confirm the returned chain matches/checks out | Partially covered — byte-level readback only |
 | 5 | Attest with Owner slot 2 after provisioning | Not covered |

 Also, please build the host requester/SPDM-Utils tooling against the same libspdm static libraries used by the SPDM validator/attestation tests. Concretely:

 1. Build libspdm from the same `caliptra-spdm-emu` checkout used by the validator workflow.
 2. Copy the generated `*.a` files into a known directory.
 3. Set `LIBSPDM_LIB_DIR` to that directory before building/running `ocp_dev_identity_provision_tool`.

 Without this, Caliptra attestation likely will not work, and the provisioning test may use different libspdm behavior than the validator/attestation path

runtime/userspace/api/spdm-lite/codec/src/set_certificate.rs
```diff
@@ -0,0 +1,74 @@
+// Licensed under the Apache-2.0 license
+
+//! SET_CERTIFICATE / SET_CERTIFICATE_RSP wire types (DSP0274 §10.13).
```
parvathib: Nit: Let’s drop the spec section references from these comments. The command/file names already make the SPDM spec context clear, and section references tend to become stale/noisy across spec versions.

runtime/userspace/api/spdm-lite/codec/src/capabilities.rs
```diff
@@ -59,6 +62,18 @@ impl CapFlags {
...
+
+    /// 2-bit `MULTI_KEY_CAP` field value (bits 26..=27).
+    #[inline]
+    pub fn multi_key_field(self) -> u8 {
+        ((self.into_bits() >> 26) & 0b11) as u8
+    }
+
+    /// True only for the exact `MULTI_KEY_CAP = 10b` MultiKeyConnRsp value.
+    #[inline]
+    pub fn multi_key_conn_rsp(self) -> bool {
```
parvathib: This looks incorrect per the spec table for deriving `MULTI_KEY_CONN_RSP`.

 The current helper is used as the connection-level `MULTI_KEY_CONN_RSP` condition in `DIGESTS`, `CERTIFICATE`, and `SET_CERTIFICATE`, but it only handles the `MULTI_KEY_CAP == 10b`case. It does not derive the final state from `(MULTI_KEY_CAP, ResponderMultiKeyConn)` and misses the valid `01b + ResponderMultiKeyConn=1` case, plus the invalid combinations.

 Can we keep a shared helper, but make it derive the actual connection state?

 ```rust
 fn multi_key_conn_rsp(state: &ConnectionState<...>) -> SpdmResult<bool> {
     let cap = state.advertised_cap_flags.multi_key_field();
     let selected = state.other_param_sel.contains(OtherParamSupport::MULTI_KEY_CONN);

     match (cap, selected) {
         (0b00, false) => Ok(false),
         (0b00, true) => Err(SPDM_INVALID_REQUEST),
         (0b01, false) => Err(SPDM_INVALID_REQUEST),
         (0b01, true) => Ok(true),
         (0b10, false) => Ok(false),
         (0b10, true) => Ok(true),
         _ => Err(SPDM_INVALID_REQUEST),
     }
 }

```
Then use this helper from DIGESTS, CERTIFICATE, and SET_CERTIFICATE instead of re-checking the raw capability field at each call site.

File: runtime/userspace/api/spdm-lite/stack/src/capabilities.rs [hunk 3]
```diff
@@ -79,9 +79,19 @@ pub(crate) async fn handle_get_capabilities<'a, Pal: SpdmPal>(
 
     // DataTransferSize == MaxSPDMmsgSize since we don't yet advertise CHUNK.
     let mtu = pal.mtu() as u32;
+    let mut flags = state.cap_flags;
+    if version < SpdmVersion::V13 || !pal.set_certificate_supported() {
```
parvathib: `SET_CERT_CAP` should come from the Caliptra device’s configured local capabilities, not from PAL runtime storage state.

 The PAL should provide `supported_slots()` / `provisioned_slots()` and perform platform policy checks in `write_cert_chain()` / `erase_cert_chain()` or an authorization hook. 

Using`pal.set_certificate_supported()` here to decide whether to advertise `SET_CERT_CAP` is wrong because it makes protocol's capability negotiation depend on runtime storage policy.

`SET_CERT_CAP` advertisement should align with libspdm responder behavior. It should come from the configured local capability, not from PAL runtime storage state. Slot writability and authorization should be checked later in the SET_CERTIFICATE request path.

runtime/userspace/api/spdm-lite/stack/src/set_certificate.rs
```diff
@@ -0,0 +1,1059 @@
...
+
+    build_response(pal, io, state.version, &SetCertificateRsp { slot_id })
+}
+
+fn unsupported_set_certificate() -> SpdmError {
+    SPDM_UNSUPPORTED_REQUEST.with_data(ReqRespCode::SET_CERTIFICATE.0)
+}
+
+fn validate_request_slot(slot_id: u8, supported_slots: u8) -> SpdmResult<()> {
+    if slot_id == 0 || slot_id >= MAX_SLOTS {
```
parvathib, line 122: Slot 0 is read-only on Caliptra, so SET_CERTIFICATE write/erase to slot 0 should be rejected. Since the spec allows SlotID 0..7, add a comment on the top why slot 0 is rejected.

runtime/userspace/api/spdm-lite/stack/src/set_certificate.rs
```diff
@@ -0,0 +1,1059 @@
...
+    req: &SetCertificateReqBody,
+) -> SpdmResult<()> {
+    if state.version < SpdmVersion::V13 {
+        if req.key_pair_id != 0 || req.cert_model() != 0 || req.erase() {
+            return Err(SPDM_INVALID_REQUEST);
+        }
+        return Ok(());
+    }
+
+    if multi_key_conn_rsp(state) {
```
parvathib: `multi_key_conn_rsp(state)` is not deriving the actual `MULTI_KEY_CONN_RSP` state correctly. Check my previous comment.

parvathib:
Can we simplify the SET_CERTIFICATE cert-store API boundary?

 Changes requested:

 1. Remove `set_certificate_supported()` from `SpdmPalCertStore`.
    - `SET_CERT_CAP` should come from configured local capabilities.
    - PAL should not decide whether the device advertises the SET_CERTIFICATE command.

 2. Keep request-time PAL hooks:
    - `set_certificate_authorized(...)`
    - `validate_set_certificate_chain(...)`
    - `write_cert_chain(...)`
    - `erase_cert_chain(...)`

 3. Make request-time hooks fail closed by default.
    - `set_certificate_authorized(...)` should default to `false`.
    - `validate_set_certificate_chain(...)` should default to `NOT_IMPLEMENTED`.
    - If the test implementation wants to authorize requests or use no-op validation, it should override these methods explicitly.

 4. Fix the derived `MULTI_KEY_CONN_RSP` state.
    - The current code computes it incorrectly by checking the raw `MULTI_KEY_CAP == 10b` capability value.
    - It should be derived once from `(MULTI_KEY_CAP, ResponderMultiKeyConn)`.
    - Use that derived value in `SET_CERTIFICATE`, `DIGESTS`, and `CERTIFICATE`.

 This keeps capability advertisement in protocol/local config and platform policy in the PAL/cert-store.

parvathib (changes requested):

ArthurHeymans:
Rebased this branch onto the latest `spdm-lite` and pushed `55a1ec5ed` addressing the latest SET_CERTIFICATE review comments:
- removed `set_certificate_supported()` from the PAL cert-store API; `SET_CERT_CAP` now comes from configured local capabilities
- made SET_CERTIFICATE PAL hooks fail closed by default (authorization false, validation NOT_IMPLEMENTED)\n- added a shared `multi_key_conn_rsp(state) -> SpdmResult<bool>` derived from `(MULTI_KEY_CAP, ResponderMultiKeyConn)` and used it in DIGESTS, CERTIFICATE, and SET_CERTIFICATE
- removed the stale spec-section comments and documented why Caliptra rejects SET_CERTIFICATE for slot 0

parvathib (approved):

PR #1534, opened 2026-05-29
[firmware-update] Add FirmwareUpdateHooks trait

### LINKED ISSUE #1337: [firmware-update] Provide platform hook trait for pre-activation actions
Currently, `FirmwareUpdater::start()` in `runtime/userspace/api/caliptra-api/src/firmware_update/mod.rs` transitions directly from `pldm_wait(State::Activate)` into `set_auth_manifest()` and `update_mcu()`:
```
pldm_client::pldm_wait(State::Activate).await?;

self.set_auth_manifest().await?;
self.update_mcu(&flash_header).await?;
```

There is no mechanism for SoC integrators to perform platform-specific actions before activation sequence. For example, our integration requires quiescing host I/O traffic before resetting components to avoid in-flight transactions corrupting state during the reset.

Would it be possible to provide a trait so that platform implementations can inject
SoC-specific logic before the activation sequence (i.e., before  `set_auth_manifest()`
and `update_mcu()`)?

mlvisaya:
Add an async FirmwareUpdateHooks trait that allows platform integrators to inject SoC-specific logic at key points in the firmware update activation sequence. The trait provides three hooks, each with a default no-op:

- pre_caliptra_activation: called before updating Caliptra firmware
- pre_auth_manifest_activation: called before setting the SoC auth manifest
- pre_mcu_activation: called before updating MCU firmware and activating

The hooks are passed as an optional parameter to FirmwareUpdater::new(). Returning an error from any hook aborts the firmware update.

Add a test hook implementation in the emulator test-firmware-update-flash test that verifies hooks are called in the expected order.

Resolves #1337

### DIFF
--- a/platforms/emulator/runtime/userspace/apps/user/src/firmware_update/mod.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/firmware_update/mod.rs
@@ -23,6 +23,9 @@ use crate::EXECUTOR;
 ))]
 use caliptra_mcu_libapi_caliptra::firmware_update::{FirmwareUpdater, PldmFirmwareDeviceParams};
 
+#[cfg(feature = "test-firmware-update-flash")]
+use caliptra_mcu_libapi_caliptra::firmware_update::FirmwareUpdateHooks;
+
 use caliptra_mcu_libtock_platform::ErrorCode;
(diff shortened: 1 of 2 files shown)


helloxiling (approved):
Looks good to me.

PR #1541, opened 2026-06-01
spdm-lite: add GET_MEASUREMENTS handler

parvathib:
Memory added by this commit:
    user-app: .text +9,984 B  .rodata +2,072 B  .bss +784 B
    SPDM task: .text +9,498 B  .rodata +16 B  .bss +808 B
    kernel: no change

    Current totals:
    kernel: .text 151,840 B  .bss 24,536 B
    user-app: .text 54,642 B  .rodata 12,712 B  .bss 60,864 B

SPDM responder validator results:
    DOE: pass: 607, fail: 0
    MCTP: pass: 607, fail: 0

    group 1 (version): pass: 6, fail: 0
    group 2 (capabilities): pass: 58, fail: 0
    group 3 (algorithms): pass: 70, fail: 0
    group 4 (digests): pass: 15, fail: 0
    group 5 (certificate): pass: 113, fail: 0
    group 6 (challenge_auth): pass: 239, fail: 0
    group 7 (measurements): pass: 106, fail: 0


runtime/userspace/api/spdm-lite/stack/src/measurements.rs
```diff
@@ -0,0 +1,354 @@
...
+        return Err(crate::error::SPDM_VERSION_MISMATCH);
+    }
+
+    // Decode GET_MEASUREMENTS body.
+    let (meas_req, after) =
+        GetMeasurementsReqBody::ref_from_prefix(rest).map_err(|_| SPDM_INVALID_REQUEST)?;
+
+    let signature_requested = meas_req.signature_requested();
+    let _raw_bitstream = meas_req.raw_bitstream_requested();
+    let meas_op = meas_req.measurement_operation;
```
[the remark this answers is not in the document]
parvathib: I don’t think rejecting `RawBitStreamRequested=1` is correct here. DSP0274 says that if the requester asks for raw but the responder cannot return raw, it shall return the corresponding hash; and if the requester asks for hash but a digest is not applicable, it shall return the corresponding raw bit stream.

 So the issue is not that we need to reject the flag. The actual fix is to label the OCP EAT structured manifest correctly as raw (`is_raw = true`), since it is not a digest/hash representation.

runtime/userspace/api/spdm-lite/stack/src/measurements.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+        let mut ctx = [0u8; REQUESTER_CONTEXT_LEN];
+        ctx.copy_from_slice(&after_sig_fields[..REQUESTER_CONTEXT_LEN]);
+        requester_context = Some(ctx);
+    }
+
+    // Measurement enumeration from PAL.
+    let meas_info = pal.measurement_info();
+    let total_count = meas_info.len() as u8;
+
```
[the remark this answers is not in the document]
parvathib: I addressed this by replacing the lossy `as u8` cast with checked conversion. The stack now rejects a PAL measurement list that cannot be represented in the SPDM one-bytecount field, and also rejects count `0xFF` because `0xFF` is reserved as the GET_MEASUREMENTS “all measurements” operation.

 I also added validation that PAL-provided measurement indices are not `0` or `0xFF`.

runtime/userspace/api/spdm-lite/stack/src/challenge.rs
mhatrevi: Are we planning to address this TODO in this PR?

runtime/userspace/api/spdm-lite/stack/src/measurements.rs
```diff
@@ -0,0 +1,369 @@
...
+        after_sig_fields = &after[SPDM_NONCE_LEN + 1..];
+
+        // Slot 0xF (public-key-only measurement signature) is not
+        // supported yet; spdm-lite signs measurements through
+        // provisioned cert-store slots.
+        if slot_id >= MAX_SLOTS || (pal.provisioned_slots() & (1 << slot_id)) == 0 {
+            return Err(SPDM_INVALID_REQUEST);
+        }
+    }
+
```
mhatrevi, line 72: For unsigned GET_MEASUREMENTS requests, should we validate that signature-only fields are absent?

mhatrevi (changes requested):

runtime/userspace/api/spdm-lite/stack/src/challenge.rs
parvathib: Good catch! The placeholder zero MeasurementSummaryHash is removed. CHALLENGE_AUTH now computes the real measurement summary hash. The manifest measurement is marked as TCB.

runtime/userspace/api/spdm-lite/stack/src/measurements.rs
parvathib: Addressed. GET_MEASUREMENTS now computes the expected SPDM request length from SignatureRequested and version. Unsigned requests must not include Nonce/SlotIDParam and signed requests must include them. The handler rejects requests whose SPDM payload length does not match the expected layout.

mhatrevi (approved):

PR #1556, opened 2026-06-03
rom: burn CPTRA_CORE_RUNTIME_SVN from Caliptra FW_INFO

swenson:
Caliptra Core cannot write its own fuses. After Caliptra's runtime mailbox is ready in cold boot and hitless update, MCU ROM issues FW_INFO and advances CPTRA_CORE_RUNTIME_SVN to the reported min_fw_svn (read-back verified). Skipped when
CPTRA_CORE_ANTI_ROLLBACK_DISABLE is set or the floor is unchanged.

- New rom::caliptra_svn module with linear-OR encode/decode helpers for the 128-bit fuse.
- Wired into ColdBoot::run after stash measurement and into FwHitlessUpdate::run after the activate response is released. Adds CaliptraRuntimeSvnBurnComplete boot checkpoint.
- Adds ROM_CALIPTRA_RUNTIME_SVN_BURN_ERROR (0x1_001c).
- docs/src/svn.md: Caliptra Core does not request the burn; MCU ROM drives it via FW_INFO. SoC manifest SVN burn is called out as planned future work.
- 4 unit tests; 2 integration tests covering burn and skip paths.

### DIFF
--- a/docs/src/svn.md
+++ b/docs/src/svn.md
@@ -21,9 +21,9 @@ This document covers four categories of components:
    (the same value Caliptra Core authenticates and binds into the MCU
    Runtime's DPE context). Enforcement and the fuse floor are owned by
    Caliptra Core (`CPTRA_CORE_SOC_MANIFEST_SVN`). MCU ROM, as the OTP
-   writer for the subsystem, performs the actual OTP burn on Caliptra
-   Core's behalf when the authenticated SoC manifest SVN advances; it
-   does not maintain a separate MCU-side floor.
+   writer for the subsystem, performs the actual OTP burn when the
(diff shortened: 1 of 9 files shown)


swenson:
I don't think this works because FW_INFO doesn't report any new min_svn, but just what was already burned. We'll need some other way to determine a new min_svn.

RaunakGu:
Can we potentially handle this in MCU RT itself?
After a successful boot, once execution transitions to MCU RT, if the current SVN (fetched from Caliptra via mailbox) is lower than the provisioned SVN(eFuse Read), we could perform the SVN update at that point.?

PR #1562, opened 2026-06-03
[rom] Add more unique error codes

zhalvorsen:
Previously many of the error paths re-used error codes and distinguished the error using print statements. This added a lot of unnecessary codespace for those strings and formatting logic. This adds more unique error codes and removes the corresponding print statements.

This saves ~2.5KB in ROM.


swenson (approved):
LGTM

PR #1567, opened 2026-06-04
rom: introduce DOT_BLOB_VERSION and DOT_FUSE_ARRAY_BITS constants

### LINKED ISSUE #1566: [Enhancement] Replace magic numbers in device_ownership_transfer.rs with named constants
"device_ownership_transfer.rs" used bare integer literals (1 for the blob version, 256 for fuse array bits) in several places, making the code fragile and opaque to readers.

Enhancement suggested:

Introduce two named constants:

**DOT_BLOB_VERSION (u32 = 1):** the current DOT blob format version, used in DotBlobFields::default() and wherever a blob is constructed. A future format bump becomes a single-line change.
**DOT_FUSE_ARRAY_BITS (u16):** total number of OneHot bits in the dot_fuse_array OTP field, derived at compile time from fuses::DOT_FUSE_ARRAY.byte_size * 8 so it stays in sync with the RDL definition.

giladliber:
Replace magic numbers in device_ownership_transfer.rs with named constants:

- DOT_BLOB_VERSION (u32 = 1): the current DOT blob format version. Used in DotBlobFields::default() and anywhere a blob is constructed, so a future format bump is a single-line change and readers know the value is a version identifier rather than an arbitrary integer.

- DOT_FUSE_ARRAY_BITS (u16): total number of OneHot bits in the dot_fuse_array OTP field, derived at compile time from the generated fuse map (fuses::DOT_FUSE_ARRAY.byte_size * 8) so it stays in sync with the RDL definition instead of being a hardcoded 256.

Additional clean-ups enabled by the constants:
- Redundant new_blob.fields.version = 1 after DotBlobFields::default() in apply_override is removed.
- even_fuses construction in apply_override uses ..dot_fuses.clone() instead of spelling out every field.
- create_and_seal_dot_blob uses ..Default::default() for DotBlobFields partial initialisation, removing the duplicate version and reserved literals.

Fixes #1566

### DIFF
--- a/rom/src/device_ownership_transfer.rs
+++ b/rom/src/device_ownership_transfer.rs
@@ -27,6 +27,16 @@ use zerocopy::{transmute, FromBytes, Immutable, IntoBytes, KnownLayout};
 const DOT_LABEL: &[u8; 23] = b"Caliptra DOT stable key";
 pub const DOT_BLOB_SIZE: usize = core::mem::size_of::<DotBlob>();
 
+/// Current DOT blob format version.
+pub const DOT_BLOB_VERSION: u32 = 1;
+
+/// Total number of bits in the `dot_fuse_array` OTP fuse field.
+///
+/// Derived from the generated fuse map so it stays in sync with the RDL
+/// definition of `DOT_FUSE_ARRAY` (a OneHot field).
+pub const DOT_FUSE_ARRAY_BITS: u16 =
+    (caliptra_mcu_registers_generated::fuses::DOT_FUSE_ARRAY.byte_size * 8) as u16;
+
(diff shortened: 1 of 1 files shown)


rom/src/device_ownership_transfer.rs
mhatrevi, line 96: `load_from_otp` still reads the fuse array into a fixed `[0u8; 32]`, so if `fuses::DOT_FUSE_ARRAY.byte_size` ever changes, `total` would update but the number of bytes counted for `burned` would not.

Could we derive the raw buffer size from `DOT_FUSE_ARRAY.byte_size` as well, or otherwise add an assertion tying the fixed buffer length to the generated fuse size? That would keep the read/count logic synchronized with the same source of truth as `DOT_FUSE_ARRAY_BITS`.

mhatrevi (changes requested):

PR #1578, opened 2026-06-05
Add defmt-based userspace logging infrastructure

helloxiling:
Adds `caliptra-mcu-userlog`: a `defmt` `global_logger` that stages log frames into an SPSC ring and drains them to the flash log capsule for host-side decoding. Format strings stay in the non-loadable `.defmt` section, so the loadable image is unaffected. Gated behind `userspace-log`; an emulator round-trip test runs under `test-defmt-logging`.

## Code size & RAM impact

Measured on the `user-app` binary (riscv32, `devel` profile), comparing baseline vs `userspace-log`:

| Metric | Baseline | With logging | Delta |
|---|--:|--:|--:|
| Flash — code (`.text`) | 151,378 | 154,098 | +2,720 |
| Flash — rodata (`.rodata`) | 23,788 | 24,420 | +632 |
| **Flash — loadable `.bin`** | **175,588** | **178,940** | **+3,352 (+1.9%)** |
| RAM — `.bss` | 49,608 | 51,244 | +1,636 |
| RAM — `.data` | 300 | 300 | 0 |
| RAM — `.stack` reserved | 44,544 | 44,544 | 0 |
| `.defmt` (non-loadable) | 0 | 5 | not in flash |

- **Flash: +3,352 bytes (+1.9%)** for the encoder, framing, ring/logger/drain, and syscall append path.
- **RAM: +1,636 bytes**, mostly `.bss` for the 1 KiB SPSC ring plus staging/logger state. No extra stack reserved.
- `.defmt` format strings cost **0 flash bytes** — the section is non-loadable and excluded from the `.bin`.
- With `userspace-log` **off** (default), `defmt` and `caliptra-mcu-userlog` are optional deps, so non-logging builds are byte-for-byte unchanged.

Follow-up PR: host-side decode in `caliptra-util-host`, FPGA VDM channel, and call-site migration.


mlvisaya (approved):
I think code is well written.

PR #1582, opened 2026-06-05
[xtask] support building multiple bare metal binaries

### LINKED ISSUE #1446: [provisioning] enable bare metal binaries on FPGA

amit-kumarh:
Support building multiple bare metal binaries in one firmware bundle instead of just the template binary.

This will become necessary as we will have multiple provisioning binaries that we need to compile.

Add an integration test for the template bare metal binary and include it in CI (the test is < 10 seconds, so shouldn't bloat time much)

### DIFF
--- a/builder/src/all.rs
+++ b/builder/src/all.rs
@@ -716,13 +716,8 @@ pub fn all_build(args: AllBuildArgs) -> Result<()> {
         ..Default::default()
     })?;
 
-    // Only build the bare-metal runtime for the emulator platform, as it
-    // is currently only configured and supported for the emulator.
-    let mcu_bare_metal_runtime = if platform == "emulator" {
-        Some(crate::bare_metal_build()?)
-    } else {
-        None
-    };
+    let mcu_bare_metal_runtime =
+        crate::bare_metal_build(Some(platform), "caliptra-mcu-bare-metal")?;
 
     let mcu_image_cfg = get_image_cfg_feature(&mcu_cfgs.clone().unwrap_or_default(), "none");
     let mut caliptra_builder = crate::CaliptraBuilder::new(&CaliptraBuildArgs {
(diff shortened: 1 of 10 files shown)


builder/src/all.rs
```diff
@@ -346,13 +344,33 @@ impl FirmwareBinaries {
                 name if name.contains("mcu-test-flash-image") => {
                     binaries.test_flash_images.push((name.to_string(), data));
                 }
+                name if name.starts_with("bare_metal/") => {
+                    let stripped_name = name
+                        .strip_prefix("bare_metal/")
+                        .unwrap()
+                        .strip_suffix(".bin")
+                        .unwrap_or(name.strip_prefix("bare_metal/").unwrap())
+                        .to_string();
```
zhalvorsen: nit: it would probably be cleaner to split these to avoid duplicated code:


```suggestion
                    let stripped_name = name
                        .strip_prefix("bare_metal/")
                        .unwrap();
                    let stripped_name = stripped_name
                        .strip_suffix(".bin")
                        .unwrap_or(stripped_name)
                        .to_string();
```

xtask/src/main.rs
```diff
@@ -116,8 +116,8 @@ enum Commands {
         platform: Option<String>,
 
         /// Build bare metal runtime instead of TockOS runtime
-        #[arg(long, default_value_t = false)]
-        bare_metal: bool,
+        #[arg(long, default_missing_value = "caliptra-mcu-bare-metal", num_args = 0..=1)]
+        bare_metal: Option<String>,
```
zhalvorsen, line 120: Any reason we aren't using an `Option<PathBuf>`? is the `num_args` in preparation for more bare-metals that would be included at the same time with this invocation?

amit-kumarh: The name here is the crate name of the binary, not the path,  hence the `String` instead of `PathBuf` . Re: num args, yes, it's there because the underlying `bare_metal_build` supports only one binary, so we can catch too many arguments at parse time. but eventually, I'll modify it to allow building multiple binaries

builder/src/all.rs
amit-kumarh: Done

zhalvorsen (approved):

PR #1583, opened 2026-06-05
rom: install owner PK hash via mailbox instead of locked register

swenson:
The cptra_owner_pk_hash register is locked once fuse write done is set, so writing it directly dropped the value; send it via INSTALL_OWNER_PK_HASH instead, skipping the all-zero unprovisioned case.


swenson:
Closing in favor of #1584

PR #1584, opened 2026-06-05
rom: install owner PK hash via mailbox and allow forcing the fused owner

swenson:
The cptra_owner_pk_hash register is locked once fuse write done is set, so
writing it directly dropped the value; send it via INSTALL_OWNER_PK_HASH
instead, skipping the all-zero unprovisioned case.
    
Also add an OwnerPkHashPolicy RomParameter (DotThenFuse default / ForceFuse).
ForceFuse bypasses the DOT blob and uses CPTRA_SS_OWNER_PK_HASH, fataling if
that fuse is unprovisioned; the emulator and FPGA ROMs select it from MCI
generic input wire bit 28.

### DIFF
--- a/error/src/lib.rs
+++ b/error/src/lib.rs
@@ -179,6 +179,16 @@ impl McuError {
             "Cold boot ROM integrity check failed"
         ),
         (
+            ROM_DOT_INSTALL_OWNER_PK_HASH_FAILED,
+            0x1_001c,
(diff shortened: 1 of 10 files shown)


helloxiling (approved):
Looks good to me.

mlvisaya (approved):

swenson:
Thanks!

PR #1589, opened 2026-06-08
romtime: use core::hint::spin_loop() for mailbox polling delay

### LINKED ISSUE #1588: [Enhancement] Use ```core::hint::spin_loop()``` for mailbox polling delay
The current delay approach for mailbox polling in "soc_manager.rs"
```rust
fn delay(&mut self) {
    self.counter = core::hint::black_box(self.counter) + 1;
}
```
…uses solely to prevent the compiler from optimizing away the loop. It doesn't convey any intent to the hardware.

```core::hint::spin_loop()``` is the idiomatic Rust replacement for this pattern. It can emit a pause instruction telling the pipeline it's in a spin-wait and compiler won't optimize it away (so no black_box hack needed).
The timeout behavior stays the same since it's controlled by the MAX_WAIT_CYCLES loop counter.

----

**As a side note -** CaliptraSoC implements SocManager but sets the required address constants to dummy zero values and overrides all three accessor methods (soc_ifc(), soc_ifc_trng(), soc_mbox()) with runtime pointers. 
None of the trait's default higher-level methods (mailbox_exec, start_mailbox_exec, finish_mailbox_exec, mailbox_exec_req, init_fuses, setup_mailbox_users) are ever called — the ROM exclusively uses CaliptraSoC's own methods (start_mailbox_req, finish_mailbox_resp, exec_mailbox_req_u32, etc.) which duplicate similar functionality. 
The trait implementation effectively serves only as a way to access the low-level register blocks and delay(), raising the question of whether the new implementation should be added to the default caliptra-sw trait itself (if it doesn't fit our needs).

giladliber:
Replace the black_box counter increment with spin_loop(), which is the idiomatic Rust spin-wait primitive. On RISC-V targets this can emit a pause hint (Zihintpause), reducing power consumption during busy-waits. Also removes the now-unused counter field from CaliptraSoC.

Fixes #1588

### DIFF
--- a/romtime/src/soc_manager.rs
+++ b/romtime/src/soc_manager.rs
@@ -14,7 +14,6 @@ use zerocopy::{FromBytes, IntoBytes};
 const MAILBOX_SIZE: usize = 256 * 1024;
 pub struct CaliptraSoC {
     _private: (), // ensure that this struct cannot be instantiated directly except through new
-    counter: u64,
     soc_ifc_addr: *mut u32,
     soc_ifc_trng_addr: *mut u32,
     soc_mbox_addr: *mut u32,
@@ -39,7 +38,7 @@ impl SocManager for CaliptraSoC {
 
     /// Provides a delay function to be invoked when polling mailbox status.
     fn delay(&mut self) {
-        self.counter = core::hint::black_box(self.counter) + 1;
+        core::hint::spin_loop();
     }
 
     /// A register block that can be used to manipulate the soc_ifc peripheral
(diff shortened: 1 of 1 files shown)


clundin25:
This cleans the assembly up a bit but I want to raise that the VEER core used by MCU doesn't support `Zihintpause ` https://chipsalliance.github.io/Cores-VeeR-EL2/html/main/docs_rendered/html/overview.html.

It seems the `mcpc` CSR should be used for power savings. I'm going to approve and merge the PR because the code is technically more correct, and if you want to add the power savings please send a follow up PR.

clundin25 (approved):

PR #1591, opened 2026-06-08
rom: reset start_cycle and next_print_offset between image transfers

### LINKED ISSUE #1592: [Bug] Recovery transfer progress and throughput reported incorrectly for second image
In ```load_flash_image_to_recovery``` the start_cycle and next_print_offset variables are declared once before the main loop but were never reset when the state machine cycles back to transfer a second image.

This causes two issues:
* Incorrect cycle measurement: the throughput reported for subsequent images includes all the time spent on the previous transfer and intermediate state transitions.
* Missing progress messages: next_print_offset retains the large value accumulated during the first transfer, so progress prints for a subsequent (potentially smaller) image are never triggered.

giladliber:
These variables were not reset when the state machine looped back for a second image transfer, causing incorrect cycle measurements and missing progress messages for subsequent images.

Fixes #1592

### DIFF
--- a/rom/src/recovery.rs
+++ b/rom/src/recovery.rs
@@ -321,6 +321,8 @@ pub fn load_flash_image_to_recovery(
                     state_machine.context_mut().flash_offset = image_info.0;
                     state_machine.context_mut().image_size = image_info.1;
                     state_machine.context_mut().transfer_offset = 0;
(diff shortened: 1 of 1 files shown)


mlvisaya (approved):
looks good to me

PR #1598, opened 2026-06-08
rom: Pass I3C timing data as ROM parameters

### LINKED ISSUE #1063: Question about I3C initialization
Q1. 
Hi, the i3c-core initialization sequence link looks broken now.
In this page : https://chipsalliance.github.io/caliptra-web/docs/2.1/mcu/rom.html#reference-rom-specification
-> https://chipsalliance.github.io/i3c-core/initialization.html

May I have the original document from anyware?

Q2.
https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/rom/src/i3c.rs
around L56~57
`
        // TODO: pass this timing information in
        let clocks = 0;
`
'clocks' is currently zero and it is applied to all timing registers.
should it be modified as per tuning or needs something similar?
please share information about this point if it possible.

thanks.

swenson:
Replace the hard-coded I3C timing register values (clocks=0) with configurable timings supplied through RomParameters. Add an I3cTimings struct whose fields are sized to the 20-bit timing register fields (u32), with defaults that are the recommended settings for high-speed parts.

Bundle the per-controller address, recovery flag, and timings into a single I3cConfig argument to I3c::configure, and expose separate i3c_timings and i3c1_timings RomParameters so the primary and secondary controllers can be tuned individually.

Fixes #1063

### DIFF
--- a/rom/src/cold_boot.rs
+++ b/rom/src/cold_boot.rs
@@ -505,9 +505,17 @@ impl BootFlow for ColdBoot {
         caliptra_mcu_romtime::println!("[mcu-rom] Initializing I3C");
         if straps.active_i3c == 1 {
             caliptra_mcu_romtime::println!("[mcu-rom] Initializing I3C1 (active)");
-            i3c1.configure(straps.i3c1_static_addr, true);
+            i3c1.configure(crate::I3cConfig {
+                static_addr: straps.i3c1_static_addr,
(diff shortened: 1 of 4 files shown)


mlvisaya (approved):
LGTM

swenson:
Thanks!

PR #1603, opened 2026-06-09
tests: Provision otp_memory on FPGA to enable DOT I3C recovery test

### LINKED ISSUE #1507: I3C DOT Recovery flow not tested on FPGA
Currently these are marked as ignore because the fuses are not available in FPGA, but I think we can either work around this or use the fuses in hw/fuses.hjson for this, and get this tested on FPGA.

swenson:
The FPGA hw model dropped InitParams.otp_memory, so DOT fuses (dot_initialized, dot_fuse_array, vendor_recovery_pk_hash) could not be provisioned on FPGA. This forced test_i3c_services_dot_override_full_flow to be ignored on fpga_realtime.

Overlay params.otp_memory onto the OTP backing memory via otp_slice() before boot, OR-ing the bytes since OTP is one-time programmable. The DOT and vendor-secret partitions are zero in the base image, so the standard fuses provisioned by the base model are preserved. Remove the FPGA ignore on the DOT override test.

Fixes #1507


swenson:
I think the root of the problem is that we are using a secret partition for the vendor recovery pk hash, when we should not, as that is not software-accessible.

mhatrevi (approved):

PR #1612, opened 2026-06-10
rom: burn Caliptra-owned SVN floors from the MCU runtime SVN header

### LINKED ISSUE #1421: SVN header implementation
Implementation of the optional SVN header in the runtime FW, which will enforce docs/src/svn.md design.

swenson:
Caliptra Core cannot write its own fuses and FW_INFO only reports the
running SVNs (and never the SoC manifest SVN), so it cannot express a
new minimum. Instead, the authenticated MCU runtime SVN header now
carries the requested floors for CPTRA_CORE_RUNTIME_SVN and
CPTRA_CORE_SOC_MANIFEST_SVN, and MCU ROM (the subsystem OTP writer)
burns them.

- Move the SVN header type/parse/validate into caliptra-mcu-romtime so
  it can be parsed by MCU ROM and later stages (early runtime / FMC).
  rom-common re-exports it for back-compat.
- Extend the header (format_version 1, 1024 B, 126 entries) with
  caliptra_runtime_min_svn + soc_manifest_min_svn (u8 each), validated
  against their fuse ranges.
- Drive both Caliptra-owned burns from process_svn_manifest (runs in
  FwBoot and FwHitlessUpdate, where the header is decrypted and the
  Caliptra mailbox is reachable). Guard the runtime floor with a
  FW_INFO.fw_svn >= caliptra_runtime_min_svn check; the SoC manifest
  floor is burned on trust (FW_INFO does not expose it).
- Remove the prior unconditional min_fw_svn burn from cold_boot and
  fw_hitless_update.
- Add ROM_CALIPTRA_RUNTIME_SVN_BURN_ERROR; update docs/src/svn.md.
- Tests: romtime header unit tests; integration tests for the SoC
  manifest burn, the Caliptra runtime burn (Caliptra FW built at SVN 7
  via a new TestParams.caliptra_svn override), and the runtime
  min-svn rejection.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -53,7 +53,6 @@ default-filter = """
     - test(test_lc_dev_to_scrap)
     - test(test_lc_invalid_transition_error)) |
 (package(caliptra-mcu-tests-integration) and test(test_raw_lifecycle_boot)) |
-(package(caliptra-mcu-tests-integration) and test(test_rom_hooks_fire_in_order)) |
 (package(caliptra-mcu-tests-integration) and test(test_external_otp))
 """
 # Disabled because they hang the FPGA at boot (LifecycleControllerState::Dev):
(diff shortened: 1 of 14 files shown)


mhatrevi (commented):
I think there’s a blocking sequencing issue here: `process_svn_manifest()` can perform irreversible OTP burns before all validation that can fatal has completed.

Currently the order is:

1. `enforce_and_burn_manifest_min_svn(...)`
2. `burn_per_component_min_svn(...)`
3. `burn_caliptra_owned_svns(...)`

But `burn_caliptra_owned_svns()` can still reject the boot after querying `FW_INFO`, e.g. when:

```rust
caliptra_runtime_min_svn > FW_INFO.fw_svn

mhatrevi (approved):

PR #1614, opened 2026-06-11
Use common error codes for MCU MBox and PLDM userspace APIs

mlvisaya:


helloxiling (approved):
Looks good. 
Do we plan to consolidate other user-space libraries to use common error codes?

mlvisaya:
thanks! ill have a follow up PR for the other API modules

PR #1619, opened 2026-06-11
[cherry-picker]: Prebuild xtask before beginning flow

clundin25:
Otherwise the cherry pick process can prevent the xtask binary from building.


swenson (approved):
LGTM

PR #1622, opened 2026-06-12
[fix] Add OTP status offset in SS_STRAP_GENERIC[0] register

### LINKED ISSUE #1608: [Bug] Missing configuration in SS_STRAP_GENERIC[0]
Per the [spec](https://github.com/chipsalliance/caliptra-rtl/blob/main/docs/CaliptraIntegrationSpecification.md#late-binding-interface-signals), the SOC_OTP_CTRL_STATUS register offset should be configured in the lower 16 bits of SS_STRAP_GENERIC[0]:

| 0 | strap_ss_strap_generic_0 | Provides the Caliptra ROM with a 32-bit pointer that encodes the location of the fuse controller's status register and the bit position of the idle indicator. Upper 16 bits: Bit index of the IDLE_BIT_STATUS within SOC_OTP_CTRL_STATUS. **Lower 16 bits: Offset address of SOC_OTP_CTRL_STATUS within the SOC_IFC_REG space, relative to SOC_OTP_CTRL_BASE_ADDR.**
|-- | -- | --|

Currently in the ["rom.rs"](https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/rom/src/rom.rs#L206) code only the upper bits are set.

mhatrevi:


mlvisaya (approved):
LGTM

PR #1623, opened 2026-06-12
spdm-lite: add TDISP VDM support

ArthurHeymans:
## Summary
- add PCI-SIG TDISP handling on top of the spdm-lite VDM backend
- add a size-conscious TDISP responder with borrowed request decoding and caller-provided response buffers
- wire the DOE emulator validator feature to the TDISP VDM backend
- document the SPDM-Lite TDISP driver interface and supported requests

## Validation
- `cargo test -p mcu-spdm-lite-vdm-handler --features emulated-tdisp`
- `cargo check -p user-app --features test-doe-spdm-tdisp-ide-validator`
- `cargo fmt --check`
- `git diff --check`

## Memory report
Generated locally with `../caliptra-mcu-sw/measure-memory.sh` (`profile=devel`).

### Kernel (`mcu-runtime-emulator`)
| Section | Size |
| --- | ---: |
| `.text` | 142,252 B |
| `.data` | 36 B |
| `.bss` | 24,536 B |

### User-app
| Section | Size |
| --- | ---: |
| `.text` | 98,556 B |
| `.rodata` | 20,764 B |
| `.data` | 12 B |
| `.bss` | 61,384 B |
| FLASH | 119,332 B |
| RAM | 61,396 B |

### SPDM task (symbol attribution from user-app)
| Section | Size |
| --- | ---: |
| `.text` | 84,680 B |
| `.rodata` | 732 B |
| `.data` | 0 B |
| `.bss` | 36,569 B |
| FLASH | 85,412 B |
| RAM | 36,569 B |

| Bundle | Size |
| --- | ---: |
| Runtime bundle | 262,912 B |

### DIFF
--- a/.github/workflows/caliptra-sw-smoke-test.yml
+++ b/.github/workflows/caliptra-sw-smoke-test.yml
@@ -8,7 +8,10 @@ on:
 
 jobs:
   caliptra-sw-mcu-rom-smoke-test:
-    if: ${{ !startsWith(github.head_ref, 'spdm-lite') }}
+    # The caliptra-sw caliptra-2.0 smoke workflow is not compatible with
+    # spdm-lite integration branches. Skip it for PRs targeting spdm-lite,
+    # matching the existing skip for spdm-lite-named source branches.
+    if: ${{ github.base_ref != 'spdm-lite' && !startsWith(github.head_ref, 'spdm-lite') }}
     uses: chipsalliance/caliptra-sw/.github/workflows/fpga-subsystem.yml@caliptra-2.0
     with:
       mcu-commit: ${{ github.event.pull_request.head.sha || github.sha }}
--- a/common/testing/src/spdm_responder_validator/common.rs
+++ b/common/testing/src/spdm_responder_validator/common.rs
@@ -257,12 +257,15 @@ pub fn execute_spdm_tee_io_validator(transport: &'static str) {
                     match child.try_wait() {
                         Ok(Some(status)) => {
                             println!("spdm_tee_io_validator exited with status: {:?}", status);
+                            if !status.success() {
+                                std::process::exit(1);
+                            }
                             break;
                         }
                         Ok(None) => {}
                         Err(e) => {
                             println!("Error: {:?}", e);
-                            break;
+                            std::process::exit(1);
                         }
                     }
                     std::thread::sleep(std::time::Duration::from_millis(100));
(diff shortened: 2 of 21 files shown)


parvathib:
For this PR, please treat the existing `spdm-lib` TDISP stack as the behavioral source of truth. The goal is to port TDISP into `spdm-lite` without changing functionality. 
Please compare command-by-command and error-code-by-error-code against `spdm-lib`, and optionally `spdm-rs` as a secondary reference. Configured values, command validation order, state transitions, nonce handling, report sizing, unsupported-command behavior, and generated error responses should match `spdm-lib` unless there is an explicit spec-backed reason to diverge.

The spdm-lite guidelines should mainly affect structure and resource ownership, not protocol behavior:
- Use static dispatch only: no  dyn , no  Send + Sync , and no  async_trait ; all generics should  resolve at compile time. 
- Keep the stack no-std friendly, avoid heap/boxed futures, and avoid dependencies on `spdm-lib` types.
- Use a single unified scratch buffer per SPDM task. Do not add separate static scratch buffers. All per-task allocations should go through the task’s  `BitmapAllocator`. VDM handlers should receive allocator access, similar to how the OCP VDM handlers are wired. If TDISP handlers or driver operations need temporary storage, they should use task scratch allocator and allocate from it. So, pass the allocator to the methods. 
- Small inline buffers are acceptable only when clearly bounded and justified by a better code-size/BSS tradeoff.

Please also split the TDISP implementation by command instead of keeping all handlers in one large  `tdisp.rs` . 

The spdm-lite port should mirror spdm-lib’s command organization where practical: shared  protocol/wire types and state can stay in common modules, but command handlers such as GET_TDISP_VERSION, GET_TDISP_CAPABILITIES, LOCK_INTERFACE, GET_DEVICE_INTERFACE_REPORT, GET_DEVICE_INTERFACE_STATE,  START_INTERFACE, and STOP_INTERFACE should live in separate command files/modules. That makes the command-by-command parity review easier and keeps future ports/debugging localized.

Documentation changes should be limited to what is required by the `spdm-lite` TDISP port. 

Any functional divergence from `spdm-lib` should either be fixed or explicitly called out with the spec rationale.

Nice-to-have: Use `spdm-lite-style` constants for TDISP wire error codes instead of a `spdm-lib-style` enum. Keep the numeric values and wire behavior identical to spdm-lib/spec; this is only a representation cleanup for consistency with existing spdm-lite-style error constants.

ArthurHeymans:
Closing in favor of #1639 from the renamed spdm-lite-tdisp-on-vdm branch, without the temporary .github workflow change.

PR #1627, opened 2026-06-12
Size Tracking for Image and Stack sizes

### LINKED ISSUE #754: Setup CI for size / usage analysis

v-focutter:
**Summary**

* Tracks the size of the loadable sections in the Kernel, Firmware and User roms
* In addition, the size of the Kernel and User stacks are also tracked
* Generates a HTML report that shows the current & past sizes, and highlights changes.
* Added as a workflow to automatically run on check ins to the main branch.
* Can be run from the command line with: `cargo xtask ci size-history`

Issue https://github.com/chipsalliance/caliptra-mcu-sw/issues/754


builder/src/firmware.rs
```diff
@@ -48,6 +48,24 @@ pub mod hw_model_tests {
     };
 }
 
+pub const MCU_KERNEL_FPGA: FwId = FwId {
+    crate_name: "mcu-runtime-fpga",
+    bin_name: "mcu-runtime-fpga",
+    features: &[],
+};
+
```
mhatrevi: Are these new public FwId constants expected to be reused outside the size-history CI command? If they’re only needed by xtask/src/ci.rs, consider keeping them as private constants there instead of adding them to the shared firmware registry module. That would keep builder::firmware limited to broadly reusable firmware IDs.

builder/src/rom.rs
```diff
@@ -1,148 +1,148 @@
...
-        "ROM binary ({}) is at {:?} ({} bytes)",
-        platform,
-        &rom_binary,
-        std::fs::metadata(&rom_binary)?.len()
-    );
-    let rom_size = rom_size_for_platform(platform);
-    append_rom_digest(&rom_binary, rom_size)?;
-    Ok(rom_binary.to_string_lossy().to_string())
-}
+// Licensed under the Apache-2.0 license
```
mhatrevi: Why is this file modified?

platforms/fpga/runtime/Cargo.toml
```diff
@@ -1,96 +1,96 @@
-# Licensed under the Apache-2.0 license
-
-[package]
```
mhatrevi, line 3: Why is this file modified?

mhatrevi (changes requested):

mhatrevi (approved):

PR #1635, opened 2026-06-17
[fix] rom: program FIPS zeroization mask before config lock

### LINKED ISSUE #1120: [Bug]FIPS zeroization
@swenson @timothytrippel @stellargr 

**FIPS Zeroization Bug Summary**
**Issue**: Register write ordering violation in FIPS zeroization flow causes the zeroization mask to fail.

**Problem Flow:**

**Step 1 (Early Detection)**: PPD signal detected during boot initialization
fips_zeroization = true flag is set
Code continues with normal boot flow

**Step 2 (Configuration Lock)**: After configuring all MCI settings, mci.set_ss_config_done() is called
This locks all configuration CSRs, including the FC_FIPS_ZEROZATION register
Registers become read-only from this point forward

**Step 3 (Caliptra Ready)**: Wait for Caliptra to be ready for mailbox commands

**Step 4 (Zeroization Execution)**: handle_fips_zeroization() is called
Attempts to write mci.set_fips_zeroization_mask()
FAILS silently - the register was already locked in Step 2, so the write is ignored

mhatrevi:
Add an optional fips_zeroization_mask ROM parameter and write it to FC_FIPS_ZEROZATION before SS_CONFIG_DONE/SS_CONFIG_DONE_STICKY lock MCI configuration registers.

When the parameter is None, the register is left at its reset value to preserve existing platform behavior.

### DIFF
--- a/docs/src/rom.md
+++ b/docs/src/rom.md
@@ -59,8 +59,8 @@ These are selected based on the MCI `RESET_REASON` register that is set by hardw
     * See [the ROM fuses](rom-fuses.md) documentation for details on how these are read and interpreted.
 1. Configure MCU mailbox AXI users (see [Security Configuration](#security-configuration) below).
 1. Set mailbox AXI user lock registers.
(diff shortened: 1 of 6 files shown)


mlvisaya (approved):
LGTM

PR #1653, opened 2026-06-18
[fix] Fix OTP access granule mismatch for digest field reads

### LINKED ISSUE #1489: [bug] Access Granule bits do not match when reading the digest value of hek_ratchet_seed from the OTP
Hi. Caliptra WG

code : caliptra-mcu-sw rom sdk2.1 rc1  
 
https://github.com/chipsalliance/caliptra-mcu-sw/blob/12b6dc3f760767a17e931db3b4a4aa5dca47b95f/romtime/src/otp.rs#L649 

This function is used to read the 8ea of HEK ratchet seed, digest, and ZER from OTP.

There is no issue reading the HEK ratchet seed area, as it is defined with an Access Granule of 32 bits. 
However, the **digest and ZER** areas are defined with an **Access Granule of 64 bits**, so the values are not being read correctly.

If the value of digest is 0x1122334455667788, the actual output value when read is 0x1122334411223344.

In particular, since the HEK digest is checked by the MCU-ROM(fn compute_sw_digest(..) ), it is not possible to read a valid HEK ratchet seed after fusing.

https://github.com/chipsalliance/caliptra-mcu-sw/blob/12b6dc3f760767a17e931db3b4a4aa5dca47b95f/romtime/src/otp.rs#L744

Therefore, the code needs to be improved so that the digest and ZER areas can be read using dword reads.

Thank you.

mhatrevi:
The compute_sw_digest() function was reading 64-bit access granule fields (digest, ZER) via two separate 32-bit word reads, which violates the OTP controller's access granule specification. This caused data corruption: reading 0x1122334455667788 would yield 0x1122334411223344 (lower word duplicated).

Fix by using read_dword() for 64-bit granule data instead of two read_word() calls. Convert base address from word units to dword units (byte_offset/8). This respects the OTP DAI's granule semantics and ensures HEK digest validation succeeds after fusing.

Fixes #1489

### DIFF
--- a/romtime/src/otp.rs
+++ b/romtime/src/otp.rs
@@ -774,9 +774,9 @@ impl Otp {
             return Err(McuError::ROM_OTP_INVALID_DATA_ERROR);
         }
 
-        // Read two words at a time from OTP, yielding u64 blocks to the
-        // streaming digest. No large stack buffer required.
-        let base_word = partition.byte_offset / 4;
+        // Read one dword (64-bit) at a time from OTP, yielding u64 blocks to the
+        // streaming digest. Respects OTP 64-bit access granule for digest fields.
+        let base_dword = partition.byte_offset / 8;
         let num_u64_blocks = data_size / 8;
         let mut err: McuResult<()> = Ok(());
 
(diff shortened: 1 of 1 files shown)


mhatrevi:
This is now covered by https://github.com/chipsalliance/caliptra-mcu-sw/pull/1654

PR #1654, opened 2026-06-18
Fix HEK OTP digest and ZER granule reads

### LINKED ISSUE #1489: [bug] Access Granule bits do not match when reading the digest value of hek_ratchet_seed from the OTP

ArthurHeymans:
The HEK production partitions are mixed-granule OTP regions: the 32-byte ratchet seed is accessed as 32-bit words, while the trailing software digest and zeroization marker are 64-bit DAI granules. Reading the whole 48-byte HEK record with read_data() used two 32-bit DAI reads for those trailing fields. On hardware, a word read into the upper half of a 64-bit DAI granule returns the low half of that aligned granule, so values such as 0x1122334455667788 are reconstructed incorrectly as 0x1122334411223344. This breaks HEK digest validation after fusing and can misclassify zeroized or unused HEK slots.

Teach romtime to read HEK partitions according to their RTL layout: seed bytes with read_word(), then the software digest and ZER marker with read_dword(). Use the same HEK partition reader for read_hek_seed(), is_hek_sanitized(), and is_hek_unused() so all HEK consumers observe the same corrected layout.

Update the OTP emulator to mirror RTL DAI granule selection for HEK partitions. HEK digest and ZER marker addresses are aligned to 64-bit granules for reads and writes, while the seed area remains 32-bit. Add emulator tests for both read and write behavior.

Regenerate fuse metadata from the updated caliptra-ss RTL and fix the fuse generator's digest offset calculation for zeroizable partitions. RTL places the digest before the trailing zeroization marker, so generated digest_offset values now point at HEK +32 rather than HEK +40.

Validation: cargo xtask registers-autogen; cargo xtask registers-autogen --check; cargo test -p caliptra-mcu-romtime; cargo test -p caliptra-mcu-emulator-periph; cargo test -p caliptra-mcu-registers-generated -p caliptra-mcu-emulator-registers-generated -p caliptra-mcu-xtask.


ArthurHeymans:
For smoke tests to succeed we need https://github.com/chipsalliance/caliptra-sw/pull/3909 in

mhatrevi (approved):

Issue #961, opened 2026-02-25
[OCP-LOCK]: HEK Fuse Seed rotation


clundin25:
We need to add support for HEK Seed rotation to the reference firmware implementation. It doesn’t have to be advanced but it should provide some commands for rotating through all the HEK seeds. Some work already began in: https://github.com/chipsalliance/caliptra-mcu-sw/pull/853

clundin25:
Done in #1656

PR #1656, opened 2026-06-18
[OCP-LOCK]: Implement HEK seed rotation

clundin25:
This resolves https://github.com/chipsalliance/caliptra-mcu-sw/issues/961


tests/integration/src/runtime/test_mcu_mailbox.rs
```diff
@@ -142,3 +143,100 @@ fn test_fe_prog_authorized_req() -> Result<()> {
...
+
+    // OTP offsets for Slot 0 and 1
+    let slot0_offset =
+        caliptra_mcu_registers_generated::fuses::CPTRA_SS_LOCK_HEK_PROD_0_BYTE_OFFSET;
+    let slot1_offset =
+        caliptra_mcu_registers_generated::fuses::CPTRA_SS_LOCK_HEK_PROD_1_BYTE_OFFSET;
+
+    // Verify initial OTP state (Slot 0 and Slot 1 are not sanitized)
+    let otp_before = hw.read_otp_memory();
+    assert_ne!(otp_before[slot0_offset], 0xFF);
```
zhalvorsen, line 185: Do we expect slot 0 to also be non-zero?

Cargo.toml
```diff
@@ -443,6 +428,7 @@ tock-registers = { git = "https://github.com/tock/tock.git", rev = "release-2.2"
 inherits = "release"
 opt-level = 3
 lto = "thin"
+debug = 1
```
zhalvorsen, line 446: Should this be removed?

runtime/kernel/capsules/src/otp.rs
```diff
@@ -479,3 +551,83 @@ impl SyscallDriver for Otp {
...
+                .map_err(|_| ErrorCode::INVAL)?;
+
+            let digest = caliptra_mcu_otp_digest(&seed, OTP_DIGEST_IV, OTP_DIGEST_CONST);
+
+            ocp_lock_ctx
+                .platform
+                .sanitize_hek_slot(self.driver, active_slot as usize)
+                .map_err(|_| ErrorCode::FAIL)?;
+
+            if slot < total_slots as usize {
```
zhalvorsen: This check should probably be done at the beginning and return an error if it is too high

runtime/kernel/capsules/src/otp.rs
```diff
@@ -479,3 +551,83 @@ impl SyscallDriver for Otp {
...
+                Ok(())
+            });
+            match res {
+                Ok(Ok(())) => {}
+                Ok(Err(e)) => return Err(e),
+                Err(_) => return Err(ErrorCode::FAIL),
+            }
+
+            if self.is_perma_hek_locked()? {
+                return Ok(()); // No-op, already locked
```
zhalvorsen: What do you think of checking if the commanded slot is already burned before doing anything as another no-op state? If the same command is run twice, then fuse state stays the same

Cargo.toml
clundin25: The test binaries were too bloated to compile. This still keeps some of the debug symbols, but at a coarser level. Without this we include the full debug info (Set by the "release" profile).

tests/integration/src/runtime/test_mcu_mailbox.rs
clundin25: This is checking that both slot 0 and 1 are not sanitized. After rotation slot 0 _should_ be sanitized

runtime/kernel/capsules/src/otp.rs
clundin25: IMO let's defer that to `validate_hek_transition`. The HEK state transitions are complex and I'd prefer to push that to the vendor code

runtime/kernel/capsules/src/otp.rs
clundin25: Done!

tests/integration/src/runtime/test_mcu_mailbox.rs
zhalvorsen: Sorry, I mean is there a secret provisioned into slot 0? If so, should we check that slot 0 is non-zero and non-sanitized?

runtime/kernel/capsules/src/otp.rs
zhalvorsen: That's cool with me. Can you add a comment about that being a potential check in the implementation for that function?

tests/integration/src/runtime/test_mcu_mailbox.rs
clundin25: Ah I see what you mean, added a check that slot 0 is not empty and slot 1 is empty

runtime/kernel/capsules/src/otp.rs
clundin25: Done!

zhalvorsen (approved):

PR #1662, opened 2026-06-18
[Doc] Add Caliptra Subsystem Attestation documentation

### LINKED ISSUE #1680: Attestation Documentation
Add Caliptra Subsystem Attestation documentation covering: 
1. Attestation architecture, 
2. Measurement API and Tock capsule design, and 
3. Platform integrator guide for SoC Manifest metadata, persistent measurement storage sizing/configuration

parvathib:
Fixes #1680

Add the Attestation documentation section covering the 
- Caliptra attestation design, 
- Measurement API design, 
- Tock capsule design, and 
- Platform integration guidance.


sunguk-bin-fad:
@parvathib 
I would like to better understand the architectural rationale behind using Software PCR Storage for SoC Non-TCB components.
- Is the primary reason for not utilizing DPE for Non-TCB components due to the limitation of the max number of PL1 DPE contexts (32)?
- Is it mandatory to measure **all** SoC Non-TCB components? : Can we consider omitting measurements for certain Non-TCB components? I'm wondering what criteria we should use to define the scope of measurement

Additionally, I would like to clarify the verification mechanism during a Hitless Update:
- Could you add a section explaining how the verifier validates the device status using both current and journey measurements during a hitless update? I’d like to understand the end-to-end mechanism and whether the OCP EAT and DPE certificate chain are fully sufficient for this process.

docs/src/attestation-tock-capsules.md
```diff
@@ -0,0 +1,378 @@
+# Tock Capsules
```
mlvisaya: nit: rename title from "Tock Capsules" -> "Attestation Tock Capsules"

mlvisaya (approved):
Overall, I think it looks good.

parvathib:
Thanks for the questions!  Please find the answers inline.

> * Is the primary reason for not utilizing DPE for Non-TCB components due to the limitation of the max number of PL1 DPE contexts (32)?


Quick clarification on the numbers/locality first: max DPE contexts are **32 in Caliptra 2.0** and **64 in Caliptra 2.1**, and that pool is split across PL0 and PL1 localities. For inventory attestation, MCU creates SoC TCB contexts in PL0 locality. 

On Software PCR Storage vs DPE for non-TCB: we’re treating this primarily as a trust-semantics split, not just a context-budget split.

-  **TCB** components go into DPE contexts (Caliptra-core protected state) because they’re in protected attestation/key-lineage semantics.  
- **Non-TCB** components are measured and tracked in Software PCR Storage (access-protected MCU SRAM) and reported as claims, without pulling them into AK/CDI lineage.

So, the design goal is to give integrators flexibility to choose that boundary from platform threat model and verifier policy.

> * Is it mandatory to measure **all** SoC Non-TCB components? : Can we consider omitting measurements for certain Non-TCB components? I'm wondering what criteria we should use to define the scope of measurement

In the current design, MCU reports all components it **loads, verifies, and measures** during cold boot and hitless update flows. Non-TCB components are reported as inventory claims via Software PCR storage, and TCB components represented in DPE state but not on the selected DICE key-derivation lineage are also reported as inventory claims. The reporting scope is defined by a static, build-time attestation configuration set by integrator policy. 


> Additionally, I would like to clarify the verification mechanism during a Hitless Update:
> 
> * Could you add a section explaining how the verifier validates the device status using both current and journey measurements during a hitless update? I’d like to understand the end-to-end mechanism and whether the OCP EAT and DPE certificate chain are fully sufficient for this process.

Great point! Yes, I added a section explaining the verifier-flow in high level. The generic verifier flow is:
1. Validate the DICE/DPE certificate chain (device identity + trust anchor),
2. Validate the OCP EAT signature with the AK anchored in that chain,
4. Appraise current digests against reference values (e.g CoRIMs) and/or policy configured,
5. Appraise journey by replaying expected extensions and matching the reported journey digest.

So: DPE chain + signed OCP EAT are sufficient for device authentication and current-state appraisal.
For full journey appraisal, verifier additionally needs the expected extension/replay context (measurement log/policy). This log, however, is not coming from MCU today and still TBD in the architecture.

docs/src/attestation-tock-capsules.md
parvathib: changed.

gmandyam:
a) "TCB components outside that path are reported as OCP EAT inventory claims.".  Are the TCB components in the path reported as part of EAT inventory?
 
b) "CC branch":  This concept is not clearly defined in the document and does not appear to have been discussed in any of the Caliptra meetings.  It should be removed and brought in as a separate architectural proposal.
 
c) Device composition model, MCU RT:  Are "configured SoC TCB components" reported as part of the EAT inventory claim?  Note that 'configured' isn't really defined as far as I can tell.
 
d) DPE context tree pattern:  'retired' is not defined.
 
e) Integrator static config:  How does MCU RT authenticate the configuration?  If that is left to the implementation then it should be stated as such.

This set of documents do not appear ready for approval.

gmandyam (changes requested):
a) "TCB components outside that path are reported as OCP EAT inventory claims.". Are the TCB components in the path reported as part of EAT inventory?

b) "CC branch": This concept is not clearly defined in the document and does not appear to have been discussed in any of the Caliptra meetings. It should be removed and brought in as a separate architectural proposal.

c) Device composition model, MCU RT: Are "configured SoC TCB components" reported as part of the EAT inventory claim? Note that 'configured' isn't really defined as far as I can tell.

d) DPE context tree pattern: 'retired' is not defined.

e) Integrator static config: How does MCU RT authenticate the configuration? If that is left to the implementation then it should be stated as such.

sunguk-bin-fad:
Thanks for your feedback 

> On Software PCR Storage vs DPE for non-TCB: we’re treating this primarily as a trust-semantics split, not just a context-budget split.

For the subsystem mode, is the RTS scope being extended to the MCU? The reason I am asking is that I initially thought measurements should be stored within Caliptra, but it seems the security boundary has now expanded to the MCU, so I wanted to understand the background behind this design choice.

> The reporting scope is defined by a static, build-time attestation configuration set by integrator policy.

I would like to clarify the underlying design intent more specifically. 
Regarding the reporting scope, 
- You stated that it follows the integrator’s policy. 
- Although the [Trademark checklist](https://github.com/chipsalliance/Caliptra/blob/main/doc/trademark/CaliptraChecklistAndEvaluationMethodology.md#boot-and-initialization-process) feels a bit ambiguous, it implies that measuring PA_USER (MCU RT in this case) is mandatory, while measuring other SoC firmwares is merely recommended.

Regarding the MCU SDK design intent,
- My understanding is that the MCU SDK standardizes the SoC image loading mechanism in the MCU RT via load_and_authorize and activate_soc_images (EXEC/GO).

Based on the above, can an integrator bypass these specific APIs and load SoC Non-TCB components without using the standard authorize API or EXEC/GO mechanisms? (Is it correct to assume that there is no strict requirement enforcing the use of load_and_authorize and activate_soc_images?)
 
> So: DPE chain + signed OCP EAT are sufficient for device authentication and current-state appraisal. For full journey appraisal, verifier additionally needs the expected extension/replay context (measurement log/policy). This log, however, is not coming from MCU today and still TBD in the architecture.

Currently, SKIP_STASH = true during the MCU hitless update, but my understanding is that a change is being considered. The MCU RT TCI is stored in the DPE PL0 context. When stash_measurement internally invokes DeriveContext, it does not use the recursive flag, meaning a new DPE context is created every time. After a few hitless updates, the number of PL0 contexts could easily exceed the limit of 32. Is a fix for this potential context exhaustion planned alongside the upcoming update changes?

I understand that the measurement log is currently TBD, but I wanted to ask if Caliptra WG is planning to provide support for the measurement log in the MCU RT. Also, what would be the exact command mapped for the verifier to retrieve this log? (e.g., GET_MEASUREMENT_EXTENSION_LOG)

parvathib:
> a) "TCB components outside that path are reported as OCP EAT inventory claims.". Are the TCB components in the path reported as part of EAT inventory?
> 

Clarified:  TCB components on the selected AK lineage are represented by the DICE/DPE certificate chain and are not repeated as OCP EAT inventory claims. TCB components outside the selected AK lineage can be reported as OCP EAT inventory claims.

> b) "CC branch": This concept is not clearly defined in the document and does not appear to have been discussed in any of the Caliptra meetings. It should be removed and brought in as a separate architectural proposal.
> 

Removed CC branch concept for now; Document now only says confidential-compute attestation is out of scope.
            
> c) Device composition model, MCU RT: Are "configured SoC TCB components" reported as part of the EAT inventory claim? Note that 'configured' isn't really defined as far as I can tell.
>
Clarified “configured” in the document: a `configured` SoC component is one whose `fw_id` is explicitly listed in the `integrator-provided static attestation configuration`. That policy covers both TCB and non-TCB components; the TCB/non-TCB classification determines whether the component is maintained in DPE-backed TCI state or Software PCR Storage. The configured AK target determines which TCB components appear in the DICE/DPE certificate chain and which are reported through OCP EAT inventory claims

> d) DPE context tree pattern: 'retired' is not defined.
> 
Good catch. I clarified the distinction in the document. Caliptra-managed contexts are internal to Caliptra; when they are retired in DPE, their handles are invalid and MCU cannot reach them through DPE commands. MCU-managed contexts, including the MCU Runtime context and downstream SoC TCB contexts, remain active because MCU must keep using their handles through hitless updates.

> e) Integrator static config: How does MCU RT authenticate the configuration? If that is left to the implementation then it should be stated as such.

Clarified that integrator provided config is embedded in MCU Runtime image and authenticated implicitly by MCU image verification. Added `attestation_policy_digest` check for hitless update.

docs/src/attestation-design.md
```diff
@@ -0,0 +1,169 @@
...
+| Configured SoC AK target | If the configured AK target is a downstream SoC TCB node, the AK is derived from the path from root through MCU_RT to that node. |
+Choosing the AK node controls which DPE ancestry contributes to AK derivation. DPE walks from the selected node upward to root. It does not include descendant contexts below the selected node. Descendant SoC component contexts can still be reported as OCP EAT claims.
+
+| Evidence class | Conveying mechanism |
+| --- | --- |
+| TCB components on the AK lineage | DICE/DPE certificate chain |
+| Integrator-selected TCB contexts outside the AK lineage | OCP EAT claims assembled by MCU Runtime |
+| Non-TCB components | OCP EAT claims assembled by MCU Runtime from Software PCR Storage |
+
+`CertifyKey` establishes the AK identity. MCU Runtime assembles OCP EAT claims and the COSE signing structure, then asks Caliptra Core to sign the corresponding bytes using the configured AK. Caliptra Core does not need to interpret the OCP EAT or COSE format.
```
helloxiling, line 104: Does Caliptra core just accept signing request from MCR RT and process it?  Is there any authorization on signing requester?

parvathib: No explicit authorization is required here. The key is derived by the PL0 user, and only the PL0 user can request signing. Since MCU is a privileged user, this is permitted by Caliptra Core.

sunguk-bin-fad:
> Thanks for your feedback
> 
> > On Software PCR Storage vs DPE for non-TCB: we’re treating this primarily as a trust-semantics split, not just a context-budget split.
> 
> For the subsystem mode, is the RTS scope being extended to the MCU? The reason I am asking is that I initially thought measurements should be stored within Caliptra, but it seems the security boundary has now expanded to the MCU, so I wanted to understand the background behind this design choice.
> 
> > The reporting scope is defined by a static, build-time attestation configuration set by integrator policy.
> 
> I would like to clarify the underlying design intent more specifically. Regarding the reporting scope,
> 
> * You stated that it follows the integrator’s policy.
> * Although the [Trademark checklist](https://github.com/chipsalliance/Caliptra/blob/main/doc/trademark/CaliptraChecklistAndEvaluationMethodology.md#boot-and-initialization-process) feels a bit ambiguous, it implies that measuring PA_USER (MCU RT in this case) is mandatory, while measuring other SoC firmwares is merely recommended.
> 
> Regarding the MCU SDK design intent,
> 
> * My understanding is that the MCU SDK standardizes the SoC image loading mechanism in the MCU RT via load_and_authorize and activate_soc_images (EXEC/GO).
> 
> Based on the above, can an integrator bypass these specific APIs and load SoC Non-TCB components without using the standard authorize API or EXEC/GO mechanisms? (Is it correct to assume that there is no strict requirement enforcing the use of load_and_authorize and activate_soc_images?)
> 
> > So: DPE chain + signed OCP EAT are sufficient for device authentication and current-state appraisal. For full journey appraisal, verifier additionally needs the expected extension/replay context (measurement log/policy). This log, however, is not coming from MCU today and still TBD in the architecture.
> 
> Currently, SKIP_STASH = true during the MCU hitless update, but my understanding is that a change is being considered. The MCU RT TCI is stored in the DPE PL0 context. When stash_measurement internally invokes DeriveContext, it does not use the recursive flag, meaning a new DPE context is created every time. After a few hitless updates, the number of PL0 contexts could easily exceed the limit of 32. Is a fix for this potential context exhaustion planned alongside the upcoming update changes?
> 
> I understand that the measurement log is currently TBD, but I wanted to ask if Caliptra WG is planning to provide support for the measurement log in the MCU RT. Also, what would be the exact command mapped for the verifier to retrieve this log? (e.g., GET_MEASUREMENT_EXTENSION_LOG)

@parvathib Could you share your thoughts on this?

docs/src/attestation-design.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Integrator static attestation configuration
+
+MCU Runtime uses integrator-provided static configuration to identify the SoC components included in attestation Evidence. For each `fw_id` listed in this configuration, the policy defines:
+
+1. Whether the component is represented as DPE-backed TCI state or Software PCR Storage.
+2. Whether the component is reported as OCP EAT inventory evidence.
+3. Whether a DPE-backed component is the AK target.
```
jhand2, line 112: What is the reason for using DPE-backed TCI for storage but not including it in the AK target? It seems like you would just want to put it in a software PCR in that case.

docs/src/attestation-measurement-api.md
```diff
@@ -0,0 +1,262 @@
+# Measurement API
+
+This document describes the MCU Runtime measurement API used to collect, store, and report attestation measurement state.
+
+The measurement API provides a single userspace interface for image loading, firmware update, and OCP EAT Evidence generation. Callers do not need to know whether a component's claims are represented by Caliptra DPE state or by the Software PCR Storage capsule.
+
+![Measurement API surface](images/attestation_measurement_api.svg)
+
+## API entry points
```
jhand2, line 17: I'm struggling to understand how these APIs would be exposed outside the iRoT. For example, if I an running in OS userspace (Linux/Windows on the AP) and I want to get a signed eat from MCU, it doesn't seem like there's an API for that. `read_attestation_target_handle` seems to be a helper that MCU can use, but I wouldn't expect that to be exposed outside the iRoT. Is the end-user interface for this architecture out of scope for this PR?

docs/src/attestation.md
```diff
@@ -0,0 +1,20 @@
...
+| [Measurement API](./attestation-measurement-api.md) | Defines the MCU Runtime measurement API used by image loading, firmware update, and SPDM evidence generation. |
+| [Tock Capsules](./attestation-tock-capsules.md) | Defines the reserved SRAM partitioning, syscall drivers, and record formats used to store downstream SoC component attestation records in DPE Handle Storage and Software PCR Storage. |
+| [Attestation Integration Guide](./attestation-integrator-guide.md) | Describes platform integration requirements, including integrator static attestation configuration, reserved SRAM sizing, dummy SoC images, and SPDM/OCP EAT integration. |
+
+## Relationship to SPDM
+
+SPDM is one protocol path used to retrieve Evidence from MCU Runtime. The Evidence describes device identity and inventory claims for downstream SoC components. The attestation design is broader than SPDM transport: it covers measurement collection, persistent measurement state, DPE context handle management, Software PCR-style storage, integrator static attestation configuration, and OCP EAT claim assembly.
```
jhand2: I agree with all this.

I think one additional important difference is that SPDM is callable from outside the SoC (e.g. out-of-band interface exposed to BMC/pRoT).

With the EAT API, we can construct something that is only available to the SoC (e.g. OS) that allows it to prove something about the environment it's currently running in.

docs/src/attestation-measurement-api.md
```diff
@@ -0,0 +1,262 @@
...
+
+| Interface | Caller | Purpose |
+| --- | --- | --- |
+| `measurement_boot_init(boot_context)` | Main user-app startup | Initializes measurement state after reset. Cold boot clears stale state and creates the MCU Runtime root DPE record. MCU hitless update preserves state and validates it against the authenticated attestation policy. |
+| `authorize_and_stash(fw_id, image_metadata)` | Image loading and firmware update | Authorizes a component, enforces anti-rollback policy, and stores its measurement through either the DPE Handle Storage capsule or the Software PCR Storage capsule. `image_metadata` includes the component metadata fields derived from `GET_IMAGE_INFO(fw_id)` and the current load/update operation. |
```
jhand2, line 22: In the attestation WG, we've discussed attesting to SVNs a lot. Is there a convenient way we can allow callers to encode these SVN values in the EAT as well? cc @fdamato

docs/src/attestation-measurement-api.md
```diff
@@ -0,0 +1,262 @@
...
+
+The Tock capsule syscall drivers and reserved SRAM layout used by these APIs are described in [Tock Capsules](./attestation-tock-capsules.md).
+
+## Data interfaces
+
+The Measurement API uses three inputs: integrator configuration for component classification and target selection, SoC component image metadata for authorization/loading, and caller-provided image metadata for initial image load or component update.
+
+### Firmware identifier source
+
+`GET_IMAGE_INFO(fw_id)` is a lookup by firmware identifier; it does not enumerate which images should be loaded.
```
jhand2: nit: `fw_id` is maybe an overloaded term here. DICE uses it to refer to the hash of the image in a TcbInfo. But I think you're using it differently here. Should we maybe pick a different term?

docs/src/attestation-measurement-api.md
```diff
@@ -0,0 +1,262 @@
...
+
+The `fw_id` values used by these flows must match the integrator static attestation configuration and the SoC component image metadata returned by `GET_IMAGE_INFO(fw_id)`. A configured component is a component whose `fw_id` is explicitly listed in the integrator static attestation configuration. The Measurement API must fail unknown or unsupported `fw_id` values explicitly.
+
+### Integrator static attestation configuration
+
+This is static configuration provided by the integrator in the MCU Runtime image. The Measurement API uses it to decide whether each listed `fw_id` is recorded through the DPE Handle Storage capsule or the Software PCR Storage capsule, whether the component is reported as OCP EAT inventory evidence, and which DPE-backed component is the attestation target.
+
+| Field | Description |
+| --- | --- |
+| `fw_id` | Component firmware identifier used as the lookup key. |
```
jhand2, line 63: Is the idea that MCU will enforce this much be unique? If so, probably a good thing to state in the design.

docs/src/attestation-measurement-api.md
parvathib: End-user interfaces are out of scope for this PR. The Measurement API is intended for internal consumption by user-app tasks, such as SPDM over MCTP/DOE tasks, PLDM firmware loader and update tasks, and the MCU mailbox task.

That said, there are currently three ways to retrieve a signed EAT, depending on how the OS communicates with the MCU:


1. **SPDM GET_MEASUREMENTS**
 The standard GET_MEASUREMENTS command returns a signed EAT in measurement block index 0xFD. This is supported over both MCTP and DOE transports.

2. **Caliptra GET_ATTESTATION Command over SPDM VDM**
 The Caliptra GET_ATTESTATION command, when issued through an SPDM VDM, returns a signed EAT over the SPDM over MCTP transport.

3. **Caliptra GET_ATTESTATION Command over MCU Mailbox**
 The Caliptra GET_ATTESTATION command, when issued through the MCU mailbox transport, returns a signed EAT.

1 is already supported. But 2 & 3 are yet to be implemented.

docs/src/attestation-measurement-api.md
parvathib: EAT measurement claims can include SVN values for each target environment.

Anti-rollback (ARB) protection for the MCU RT is enforced by Caliptra Core, and the MCU SVN is encoded in the MCU's TCB Info within the DPE leaf certificate.

For SoC components loaded by the MCU, ARB protection is enforced by the MCU firmware (specifically the firmware loader and updater tasks). These tasks record the SVN of each SoC component in either the DPE context (for TCB components) or a Software PCR measurement record (for non-TCB components). This recorded SVN information can then be reflected in the EAT as a target environment claim for the corresponding SoC component. 

We still need to discuss on what SVN (min SVN since cold boot/fuse SVN/current SVN) to report in the attestation claims.  I added a note about this in the document
@fdamato correct/add if I missed anything.

parvathib:
> Thanks for your feedback
> 
> > On Software PCR Storage vs DPE for non-TCB: we’re treating this primarily as a trust-semantics split, not just a context-budget split.
> 
> For the subsystem mode, is the RTS scope being extended to the MCU? The reason I am asking is that I initially thought measurements should be stored within Caliptra, but it seems the security boundary has now expanded to the MCU, so I wanted to understand the background behind this design choice.
> 

I would not describe subsystem mode as extending Caliptra’s RTS to the MCU.                                                                                                                                                                                                                                                                                      
   In subsystem mode, Caliptra Core remains the Root of Trust: it shields DPE state, CDIs, AK material, and signing. MCU RT is treated as `PA_USER/RTM` firmware, so it is measured and submitted to Caliptra before it becomes the layered attesting environment for downstream SoC components. 
                                                                                                                                                                                                                                                            
   Any subsystem configuration that changes the security properties of MCU RT or other measured firmware is also measured before use. After that, MCU RT can manage downstream subsystem measurement state according to the authenticated Attestation Manifest embedded in MCU RT.

   Caliptra core provides a trustworthy description of MCU RT through its measured and authorized MCU RT DPE context, including any security-affecting configuration and the authenticated Attestation Manifest embedded in MCU RT. Once the verifier accepts that Caliptra-rooted MCU RT layer, trust can be extended to MCU RT for the specific purpose of measuring and reporting the downstream components it loads and updates.



> > The reporting scope is defined by a static, build-time attestation configuration set by integrator policy.
> 
> I would like to clarify the underlying design intent more specifically. Regarding the reporting scope,
> 
> * You stated that it follows the integrator’s policy.
> * Although the [Trademark checklist](https://github.com/chipsalliance/Caliptra/blob/main/doc/trademark/CaliptraChecklistAndEvaluationMethodology.md#boot-and-initialization-process) feels a bit ambiguous, it implies that measuring PA_USER (MCU RT in this case) is mandatory, while measuring other SoC firmwares is merely recommended.
> 
> Regarding the MCU SDK design intent,
> 
> * My understanding is that the MCU SDK standardizes the SoC image loading mechanism in the MCU RT via load_and_authorize and activate_soc_images (EXEC/GO).
> 
> Based on the above, can an integrator bypass these specific APIs and load SoC Non-TCB components without using the standard authorize API or EXEC/GO mechanisms? (Is it correct to assume that there is no strict requirement enforcing the use of load_and_authorize and activate_soc_images?)
> 

For MCU SDK-managed SoC images, the authorization behavior follows the `SoC/Auth Manifest policy`. The SoC Manifest can explicitly mark a component with `ignore_auth_check`, in which case Caliptra skips the digest comparison and returns the image as authorized. That is different from bypassing the MCU RT loading/update path.

 The expected flow is still that MCU RT loads the image, calls into Caliptra authorization, and then activates the image through the standard MCU RT path. The Attestation Manifest **does not replace** the SoC Manifest authorization policy; it defines how the measured component’s attestation state is maintained, for example DPE-backed TCB state or Software-PCR-style non-TCB state.

 In the current design, the Attestation Manifest entries under the MCU node are expected to match the SoC Manifest image entries, excluding MCU RT itself. That lets the Measurement API  maintain attestation state for every SoC image that MCU RT loads and authorizes.


> > So: DPE chain + signed OCP EAT are sufficient for device authentication and current-state appraisal. For full journey appraisal, verifier additionally needs the expected extension/replay context (measurement log/policy). This log, however, is not coming from MCU today and still TBD in the architecture.
> 
> Currently, SKIP_STASH = true during the MCU hitless update, but my understanding is that a change is being considered. The MCU RT TCI is stored in the DPE PL0 context. When stash_measurement internally invokes DeriveContext, it does not use the recursive flag, meaning a new DPE context is created every time. After a few hitless updates, the number of PL0 contexts could easily exceed the limit of 32. Is a fix for this potential context exhaustion planned alongside the upcoming update changes?
> 
Correct. We should not create a new PL0 DPE context for each MCU hitless update. The Measurement API model updates existing DPE-backed TCB contexts and persists any rotated handles.

 Today, MCU RT hitless update uses `SKIP_STASH = true`, so the new MCU RT image is authorized but the MCU RT DPE measurement is not updated. When we change that behavior, the update  path should update the existing MCU RT DPE-backed state and keep DPE context usage bounded. 



> I understand that the measurement log is currently TBD, but I wanted to ask if Caliptra WG is planning to provide support for the measurement log in the MCU RT. Also, what would be the exact command mapped for the verifier to retrieve this log? (e.g., GET_MEASUREMENT_EXTENSION_LOG)

The measurement-extension log is not planned to be produced or served by MCU RT. The expected direction is that replay/log context comes from outside MCU RT, such as the BMC or update agent that orchestrates the firmware updates.
MCU RT reports the current and journey measurement values in evidence, but it does not own the verifier replay log.

docs/src/attestation-measurement-api.md
parvathib: Good point. I kept `fw_id` because it matches the existing `GET_IMAGE_INFO` / SoC Manifest terminology. I added a clarification that, in this document, `fw_id` represents the `fw_id` field in the SoC Manifest image metadata.

docs/src/attestation-measurement-api.md
parvathib: Yes. The SoC/Auth Manifest path already enforces unique `fw_id` values, and the Attestation Manifest builder/runtime validation also rejects duplicate `fw_id` entries. I clarified the design to state that Attestation Manifest entries under the MCU node must be unique and match the SoC Manifest image entries, excluding MCU RT itself.

docs/src/attestation.md
parvathib: Agreed. I clarified the requester-facing paths more precisely.

 The Measurement API is internal to MCU Runtime tasks. Signed Evidence retrieval is transport-specific:
 * BMC/pRoT requesters use SPDM over MCTP.
 * PCIe requesters use SPDM over DOE, which is the PCIe DOE path into the MCU.
 * SoC-local requesters, such as AP OS or TEE, can use the MCU mailbox path.

 All of these paths consume the same MCU Runtime evidence-generation flow. This PR defines the internal measurement and evidence assembly model; the requester-facing transport APIs are defined separately.

docs/src/attestation-design.md
parvathib: That’s a fair question. The value of DPE-backed TCI outside the AK lineage is measurement-state provenance, not AK derivation.

 The intent is to give the integrator/verifier two separate choices:

 1. Where the measurement state is maintained: Caliptra/DPE-backed TCI state vs MCU-managed software PCR record.
 2. Which DPE-backed context, if any, is selected as the AK target.

 While the verifier cannot infer this storage choice from the EAT claim alone, it appraises the claim according to the authenticated Attestation Manifest / platform policy for that MCU Runtime image. 

If the policy says DPE-backed provenance is not required for a component, the MCU-managed software PCR record is the right choice. If the policy requires the component’s measurement to be maintained by DPE but not folded into AK identity, then DPE-backed but not AK-targeted is useful.

That said, I’m open to discussing whether this distinction is useful in practice or whether we should simplify the policy model.

docs/src/attestation-design.md
```diff
@@ -0,0 +1,169 @@
...
+                └─ <integrator-selected SoC TCB context tree>
+```
+
+The chain above `MCU_RT` is common across integrations. Device-specific variation occurs below `MCU_RT`, where TCB components are arranged according to integrator static attestation configuration and the platform's attestation scenario.
+
+| Level | DPE context | Description | Update and handle model |
+| --- | --- | --- | --- |
+| 0 | `Root("RTMR")` | DPE root context | Caliptra-managed lineage context; no MCU command handle. |
+| 1 | `CCIV("CCIV")` | Caliptra Runtime context | Caliptra-managed lineage context; no MCU command handle. |
+| 2 | `ROM_Stash_1..N` | ROM-stashed measurements | Caliptra-managed lineage context; immutable; no MCU command handle. |
```
sunguk-bin-fad, line 76: In this context, does 'immutable' mean a FW that does not support hitless updates? Can we think of it as a FW that is updatable after a cold reset, even though it doesn't support hitless updates? I am asking this to clarify whether a device-specific FW can be loaded and stashed before the Caliptra RT is loaded, and if so, whether that FW's attribute must be immutable.

sunguk-bin-fad (commented):
I am asking follow-up questions to establish the precise direction for the SoC-side HW/FW architecture when integrating the Caliptra subsystem.

> I would not describe subsystem mode as extending Caliptra’s RTS to the MCU. 
My understanding of and expectations for the Caliptra PCR were as follows:
1. While the MCU SRAM is access-protected, choosing SW PCRs over the available Caliptra PCRs feels a bit like opting for an fTPM over a discrete TPM. I was wondering if this is due to specific technical requirements, like Dynamic PCRs, or resource constraints within Caliptra.
2. Previously, I understood that Caliptra measurements could be stored in either a DPE context or a PCR. However, based on the requirements below (2-1, 2-2), I thought that using PCRs for all measurements was mandatory and that both DPE and PCR must be used together (e.g., via stash_measurement) within a TCB like Caliptra. This is why the SW PCR storage concept feels a bit unnatural to me, and I am curious if this SW PCR design actually accounts for these requirements.
> The measurement-extension log is not planned to be produced or served by MCU RT. The expected direction is that replay/log context comes from outside MCU RT, such as the BMC or update agent that orchestrates the firmware updates. MCU RT reports the current and journey measurement values in evidence, but it does not own the verifier replay log.

2-1. Caliptra leverages GET_PCR_LOG to enable the SoC to construct its measurement log. Since the [Caliptra Trademark requires log functionality](https://github.com/chipsalliance/Caliptra/blob/main/doc/trademark/CaliptraChecklistAndEvaluationMethodology.md#log-functionality-in-soc) to support compounded measurements, it mandates that the integrator implement this capability. Therefore, I believed that maintaining strict consistency with Caliptra’s native GET_PCR_LOG would be the ideal approach.

Given this, I expected that under Subsystem Mode, the MCU RT would at least provide a consistent measurement log mechanism identical or highly similar to GET_PCR_LOG. However, the introduction of this new SW PCR storage concept feels somewhat divergent from that direction. Even when utilizing SW PCR storage, shouldn't we consider managing these logs through GET_PCR_LOG or an equivalent, unified mechanism?

2-2. The [Attestation of SoC Update Journey](https://chipsalliance.github.io/caliptra-web/docs/2.1/caliptra_overview.html?highlight=attestation#attestation-of-soc-update-journey) spec states that Caliptra shall maintain a reboot counter for each component. As far as I can find in the current specification, there is no corresponding register or counter other than PCR_RESET_COUNTER. My understanding is that the reboot counter can be used as a mechanism designed to help minimize the size of the measurement log that the SoC needs to manage. I hope that a feature equivalent to PCR_RESET_COUNTER will also be explicitly supported for the SW PCR storage.

As an integrator, it’s tough to keep track of where Caliptra’s architecture is heading. I really hope the new Attestation Architecture design stays consistent with existing features/specs like GET_PCR_LOG and PCR_RESET_COUNTER.

> For MCU SDK-managed SoC images, the authorization behavior follows the `SoC/Auth Manifest policy`. The SoC Manifest can explicitly mark a component with `ignore_auth_check`, in which case Caliptra skips the digest comparison and returns the image as authorized. That is different from bypassing the MCU RT loading/update path.
> 
> The expected flow is still that MCU RT loads the image, calls into Caliptra authorization, and then activates the image through the standard MCU RT path. 

Looking into it further, since the image loader is platform-specific, the integrator can set skip_stash when calling load_and_authorize(), select ignore_auth_check in the SoC manifest, and call activate_soc_images() optionally (meaning integrators can choose their own FW activation mechanism). This provides design flexibility for the integrator. However, my understanding is that loading all FW via the image loader (flash/PLDM) through load_and_authorize() is mandatory, while the auth check and stash are optional. Is it correct?

Under this architecture, my understanding is that a FW1 on Vendor CPU 1 is not allowed to measure/load/activate FW2 for Vendor CPU 2. Similarly, FW A is not permitted to measure/load/activate FW B on the same Vendor CPU. Instead, the MCU RT must handle the measuring/loading of all FWs belonging to the SoC, but it is possible to delegate the activation of certain FWs to another Vendor CPU. Is my understanding correct?

docs/src/attestation-measurement-api.md
jhand2: > End-user interfaces are out of scope for this PR

Ok, thanks :) I think we may have some thoughts on (3) since it is a potential replacement for exposing DPE over the MCU mailbox. But we can follow-up outside this PR. cc @zhalvorsen

docs/src/attestation.md
jhand2: Thanks!

docs/src/attestation-measurement-api.md
jhand2: Ok, makes sense. Thank you.

jhand2 (approved):

docs/src/attestation-design.md
jhand2: I don't feel super strongly about this, but I do think it makes the API surface a bit more complicated and I don't think the distinction changes too much since MCU SW can always choose to include whatever non-TCB measurements in the EAT.

I also think we might be able to reduce Caliptra SRAM/mailbox usage & overall complexity if we don't require DPE to hold all of the non-TCB TCIs.

docs/src/attestation-design.md
```diff
@@ -0,0 +1,178 @@
...
+![Layered attestation model](images/layered_attestation_model.svg)
+
+## DPE context tree pattern
+
+All device integrations share a common root chain managed by Caliptra:
+
+```text
+Root("RTMR")
+ └─ CCIV("CCIV")
+    └─ ROM_Stash_1..N
```
RaunakGu, line 63: Hi @parvathib,
What is the use of Caliptra ROM own measurement, is it considered TCB or non-TCB?" Currently it seems neither.

Can we consider adding Caliptra ROM measurement in the DPE Context tree same as MCU ROM measurement being added, total we will have 3 Stash measurement during ROM Phase?

Image

docs/src/attestation-design.md
```diff
@@ -0,0 +1,178 @@
...
+       └─ SoC_Manifest_Vendor("SOMV")
+          └─ SoC_Manifest_Owner("SOMO")
+             └─ MCU_RT("MCFW")
+                └─ <integrator-selected SoC TCB context tree>
+```
```
RaunakGu, line 68: Today the MCU ROM **pre-calculates** the MCU/SoC ROM digest (SHA-384, via Caliptra's `CM_SHA` crypto-mailbox command) and submits that finished digest to Caliptra via `STASH_MEASUREMENT`. Caliptra does not re-hash it — it records the supplied digest into PCR31 and the measurement log, and only folds it into a DPE context later, at RT `initialize_dpe`.

Is there a way to compute these MCU/SoC ROM measurements using the SHA accelerator directly and land the resulting digest into a DPE context, rather than an externally-calculated digest being used in the DPE?

docs/src/attestation-design.md
```diff
@@ -0,0 +1,178 @@
...
+
+The chain above `MCU_RT` is common across integrations. Device-specific variation occurs below `MCU_RT`, where TCB components are arranged according to integrator static attestation configuration and the platform's attestation scenario.
+
+| Level | DPE context | Description | Update and handle model |
+| --- | --- | --- | --- |
+| 0 | `Root("RTMR")` | DPE root context | Caliptra-managed lineage context; no MCU command handle. |
+| 1 | `CCIV("CCIV")` | Caliptra Runtime context | Caliptra-managed lineage context; no MCU command handle. |
+| 2 | `ROM_Stash_1..N` | ROM-stashed measurements | Caliptra-managed lineage context; immutable; no MCU command handle. |
```
RaunakGu, line 76: Today `STASH_MEASUREMENT` only accepts a pre-calculated 48-byte digest (`measurement: [u8; 48]`); the caller must hash the image externally, so Caliptra records an externally-supplied measurement rather than one it computed itself.

Request: add an option to `STASH_MEASUREMENT` to take a source address + image size and stream the image through the SHA accelerator to compute the digest internally — the same mechanism `AUTHORIZE_AND_STASH` already uses. Its request carries `source` + `image_size` (LoadAddress / StagingAddress) and the handler DMA-streams the image through `sha2_512_384_acc` (`dma_image.sha384_image(...)`) to derive the digest on-chip, instead of trusting a caller-supplied one.

This would let MCU/SoC ROM measurements be computed by Caliptra and landed into DPE, rather than relying on an externally-calculated digest.

docs/src/attestation-design.md
```diff
@@ -0,0 +1,178 @@
...
+
+The DICE/DPE certificate chain authenticates the device, and the signed OCP EAT conveys the `current` and `journey` values. Journey appraisal requires a log of the extended measurements so the Verifier can interpret and replay the reported journey digests (see [Extended-measurement log](#extended-measurement-log)).
+
+## Extended-measurement log
+
+Appraising a `journey` measurement requires a log of the extended measurements that produced it: the Verifier replays the logged measurements and confirms they reproduce the reported journey digest. This applies wherever a measurement journey is appraised, so it is a general attestation concern, not specific to inventory attestation.
+
+Generating and conveying this extended-measurement log is outside the scope of MCU Runtime and is yet to be defined. MCU Runtime does not produce this verifier replay log today.
```
RaunakGu, line 180: A DPE context's journey (cumulative) TCI accumulates across firmware updates: on each update Caliptra computes `journey = SHA-384(prev_journey || new_measurement)`, so the value depends on the ordered history of FW updates, not just the currently running firmware.

Because of this, journey/integrity-register values are runtime-computed and are **not** included in the reference-value CoRIM (the CoRIM generator notes they "cannot be pre-determined from static build artifacts"). A verifier that wants to appraise the journey value must therefore keep track of which FW updates a device has taken and **reconstruct/replay** the expected cumulative measurement itself, folding each update's measurement in order.

Does the host-side verifier tool account for this today — i.e., does it reconstruct/replay journey measurements across FW updates — or does it only corroborate the static (current) reference values?

docs/src/attestation-design.md
parvathib: I don’t think Caliptra ROM can directly measure arbitrary MCU/SoC ROM or SoC component images into DPE

In the ROM phase, Caliptra only has the firmware image bundle handed to it through the mailbox/recovery flow, and that path is for `Caliptra FMC/RT verification/loading`. ROM does not have a generic mechanism to locate and hash external SoC components by itself

As you noted, for early SoC measurements, the supported model is `STASH_MEASUREMENT` - the caller supplies the already-computed 48-byte digest. ROM extends `PCR31` with that digest and records it in the `measurement log`. Later, when `Caliptra RT` initializes `DPE`, it replays those logged measurements into DPE contexts.

So, for `pre-RT SoC measurements`, the digest has to be provided to Caliptra. ROM does not compute it directly into DPE. 
After RT is running, different flows can hash/load/authorize images through RT-supported mechanisms, but that is not the ROM-stashed measurement path.

docs/src/attestation-design.md
parvathib: This is a new Caliptra Core firmware feature request, not an MCU Runtime implementation detail.

Can you raise an RFC in caliptra-sw if this is the core requirement?

docs/src/attestation-design.md
parvathib: I'm not aware of any open-source verifier today that can process the OCP EAT profile. We do have a reference verifier that we use for integration testing and CI (see the `attestation.yml` workflow in this repository), but it does not yet verify journey measurements. We also have not defined a standard log format for this purpose yet.

docs/src/attestation-design.md
parvathib: Just to be clear, ROM_Stash_* refers to the early SoC measurements stashed by Caliptra ROM, not to the Caliptra ROM measurement itself. Whether these ROM-stashed measurements are present, and how many there are, is entirely dependent on the SoC integrator's requirements.

docs/src/attestation-design.md
parvathib: yes, that is correct! immutable here means the contexts can't be updated hitlessly. Currently, SoC component measurements stashed by ROM are immutable.

docs/src/attestation-design.md
parvathib: I agree that this option adds API and resource complexity, and we do not yet have a concrete integrator use case demonstrating its benefit. However, I would prefer to retain it for now because measurement-state provenance and AK-lineage inclusion are separate policy choices.

An integrator may require a security-critical component’s measurement state to be maintained in Caliptra-protected DPE state while selecting a different context as the inventory AK target. In this design, that component remains a SoC TCB component; it is simply outside the selected AK lineage and is reported through the OCP EAT.

Also, I think `non-TCB TCIs` may be referring to DPE contexts outside the selected AK lineage. That differs from this document’s use of “SoC non-TCB,” which specifically means Software-PCR-backed measurement state.

If we cannot identify an integrator that benefits from DPE-backed measurement provenance outside the AK lineage, I agree that we should simplify the model and route those components through Software PCR Storage.

docs/src/attestation.md
```diff
@@ -0,0 +1,34 @@
...
+| [Static Integrator Configuration](./static_integrator_configuration.md) | Defines the integrator-owned static configuration artifacts used by image loading and attestation, including the SoC/Auth Manifest metadata, Attestation Manifest, and SoC image load list. |
+| [Tock Capsules](./attestation-tock-capsules.md) | Defines the reserved SRAM partitioning, syscall drivers, and record formats used to store downstream SoC component attestation records in DPE Handle Storage and Software PCR Storage. |
+| [Attestation Integration Guide](./attestation-integrator-guide.md) | Describes platform integration requirements, including integrator static attestation configuration, reserved SRAM sizing, dummy SoC images, and SPDM/OCP EAT integration. |
+
+## Relationship to SPDM
+
+MCU Runtime produces one attestation report: the signed OCP EAT. SPDM only carries that report to the requester; no additional SPDM attestation report is created or included inside it. The MCU mailbox provides another retrieval path for the same report.
+
+The requester boundary differs by transport:
```
helloxiling, line 21: One suggestion: can we focus the description from the device view and avoid going into the requester path details?

Something like this:  
"MCU Runtime produces one attestation report: the signed OCP EAT. The device can expose this report to the requester through three transport paths:
- SPDM over MCTP
- SPDM over PCIe DOE
- MCU mailbox
When using SPDM, the device can return the signed OCP EAT through GET_MEASUREMENTS in measurement block index 0xFD or through the Caliptra GET_ATTESTATION command carried as an SPDM VDM. The MCU mailbox can expose the same GET_ATTESTATION command directly. "

docs/src/attestation.md
```diff
@@ -0,0 +1,34 @@
+# Attestation
+
+The Caliptra attestation architecture defines how Caliptra Core and MCU Runtime establish device identity, measure downstream SoC components, and produce Evidence that lets Verifiers appraise comprehensive device state.
```
helloxiling, line 3: nit: do we want to remove `The Caliptra attestation architecture defines ...`, since we call it design spec for now?

helloxiling (commented):
Overall, it looks great!
Just adding a few rephrase suggestions.

docs/src/attestation-design.md
RaunakGu: Hi @parvathib,

Thanks for the reply. My question is whether Caliptra can stash its own ROM measurement into DPE via these ROM_Stash. Since this is not something that can be done by SoC integrators?

docs/src/attestation-design.md
RaunakGu: I have raised an issue [here ](https://github.com/chipsalliance/caliptra-sw/issues/4023)

PR #1681, opened 2026-06-23
test: fix warm reset test boot initialization

eumgil0812:
## Summary

Initialize firmware through InitParams in the warm reset integration test.

The previous test setup did not reach COLD_BOOT_FLOW_COMPLETE,
causing the warm reset test to fail before exercising the warm reset flow.

## Testing

- Ran test_warm_reset_success
- Verified COLD_BOOT_FLOW_COMPLETE is reached
- Verified WARM_RESET_FLOW_COMPLETE is reached


clundin25:
@eumgil0812  can you clean up the whitespace changes?

eumgil0812:
Cleaned up the whitespace changes and updated the PR.

clundin25:
@eumgil0812 should there be a test that is enabled? I would expect this to have been failing prior to your change. Can you also share how your fix resolved the test failure?

eumgil0812:
This test is ignored on the emulator, so I executed it on the actual FPGA(vck190). However, it consistently panicked at the COLD_BOOT_FLOW_COMPLETE assertion.

The root cause was that InitParams was missing the binary file information required for the I3C streaming boot, preventing the cold boot from ever completing.

By adding the missing initialization in InitParams, the test now successfully drives the boot state-machine through both cold boot and warm reset flows, passing cleanly on the FPGA target:
test rom::test_warm_reset::test_warm_reset_success ... ok

Since I am still a junior developer and relatively new to this project, please let me know if there are any further details, logs, or code cleanups required. I would be more than happy to provide them. Thank you!

clundin25:
Thanks for your contribution @eumgil0812. We use a nextest filter at `.config/nextest.toml` to enable tests for the FPGA. Should that be updated to include this test?

It seems the tests failures in CI are unrelated to your PR

eumgil0812:
Thanks for the review! 

I've updated the nextest filter in `.config/nextest.toml` to include this test as you suggested.

eumgil0812:
Updated the integration test target to test_warm_reset::test_warm_reset_success since the previous path was incorrect.

PR #1684, opened 2026-06-24
[provisioning] add TestUnlocked provisioning binary.

### LINKED ISSUE #872: [provisioning] develop reference fuse configuration MCU programs

### LINKED ISSUE #982: [provisioning] add hw-model function to load a provisioning binary into MCU SRAM
To support #872, we will need to enhance the HW model's JTAG [library](https://github.com/chipsalliance/caliptra-mcu-sw/blob/1e800d7e3c0c7896b41d8df0354e2c39323cbb30/hw/model/src/jtag.rs) to add a function that can load a bare-metal binary to the MCU SRAM.

amit-kumarh:
This PR includes a common provisioning FW library, a bare-metal binary target that provisions TestUnlocked fuses (primarily LC and Manuf Debug Unlock tokens), and an integration test to run the flow.


hw/model/src/jtag.rs
```diff
@@ -115,3 +115,25 @@ pub fn jtag_get_caliptra_mailbox_resp(tap: &mut OpenOcdJtagTap) -> Result<Vec<u8
         .context("Unable to clear MboxExecute.")?;
     Ok(rsp_bytes)
 }
+
+pub fn sideload_binary(tap: &mut OpenOcdJtagTap, bytes: &[u8], offset: u32) -> Result<()> {
+    // Disable watchdog timers to prevent reset during JTAG sideloading
+    tap.write_memory_32(0xa800_00b0, 0)?;
+    tap.write_memory_32(0xa800_00c0, 0)?;
```
clundin25: nit: Do you mind using a constant for these addresses?

tests/integration/src/lib.rs
```diff
@@ -1315,6 +1309,13 @@ mod test {
         }
     }
 
+    pub fn clear_mcu_errors(model: &mut DefaultHwModel) {
+        model
+            .mcu_manager()
+            .with_mci(|mci| mci.fw_error_fatal().write(|_| 0x0));
+        model.output().clear_exit_status();
+    }
```
clundin25: Does it make sense to make this a model function? Seems like it could be useful in external integrations

clundin25 (approved):

tests/integration/src/lib.rs
amit-kumarh: good call, done

hw/model/src/jtag.rs
amit-kumarh: Done. to keep the model generic, took in the MCI base as a parameter and added the offset from the register package

PR #1691, opened 2026-06-24
Enable all features

mlvisaya:
cargo xtask runtime-build --profile release will now include all features.

where  all-features = ["spdm", "streaming-boot", "flash-boot", "firmware-update", "doe", "userspace-log"]


platforms/emulator/runtime/userspace/apps/user/src/mcu_mbox/mod.rs
```diff
@@ -36,16 +30,9 @@ async fn start_mcu_mbox_service() -> Result<(), ErrorCode> {
...
-        feature = "test-mcu-mbox-fips-self-test",
-        feature = "test-mcu-mbox-fips-periodic",
-        feature = "test-caliptra-util-host-validator",
-        feature = "test-defmt-logging-mailbox",
-        feature = "test-defmt-logging-release"
-    ))]
+    #[cfg(feature = "mcu-mbox-service")]
     {
-        let handler = cmd_handler_mock::NonCryptoCmdHandlerMock::default();
+        let handler = cmd_handler_mock::NonCryptoCmdHandlerMock;
```
parvathib, line 35: Should we enable `NonCryptoCmdHandlerMock` in the default build? This is mainly used for tests and gives out test platform information. The mock handler is documented as emulator integration-test code.

firmware-bundler/reference/emulator/user-app.toml
```diff
@@ -38,5 +38,5 @@ stack = 0x0_7000
 
 [[app]]
 name = "user-app"
-stack = 0xae00
+stack = 0x8e00
```
parvathib, line 41: Was this stack size reduction based on measured usage? Since this PR enables more user-app tasks by default, I would have expected the required stack to increase rather than decrease

platforms/emulator/runtime/userspace/apps/user/src/mcu_mbox/mod.rs
mlvisaya: I think we can keep it. The MCU Mbox task needs to be given handlers in order to work. At any rate, the integrators still need to replace it with their own. Also, it's good to include these mocks so that it will be included in the size calculation.

firmware-bundler/reference/emulator/user-app.toml
mlvisaya: The fpga code size got larger, so the compilation was failing, i had to find somewhere to reduce the size and found out that reducing the stack size helped and since the integ tests still pass, so i think the stack reduction is already at functional minimum.

parvathib (approved):

helloxiling (approved):
Looks good. Thanks for putting it together, Marco!  

nit: It would be great if we can add a memory footprint analysis to this PR.

mlvisaya:
> Looks good. Thanks for putting it together, Marco!
> 
> nit: It would be great if we can add a memory footprint analysis to this PR.

Thanks. The size-history workflow that runs every commit should be able to show the memory and code size foot print.

Issue #1666, opened 2026-06-23
[spdm-vdm] Debug Unlock Follow up/Cleanup


parvathib:
- [x] Confirm DebugUnlock VDM payload contract

    • Current behavior: the host builds the `AuthorizeDebugUnlockToken` VDM payload as **Caliptra RT mailbox bytes**, including `MailboxReqHeader/checksum`, and MCU forwards those bytes directly to the Caliptra mailbox.
    • Expected clean separation would be the SPDM VDM payload matches the OCP/Caliptra VDM protocol definition, and the MCU platform hook translates that VDM payload into the Caliptra mailbox request format.
    • Confirm which contract is intended and matches the SPDM VDM documentation.
    • If host-prebuilt mailbox payload is intended, preserve the exact payload bytes unchanged.
    • If VDM payload should be token-only, move mailbox header/checksum construction to the **MCU platform hook**.
 

- [x] Avoid token staging in  `execute_with_payload_stream`

    • Current path copies `DebugUnlock` token chunks into  `execute_with_payload_stream` ’s local **[u8; PAYLOAD_CHUNK_* before  `send_chunk().await `.
    • Add/use a mailbox helper that holds  MAILBOX_MUTEX  and sends caller-provided payload slices directly:
           •  `start_chunked_request` 
           •  `send_chunk(payload.chunks(...)) `
           •  `execute_chunked_request` 
    • This avoids staging token bytes in an extra async local buffer. 
 

- [x]  Validate DebugUnlock challenge response length

    • Capture the byte count returned by  `execute_mailbox_cmd` .
    • Require:
      `resp_len == core::mem::size_of::<ProductionAuthDebugUnlockChallenge>()`
    • Return a VDM completion error if the response is short or malformed.
 

- [x] Move all CHUNK_SEND large requests to request streaming

    • Ideally, `DebugUnlock` should be handled as a streamed large request. It is currently handled as a buffered large request.
    • Currently only `SET_CERTIFICATE` and `DebugUnlock` are the only large requests that use CHUNK_SEND.  We should make them both streaming large requests.
    •  Make `AuthorizeDebugUnlockToken` stream CHUNK_SEND payload chunks directly to the mailbox path.
    •  `SET_CERTIFICATE`  should also stream CHUNK_SEND payload chunks rather than reassembling the full cert chain in  large_msg_ctx .
    • Keep only small state in  ActiveLargeRequest .
[...]

PR #1694, opened 2026-06-25
Implement streaming DebugUnlock and SET_CERTIFICATE

### LINKED ISSUE #1356: [SPDM_CERT_PROVISIONING] Add support for SET_CERTIFICATE
SPDM SET_CERTIFICATE command is used to provision certificate chain in owner and tenant slots.

ArthurHeymans:
Fixes #1356 

## Summary

- Document and preserve the DebugUnlock `AuthorizeDebugUnlockToken` VDM payload as host-prebuilt Caliptra RT mailbox request bytes, including `MailboxReqHeader`/checksum.
- Stream recognized `AuthorizeDebugUnlockToken` CHUNK_SEND payloads directly into a guarded Caliptra mailbox chunked transaction without full large-request reassembly.
- Stream `SET_CERTIFICATE` CHUNK_SEND requests through a minimal non-atomic cert-store transaction and commit metadata only after final validation.
- Add mailbox chunked-request abort support and ensure interrupted streaming requests abort before SPDM state reset.
- Preserve exact DebugUnlock challenge response length validation.

Addresses #1666 

## SPDM validator stats

MCTP:

```text
test suite (spdm_responder_conformance_test) - pass: 655, fail: 0
```

DOE:

```text
test suite (spdm_responder_conformance_test) - pass: 942, fail: 0
```

## Memory stats

```text
Memory report — profile: devel

Kernel (mcu-runtime-emulator)
  Section            Size (B)
  ──────────────── ──────────
  .text               142,232
  .data                    36
  .bss                 24,536

User-app
  Section            Size (B)
  ──────────────── ──────────
  .text               114,978
  .rodata              19,336
  .data                    40
  .bss                 53,644
  FLASH               134,354
  RAM                  53,684

SPDM task (symbol attribution from user-app)
  Section            Size (B)
  ──────────────── ──────────
  .text                99,322
  .rodata                 732
  .data                    28
  .bss                 28,873
  FLASH               100,082
  RAM                  28,901

  Runtime bundle      278,016
```

### DIFF
--- a/docs/src/caliptra_spdm_vdm_cmds.md
+++ b/docs/src/caliptra_spdm_vdm_cmds.md
@@ -4,7 +4,7 @@
 
 This document describes how Caliptra external commands are transported over SPDM Vendor Defined Messages (VDM) via MCTP. This is the **out-of-band (OOB)** path for accessing Caliptra device management commands from an external agent such as a BMC.
 
-For command definitions (categories, payloads, and completion codes), see [Caliptra Common Commands](caliptra_common_commands.md).
+For command definitions (categories, payloads, and completion codes), see [Caliptra Common Commands](caliptra_common_commands.md). SPDM VDM payloads generally use those command payload definitions directly and do not include Caliptra RT mailbox `MailboxReqHeader` or checksum bytes. `Authorize Debug Unlock Token` is the exception: its VDM payload is a host-prebuilt Caliptra RT mailbox request, including `MailboxReqHeader.checksum`, so the MCU can stream the request bytes unchanged into the Caliptra mailbox FIFO.
 
 For the unified software architecture shared between OOB (SPDM VDM) and in-band (MCI Mailbox) paths, see [Unified Caliptra Command Handling](unified_caliptra_command_handling.md).
 
--- a/docs/src/caliptra_spdm_vdm_cmds.md
+++ b/docs/src/caliptra_spdm_vdm_cmds.md
@@ -92,3 +92,19 @@ The following table maps SPDM VDM command codes to Caliptra common commands. For
 | `0x12`       | Authorized Command           | O   | Execute an authorized sub-command via challenge-response MAC. Sub-commands: Get Auth Challenge (`0x01`), Program Field Entropy (`0x02`). |
 
 R = Required, O = Optional
+
+## Debug Unlock Payloads
+
+`Request Debug Unlock (0x0A)` uses a VDM-specific compact payload, not the
+MCI mailbox layout:
+
+- Request: `unlock_level` (`u8`).
+- Response: completion code (`u8`), unique device identifier (`u8[32]`),
+  then challenge (`u8[48]`).
+
+`Authorize Debug Unlock Token (0x0B)` carries a complete Caliptra RT mailbox
+request as its VDM payload. The first word is `MailboxReqHeader.checksum`,
+followed by the token request body described in
+[Caliptra Common Commands](caliptra_common_commands.md#authorize-debug-unlock-token).
+The MCU preserves these payload bytes exactly and forwards them to the Caliptra
+mailbox; the response is only the VDM completion code.
--- a/docs/src/caliptra_spdm_vdm_cmds.md
+++ b/docs/src/caliptra_spdm_vdm_cmds.md
@@ -4,7 +4,7 @@
 
 This document describes how Caliptra external commands are transported over SPDM Vendor Defined Messages (VDM) via MCTP. This is the **out-of-band (OOB)** path for accessing Caliptra device management commands from an external agent such as a BMC.
 
-For command definitions (categories, payloads, and completion codes), see [Caliptra Common Commands](caliptra_common_commands.md). SPDM VDM payloads generally use those command payload definitions directly and do not include Caliptra RT mailbox `MailboxReqHeader` or checksum bytes. `Authorize Debug Unlock Token` is the exception: its VDM payload is a host-prebuilt Caliptra RT mailbox request, including `MailboxReqHeader.checksum`, so the MCU can stream the request bytes unchanged into the Caliptra mailbox FIFO.
+For command definitions (categories, payloads, and completion codes), see [Caliptra Common Commands](caliptra_common_commands.md).
 
 For the unified software architecture shared between OOB (SPDM VDM) and in-band (MCI Mailbox) paths, see [Unified Caliptra Command Handling](unified_caliptra_command_handling.md).
 
--- a/docs/src/caliptra_spdm_vdm_cmds.md
+++ b/docs/src/caliptra_spdm_vdm_cmds.md
@@ -92,19 +92,3 @@ The following table maps SPDM VDM command codes to Caliptra common commands. For
 | `0x12`       | Authorized Command           | O   | Execute an authorized sub-command via challenge-response MAC. Sub-commands: Get Auth Challenge (`0x01`), Program Field Entropy (`0x02`). |
 
 R = Required, O = Optional
-
-## Debug Unlock Payloads
-
-`Request Debug Unlock (0x0A)` uses a VDM-specific compact payload, not the
-MCI mailbox layout:
-
-- Request: `unlock_level` (`u8`).
-- Response: completion code (`u8`), unique device identifier (`u8[32]`),
-  then challenge (`u8[48]`).
-
-`Authorize Debug Unlock Token (0x0B)` carries a complete Caliptra RT mailbox
-request as its VDM payload. The first word is `MailboxReqHeader.checksum`,
-followed by the token request body described in
-[Caliptra Common Commands](caliptra_common_commands.md#authorize-debug-unlock-token).
-The MCU preserves these payload bytes exactly and forwards them to the Caliptra
-mailbox; the response is only the VDM completion code.
--- a/platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/mod.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/mod.rs
@@ -184,7 +184,7 @@ impl CaliptraCmdHandler for CaliptraCmdBackend {
 
         let mut resp_buf = [0u8; core::mem::size_of::<ProductionAuthDebugUnlockChallenge>()];
 
-        let resp_len = execute_mailbox_cmd(
+        execute_mailbox_cmd(
             &mailbox,
             CommandId::PRODUCTION_AUTH_DEBUG_UNLOCK_REQ.0,
             req.as_mut_bytes(),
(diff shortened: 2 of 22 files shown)


docs/src/caliptra_spdm_vdm_cmds.md
```diff
@@ -4,7 +4,7 @@
 
 This document describes how Caliptra external commands are transported over SPDM Vendor Defined Messages (VDM) via MCTP. This is the **out-of-band (OOB)** path for accessing Caliptra device management commands from an external agent such as a BMC.
 
-For command definitions (categories, payloads, and completion codes), see [Caliptra Common Commands](caliptra_common_commands.md).
+For command definitions (categories, payloads, and completion codes), see [Caliptra Common Commands](caliptra_common_commands.md). SPDM VDM payloads generally use those command payload definitions directly and do not include Caliptra RT mailbox `MailboxReqHeader` or checksum bytes. `Authorize Debug Unlock Token` is the exception: its VDM payload is a host-prebuilt Caliptra RT mailbox request, including `MailboxReqHeader.checksum`, so the MCU can stream the request bytes unchanged into the Caliptra mailbox FIFO.
```
mlvisaya: nit: i think rather than having an exception here, it's probably better to just add the checksum as part of the fields in the caliptra_commands for Authorize Debug Unlock

parvathib:
Could you please rebase this PR on top of #1710? I have made some corrections to Certstore that would be conducive to your changes. 

Also, this branch seems to be adding `.text + .rodata` than expected. `DebugUnlock` streaming is required in production, so we should keep that enabled by default, but can you take a pass to reduce the footprint with these guidelines?

-  Make  CHUNK_SEND classify and route only. From the first chunk, identify whether the large request is SET_CERTIFICATE  or Caliptra  DebugUnlock  VDM, then delegate streaming to the  command-specific handler.

-  Feature-gate generic buffered request reassembly. By default, Caliptra should only stream the supported large requests and fail closed for unsupported ones. Keep buffered large-request handling only for tests/future generic use if needed.

-  Reduce async/generic growth. Keep parsing/sequencing helpers sync where possible; only make start/continue/finish command calls async.
Please check whether `process_chunk_send` can keep chunk parsing/sequencing sync and only await the command stream operations. The current async/generic helper path may be increasing   `.text `; please confirm with **size-history** after the change

-  Avoid unrelated generic expansion. Do not add broader generic VDM/request streaming framework beyond the currently supported large request paths.

runtime/userspace/syscall/src/mailbox.rs
```diff
@@ -123,6 +171,27 @@ impl<S: Syscalls> Mailbox<S> {
         }
     }
 
+    /// Initiates a guarded chunked mailbox request.
+    ///
+    /// The returned request holds the process-local mailbox mutex until it is
+    /// executed, aborted, or dropped.
+    pub async fn start_guarded_chunked_request(
```
mlvisaya: i think there is a possible denial of service vulnerability if we lock the mailbox here throughout the whole streaming sequence (that is, we pass the locked mutex to spdm, and release it after it's done)

the remote user could stop streaming the command halfway, and the mailbox will be forever locked. this can block other tasks from communicating through caliptra mailbox.

docs/src/caliptra_spdm_vdm_cmds.md
```diff
@@ -92,3 +92,19 @@ The following table maps SPDM VDM command codes to Caliptra common commands. For
...
+## Debug Unlock Payloads
+
+`Request Debug Unlock (0x0A)` uses a VDM-specific compact payload, not the
+MCI mailbox layout:
+
+- Request: `unlock_level` (`u8`).
+- Response: completion code (`u8`), unique device identifier (`u8[32]`),
+  then challenge (`u8[48]`).
+
+`Authorize Debug Unlock Token (0x0B)` carries a complete Caliptra RT mailbox
```
parvathib: Can we move the `Authorize Debug Unlock Token (0x0B)` payload description from the SPDM VDM doc into this common command section, and clarify that this is the uniform command body for both MCI mailbox and SPDM VDM? The leading `checksum` field is part of the common command body itself, not an SPDM-specific wrapper.

docs/src/caliptra_spdm_vdm_cmds.md
```diff
@@ -92,3 +92,19 @@ The following table maps SPDM VDM command codes to Caliptra common commands. For
 | `0x12`       | Authorized Command           | O   | Execute an authorized sub-command via challenge-response MAC. Sub-commands: Get Auth Challenge (`0x01`), Program Field Entropy (`0x02`). |
 
 R = Required, O = Optional
+
+## Debug Unlock Payloads
```
parvathib: Can we remove this extra `Debug Unlock Payloads` section from the SPDM VDM doc? 
The command body should be uniform across MCI mailbox and SPDM VDM transports and should be documented in the common command doc. 
We can have a follow up PR on making implementation uniform we can handle that in a follow-up PR, but this doc should not introduce a SPDM-specific Debug Unlock payload definition. 
This should probably be tracked as part of unified command handler PR.

platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/mod.rs
parvathib, line 1: This looks like a legacy/common-handler path rather than the active SPDM VDM or MCI mailbox debug-unlock path. Since we already have a unified command handler work item, can we track cleaning this up there?

runtime/kernel/capsules/src/mailbox.rs
```diff
@@ -317,6 +317,24 @@ impl<'a, A: Alarm<'a>> Mailbox<'a, A> {
...
+        // Only allowed after initiate_request (command 2).
+        if self.state.get() != MailboxState::Initiated {
+            return Err(ErrorCode::INVAL);
+        }
+        // Verify that the caller is the app that initiated the request.
+        if self.current_app.get() != Some(processid) {
+            return Err(ErrorCode::INVAL);
+        }
+        let _ = self.alarm.disarm();
+        self.driver.map(|driver| {
```
parvathib: I know this is a rare possibility, but don't disarm the alarm / reset state to Idle before confirming `driver.abort_request()` actually ran. 
If `self.driver.map(...)` returns `None`, this silently disables the timeout fallback and leaves the mailbox stuck forever. (mailbox capsule state says idle, but the hw mailbox lock is not cleared). Return an error in case map returns None instead and leave the alarm/state untouched.
On timeout, the state gets automatically cleared to Idle and HW lock is also released. 
Only reset state on confirmed success; otherwise leave state at Initiated so the timeout can clean it up.

runtime/userspace/api/spdm-lib/stack/src/set_certificate.rs
```diff
@@ -122,6 +122,265 @@ pub(crate) async fn handle_set_certificate_request<Pal: SpdmPal>(
...
 
+#[derive(Copy, Clone)]
+pub(crate) struct SetCertificateStreamState {
+    slot_id: u8,
+    key_pair_id: u8,
+    cert_model: u8,
+    root_hash: [u8; SHA384_DIGEST_SIZE],
+    der_len: usize,
+    der_received: usize,
+    data_checksum: u32,
```
parvathib: `data_checksum` shouldn't be computed in the SPDM stack at all — it's a storage-integrity concern specific to this PAL's flash format ( ManagedRecord ), not part of the SPDM protocol. It's never compared against anything the host sent or asserted. 
So, instead:
1. Remove `data_checksum` from  `SetCertificateStreamState` and from the `finish_write_cert_chain_stream / finish_stream_update` trait/PAL signatures entirely.
2. Have the PAL compute it itself in `finish_stream_update` , by reading back the flash it just wrote — the exact same pattern  `stored_checksum()` already uses in `load()`.

runtime/userspace/api/spdm-lib/pal/src/cert/endorsement.rs
```diff
@@ -395,6 +383,94 @@ impl ManagedEndorsement {
...
+    }
+
+    pub async fn finish_stream_update(
+        mut self,
+        algo: SpdmPalAsymAlgo,
+        key_pair_id: u8,
+        cert_info: u8,
+        root_hash: &[u8; 48],
+        data_len: usize,
+        data_checksum: u32,
```
parvathib: `finish_stream_update` should not blindly persist the caller-supplied  `data_checksum` value. 
Instead, it should compute the checksum from the data actually written to flash (the same way `stored_checksum()` already does load ()` ) and store that computed value, rather than trusting what the caller passes in.

runtime/userspace/api/spdm-lib/pal/src/cert/mod.rs
```diff
@@ -153,13 +213,12 @@ impl<M: MeasurementProvider> SpdmPalCertStore for McuSpdmPal<M> {
...
         cert_model: u8,
-        root_hash: &[u8; 48],
-        cert_chain: &[u8],
+        _root_hash: &[u8; 48],
+        _cert_chain: &[u8],
     ) -> McuResult<()> {
         if cert_model != CERT_MODEL_ALIAS_CERT {
             return Err(INVARIANT);
         }
-        validate_root_hash(self, root_hash, cert_chain).await?;
```
parvathib, line 162: why did you remove this? For a SET_CERTIFICATE request small enough to be handled without streaming,  `root_hash` is now never validated. right?

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
```diff
@@ -290,25 +369,304 @@ fn process_next_chunk<Pal: SpdmPal>(
...
+    secure_session: bool,
+) -> SpdmResult<bool> {
+    let (hdr, body) = SpdmMsgHdrPdu::ref_from_prefix(first).map_err(|_| SPDM_INVALID_REQUEST)?;
+    if hdr.version != state.version.to_u8() {
+        return Err(SPDM_INVALID_REQUEST);
+    }
+    match hdr.code {
+        #[cfg(feature = "set-certificate")]
+        ReqRespCode::SET_CERTIFICATE => {
+            if secure_session {
```
parvathib: Per SPDM spec, SET_CERTIFICATE and CHUNK_SEND are both allowed within a secure session. Why does this unconditionally reject chunked SET_CERTIFICATE inside a session as Unsupported ?

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
parvathib: we should remove this rejection in secure session condition too for consistency

runtime/userspace/api/spdm-lib/stack/src/chunk/mod.rs
```diff
@@ -6,6 +6,7 @@ mod get;
 mod send;
 
 pub(crate) use get::handle_chunk_get;
+pub(crate) use send::abort_streaming_request_if_needed;
```
parvathib: can we rename this function to `abort_active_streaming_request`?

runtime/userspace/api/spdm-lib/pal/src/cert/mod.rs
ArthurHeymans: The buffered path called validate_set_certificate_chain, but the MCU PAL implementation only checked the certificate model and ignored root_hash.

Fixed in https://github.com/chipsalliance/caliptra-mcu-sw/pull/1694/changes/4c6654c72443763e80850e24cf891798c7568a4c: the PAL now hashes the first DER certificate and compares it with the supplied SHA-384 root hash before write_cert_chain. The streamed path performs the equivalent validation from flash before committing metadata. Thanks for catching this.

platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/mod.rs
ArthurHeymans: I reverted this legacy-handler change in 964834bef and left its cleanup for the unified command-handler work item. Exact response-length validation remains in the active caliptra-api-lite/src/debug_unlock.rs path.

parvathib (approved):

PR #1695, opened 2026-06-25
[feat] Add DOT recovery reset coordination flow

### LINKED ISSUE #1597: DOT Recovery

mhatrevi:
- Add MCU ROM DOT recovery status path for locked DOT blob empty/corrupt/HMAC failures.
- Publish DEVICE_STATUS_0=0x94000E before reporting FW_ERROR_FATAL.
- Add RESET_CTRL decoding for 0x10/0x11 on the next cold boot.
- 0x10 forces fused owner PK hash; 0x11 continues regular DOT verification.
- Preserve Caliptra Core ROM ownership of firmware owner-auth failure reporting.
- Document reset coordination flow and status encoding in dot_i3c.md.


mlvisaya (approved):
LGTM, although I think the integ test doesn't really test some of the changes added because dot_recovery_reset_flow is not set to true.

mhatrevi:
> LGTM, although I think the integ test doesn't really test some of the changes added because dot_recovery_reset_flow is not set to true.

Added a test for the same.

PR #1707, opened 2026-06-26
[main-2.1] Fix MCTP control version discovery

parvathib:
Cherry-pick of #1702 to main-2.1.

This fixes MCTP control discovery consistency, reports version support for advertised message types, and uses IANA VDM 0x7F with the OCP Vendor ID for Caliptra VDM.


mlvisaya (approved):

helloxiling (approved):
Looks good.
Just leave a note: We'll continue cleaning up external_mctp_vmd_cmds.md to reflect the decision to use SPDM VDM instead of MCTP VDM (0x7F) for common commands. We'll also clarify that MCTP VDM (0x7E) remains available for PCI vendor/integrator extensions, but is not included in the MCU SDK by default. These updates can be made in follow-up PRs.

PR #1709, opened 2026-06-26
fix: enable test_firmware_update_streaming_fpga test

mlvisaya:
- Fix incorrect MCU staging address: use external staging SRAM (0xB00C_0000 + offset) instead of MCU SRAM base
- Fix mailbox timeout: increase from 100K to 200M ticks (10s at 20MHz) to avoid timeout during large chunked mailbox transfers
- Increase DMA transfer buffer from 128 bytes to 1024 bytes when copying firmware from staging memory to DMA staging area
- Refactor copy_to_memory into FirmwareUpdateHooks trait (transfer_to_dma_staging) so integrators can override the transfer buffer size via const generic default_copy_to_memory<N>
- Add progress logging for DMA copy and mailbox payload stream reads
- Add test-firmware-v2 feature and wire it through all platforms
- Point streaming FPGA test to use test-firmware-v2 PLDM package


runtime/userspace/api/caliptra-api/src/firmware_update/mod.rs
```diff
@@ -588,6 +586,15 @@ impl<'a, D: DMAMapping> FirmwareUpdater<'a, D> {
         let mut payload_stream =
             MailboxPayloadStream::new(self.staging_memory, image_offset, image_len);
 
+        console_writeln!(
+            Console::<DefaultSyscalls>::writer(),
+            "[FW Upd] process_caliptra_fw: action={:?}, offset={}, len={}, cmd={:#x}",
```
parvathib: - This print is in shared library code and will be present in the default path, so I’d remove it’s behind a tighter test cfg/feature.
- Also, {:?} can add to code size via Debug fmt. It is best avoided.
- This task seems to be feature gated. So, the Size history doesn't show any difference. It is good to keep an eye on the app-size changes, so we don't have surprises when we enable all.

runtime/userspace/api/caliptra-api/src/firmware_update/mod.rs
```diff
@@ -718,33 +726,26 @@ impl<'a, D: DMAMapping> FirmwareUpdater<'a, D> {
...
+                .transfer_to_dma_staging(
+                    self.staging_memory,
+                    self.dma_mapping,
+                    mem_address,
+                    offset,
+                    img_size,
+                )
+                .await
+        } else {
+            default_copy_to_memory::<128>(
```
parvathib: I don’t see the DMA transfer buffer size actually changed to 1024 here — copy_to_memory() still calls default_copy_to_memory::<128>(), and the hook default also uses 128. If the intent is to increase it for FPGA, should this be overridden in the FPGA hook implementation instead?

parvathib (commented):
LG, a few follow up questions.

runtime/userspace/api/caliptra-api/src/firmware_update/mod.rs
mlvisaya: this can removed. It's just a debug message I forgot to remove

runtime/userspace/api/caliptra-api/src/firmware_update/mod.rs
mlvisaya: my mistake, the committed file didn't have the intended 1024 change I have in my local. i'll update

parvathib (approved):

PR #1710, opened 2026-06-28
Use chunked CertifyKey for DPE leaf cert

### LINKED ISSUE #1667: [SPDM-CertStore] Add DPE leaf certificate chunking support
Add support for retrieving DPE leaf certificate in chunks using `CertifyKeyChunks` command in CertStore.

parvathib:
Fixes #1667

### DIFF
--- a/runtime/userspace/api/spdm-lib/pal/src/cert/mod.rs
+++ b/runtime/userspace/api/spdm-lib/pal/src/cert/mod.rs
@@ -23,6 +23,8 @@ use mcu_caliptra_api_lite::{
     dpe_certify_key_cert_size, dpe_certify_key_cert_slice, dpe_get_cert_chain_chunk,
     dpe_sign_ecc_p384, walk_dpe_chain, DpeChainSink, DPE_LABEL_LEN, DPE_MAX_CHUNK_SIZE,
 };
+#[cfg(feature = "set-certificate")]
+use mcu_caliptra_api_lite::{sha_finish, sha_init, sha_update, HashAlgo, SHA_CONTEXT_SIZE};
 use mcu_error::codes::{INTERNAL_BUG, INVARIANT};
 
 /// 48-byte label fed to DPE `CertifyKey`. Keep this stable so DPE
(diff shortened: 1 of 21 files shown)


mlvisaya:
I will look into the test_firmware_update_flash_successful_fast CI failure

mlvisaya:
Firmware update test fixes: https://github.com/chipsalliance/caliptra-mcu-sw/pull/1709

mlvisaya (approved):

runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
```diff
@@ -180,3 +117,107 @@ impl SharedCertStore {
...
+    chain_len: Option<u32>,
+    leaf_len: Option<u32>,
+    chain_digest: Option<[u8; 48]>,
+}
+
+/// Per-task cert store wrapper.
+///
+/// Wraps a reference to the global `'static SharedCertStore` alongside
+/// task-local caches (lengths, digests, etc.) to ensure complete task isolation.
+pub struct TaskCertStore {
```
helloxiling, line 132: The cert slots are shared globally via `CERT_STORE`, but these derived caches are per `TaskCertStore` / per `McuSpdmPal`. Since both MCTP and DOE pass the same `&CERT_STORE` but get separate `TaskCertStore`, a `SET_CERTIFICATE` over one transport can update or erase a shared slot, while the other transport keeps serving a stale cached chain length or digest. Should we consider cert cache invalidation across tasks?

runtime/userspace/api/spdm-lib/pal/src/cert/mod.rs
```diff
@@ -105,37 +120,26 @@ impl<M: MeasurementProvider> SpdmPalCertStore for McuSpdmPal<M> {
...
     }
 
+    #[cfg(feature = "set-certificate")]
     async fn validate_set_certificate_chain(
         &self,
         _io: &Self::Io<'_>,
         _slot: u8,
         _key_pair_id: u8,
-        _cert_model: u8,
+        cert_model: u8,
```
helloxiling, line 155: This ignores `root_hash` and `cert_chain`, but the trait contract says the PAL must verify `SHA-384(first DER certificate) == root_hash`. Is it inconsistent?

mhatrevi (approved):

runtime/userspace/api/spdm-lib/pal/src/cert/mod.rs
parvathib: Good catch. Fixed : the PAL now verifies the SPDM root hash against the first DER certificate:  `SHA-384(first DER cert) == root_hash` . It also keeps Caliptra’s  `AliasCert`  policy check. We should extend this validation beyond the root-hash check; the specific certificate attributes/policy checks will depend on the integrator requirements.

runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
parvathib: I agree with the concern. 
The stale-cache risk is mainly in provisioning: `SET_CERTIFICATE write/erase` can change shared slot state while each transport still has its own derived cache.

I’m thinking of handling this with a `per-managed-slot generation counter`: provisioning bumps the counter, and cached values are used only if their recorded counter still matches. Otherwise, the transport recomputes lazily. I’ll handle that synchronization in a separate PR since it touches several files.

Longer term, staging provisioning into a backup copy and publishing atomically may make this state transition cleaner.

PR #1714, opened 2026-06-30
feat: reserve persistent storage region in MCU SRAM

### LINKED ISSUE #1670: Reserved measurement storage SRAM carve-out
**caliptra-mcu-sw / firmware-bundler + platform startup:** add a configurable reserved SRAM region for attestation measurement storage and partition it into non-overlapping `DPE Handle Storage` and `Software PCR Storage` subregions. Provide a default reference reservation and allow integrators to configure size/capacity from SoC inventory. The region must be outside kernel/app RAM and startup zeroing and must survive **FW_HITLESS_UPD_RESET**.

mlvisaya:
Fixes #1670 

Carve out a configurable storage region at the end of MCU SRAM for attestation records and other data that must survive across firmware updates.

Linker/bundler changes:
- Add storage_size field to Platform manifest (TOML) and McuMemoryMap.
- Generate a 'storage' memory region and .storage section in the kernel linker script, with _sstorage/_estorage symbols for runtime access.
- Set storage_size = 0x400 (1 KB) in all emulator and FPGA manifests.

ROM changes:
- Compute mcu_fw_sram_exec_region_size excluding the storage region so the exec-region lock does not cover it.

Runtime changes:
- Declare _sstorage/_estorage extern symbols in emulator and FPGA board.rs for kernel-side access.

Firmware update test (test-firmware-activate):
- Add a new firmware update path that skips the PLDM download and goes directly to MCU image activation, allowing faster iteration on the activate + hitless-update flow.
- Write a monotonically increasing test pattern to the storage region before activation (test-firmware-activate) and verify it persists after the hitless update (test-firmware-v2).


firmware-bundler/reference/emulator/user-app.toml
```diff
@@ -10,6 +10,7 @@ name = "emulator-user-app"
 tuple = "riscv32imc-unknown-none-elf"
 dynamic_sizing = true
 page_size = 256
+storage_size = 0x400
```
parvathib, line 16: Nit:  `storage_size`  should be configured per platform based on the number of configured SoC TCB and non-TCB components and the record sizes used by the DPE Handle Storage and  Software PCR Storage capsules.  0x400  may be fine as a small reference/default, but can we add a short note in the manifest or docs that integrators must size this for their platform  inventory? The follow-up capsule PRs (#1722, #1723) can own the exact sizing formula/validation

firmware-bundler/src/ld.rs
```diff
@@ -525,14 +526,23 @@ INCLUDE $BASE_LD_CONTENTS
...
         sub_map.insert("DATA_RAM_START", format!("{:#x}", kernel_data.offset));
-        sub_map.insert("DATA_RAM_LENGTH", format!("{:#x}", dtcm.size));
+        sub_map.insert(
+            "DATA_RAM_LENGTH",
+            format!("{:#x}", dtcm.size - storage_size),
+        );
 
         let app_offset = kernel_data.offset + kernel_data.size;
-        let app_length = dtcm.size - kernel_data.size;
+        let app_length = dtcm.size.saturating_sub(kernel_data.size + storage_size);
```
parvathib, line 550: Could we make `storage_size` 4 KB aligned and validate it before generating the linker layout? 

ROM programs  `mcu_fw_sram_exec_region_size`  in 4 KB units, so this keeps that value and  the linker storage  region consistent and avoids implicit unused slack for sub-4 KB values.

Failing earlier on  `storage_size` validation also gives integrators a clear layout error.

platforms/fpga/runtime/src/board.rs
parvathib, line 379: Can we add a separate  PlatformRegion  for  _sstorage.._estorage  after this SRAM data region? Something like:

```
 platform_regions.push(PlatformRegion {
     start_addr: addr_of!(_sstorage),
     size: addr_of!(_estorage) as usize - addr_of!(_sstorage) as usize,
     is_mmio: false,
     user_accessible: false,
     read: true,
     write: true,
     execute: false,
 });
```
 This makes the persistent storage range a kernel-only RW region, separate from app RAM

platforms/emulator/runtime/src/board.rs
parvathib, line 368: similarly for emulator:

 Can we add a separate  PlatformRegion  for  _sstorage.._estorage  after this SRAM data region? Something like:
```
 platform_regions.push(PlatformRegion {
     start_addr: addr_of!(_sstorage),
     size: addr_of!(_estorage) as usize - addr_of!(_sstorage) as usize,
     is_mmio: false,
     user_accessible: false,
     read: true,
     write: true,
     execute: false,
 });
```
 This makes the persistent storage range a kernel-only RW region, separate from app RAM.

firmware-bundler/src/manifest.rs
```diff
@@ -192,6 +192,11 @@ pub struct Platform {
     /// Data memory, outside of the RAM used by the application.  This is used for the
     /// _pic_vector_table on VeeR chips.  Defaults to a size and offset of 0 if not defined.
     pub dccm: Option<Memory>,
+
+    /// Size in bytes of the persistent storage region reserved at the end of SRAM.
```
parvathib, line 196: Can we expand this field comment to say  `storage_size`  is platform-dependent and should be chosen based on the configured SoC TCB/non-TCB component counts and the DPE Handle Storage /  Software PCR Storage record sizes? The exact sizing formula can come with the capsule integration, but integrators should know this is not a universal default.

firmware-bundler/reference/emulator/user-app.toml
mlvisaya: Done — added a comment above the field in  pointing integrators to the sizing formula, and expanded the  field doc. Also added a dedicated **MCU SRAM Partitioning** section to the integrator guide with the layout diagram, per-capsule sizing formulas, and a manifest snippet showing exactly which field to edit.

firmware-bundler/src/ld.rs
mlvisaya: Done — `storage_size` is now rounded up to a 4 KiB boundary (`(storage_size + 0xFFF) & !0xFFF`) and validated to be strictly less than `dtcm.size` before the linker layout is generated, so integrators get a clear error instead of silent truncation.

platforms/fpga/runtime/src/board.rs
mlvisaya: Done — added a separate `PlatformRegion` for `_sstorage.._estorage` after the SRAM data region on both FPGA and emulator. The push is guarded by a size check (`_sstorage != _estorage`) so it is a no-op when `storage_size = 0`.

platforms/emulator/runtime/src/board.rs
mlvisaya: Done — same change applied to the emulator board. The emulator SRAM data region still ends at `_esram` (no `+ 0x80` extension), and the storage region is covered by its own dedicated `PlatformRegion`. Also bumped the `ArrayVec` capacity from 9 to 10 to accommodate the extra entry.

firmware-bundler/src/manifest.rs
mlvisaya: Done — expanded the `storage_size` field comment to explain it is platform-dependent, must be sized based on the configured SoC TCB/non-TCB component counts and the DPE Handle Storage / Software PCR Storage record sizes, and that the firmware-bundler rounds the value up to a 4 KiB boundary automatically.

mlvisaya: Done. Added detailed comment for storage_size

parvathib (approved):
LGTM

PR #1717, opened 2026-06-30
Fix FE_PROG core response handling

### LINKED ISSUE #1513: Response error in the ZEROIZE_UDS_FE command
Regarding the response size of mailbox command of ZEROIZE_UDS_FE, we found a mismatch between Core side and SS side.
This causes a response error for the ZEROIZE_UDS_FE command.

The response data is defined in Core side as follows:
`pub struct ZeroizeUdsFeResp {`
`    pub hdr: MailboxRespHeader,`
`    pub dpe_result: u32,`
`}`
https://github.com/chipsalliance/caliptra-sw/blob/rom-2.1.1/api/src/mailbox.rs#L5586-L5589

However, SS side expects the response data to only MailboxRespHeader part:
`                let mut resp_buf = [0u8; core::mem::size_of::<MailboxRespHeader>()];`
`                if let Err(err) = soc_manager.finish_mailbox_resp_bytes(&mut resp_buf) {`
https://github.com/chipsalliance/caliptra-mcu-sw/blob/rom-sdk-2.1.0rc1/rom/src/cold_boot.rs#L205-L206

We believe that this error can be fixed by making the following changes:
`- let mut resp_buf = [0u8; core::mem::size_of::<MailboxRespHeader>()];`
`+ let mut resp_buf = [0u8; core::mem::size_of::<caliptra_api::mailbox::ZeroizeUdsFeResp>()];`

ArthurHeymans:
## Summary
- Size Core `FE_PROG` / `ZEROIZE_UDS_FE` response buffers for `MailboxRespHeader + dpe_result` instead of header-only responses.
- Consume and checksum the full Core response, and fail when `dpe_result != 0`.
- Preserve legacy 8-byte header-only response compatibility while rejecting malformed partial `dpe_result` responses.

Fixes #1513.

### DIFF
--- a/platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/mod.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/mod.rs
@@ -137,10 +137,14 @@ impl CaliptraCmdHandler for CaliptraCmdBackend {
     }
 
     async fn program_field_entropy(&self, partition: u32) -> CaliptraCmdResult<()> {
-        use caliptra_api::mailbox::{CommandId, FeProgReq};
+        use caliptra_api::mailbox::{CommandId, FeProgReq, MailboxRespHeader};
         use caliptra_mcu_libapi_caliptra::mailbox_api::execute_mailbox_cmd;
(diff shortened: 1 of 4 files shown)


ArthurHeymans:
actually fixed in #1517 already....

PR #1722, opened 2026-06-30
feat: DPE handle store capsule, syscall layer, and integration test

### LINKED ISSUE #1672: DPE Handle Storage capsule, syscall driver, and libtock API
**caliptra-mcu-sw / Tock capsule + libtock syscall API:** implement `DPE handle record storage` Tock capsule backed by the DPE Handle Storage subregion. 
Expose `record read/write, clear, leaf read, attestation target mark/read` operations.

mlvisaya:
Fixes #1672 

## Summary

Implements the DPE Handle Store driver for managing DPE context handle records in the reserved persistent SRAM region (backed by `_sstorage`..`_estorage` linker symbols generated by the firmware-bundler).

### Kernel capsule (`SyscallDriver`)
- `runtime/kernel/capsules/src/dpe_handle_store.rs`: Tock capsule backed by a raw `&'static mut [u8]` slice. Supports 6 commands at driver number `0x8000_0020`:
  - `EXISTS` (0), `READ_RECORD` (1), `WRITE_RECORD` (2, upsert-by-fw_id), `CLEAR_RECORDS` (3), `READ_LEAF_RECORD` (4), `MARK_ATTESTATION_TARGET` (5), `READ_ATTESTATION_TARGET` (6)
- `runtime/kernel/components/src/dpe_handle_store.rs`: Component wrapper for board instantiation

### Userspace syscall layer
- `runtime/userspace/syscall/src/dpe_handle_store.rs`: `DpeHandleStore` API with `DpeHandleRecord` / `DpeHandleRecordFlags` types. Uses raw `syscall4` for allow_rw/allow_ro paths since driver_num is runtime-determined.

### Emulator board wiring
- `platforms/emulator/runtime/src/board.rs`:
  - Instantiates `DpeHandleStoreComponent` from the `_sstorage`..`_estorage` SRAM slice
  - Extends the kernel PMP data region end from `_esram + 0x80` to `_estorage` so the capsule can write to the full storage area

### Infrastructure fixes
- `common/config/src/lib.rs`: Increase `McuMemoryMap::default` `sram_size` from 512 KB to 1 MB to match the `user-app-devel.toml` firmware-bundler profile
- `emulator/consts/src/lib.rs`: Increase `RAM_SIZE` to 1 MB accordingly (was marked TEMPORARY at 512 KB)

### Integration test
- `tests/integration/src/test_dpe_handle_store.rs`: boots emulator with `test-dpe-handle-store` feature, asserts clean exit
- `platforms/emulator/runtime/userspace/apps/example/src/test_dpe_handle_store.rs`: exercises `exists`, `clear_records`, write/read records, upsert, leaf record tracking, `mark_attestation_target`, `read_attestation_target`, and post-clear verification


runtime/kernel/capsules/src/dpe_handle_store.rs
```diff
@@ -0,0 +1,504 @@
...
+//!
+//! ## Driver number
+//!
+//! `0x8000_0020`
+//!
+//! ## SRAM layout
+//!
+//! ```text
+//! offset 0 : record_count           u16 LE
+//! offset 2 : _pad                   u16
```
parvathib: Could we align this metadata layout with the DPE Handle Store header from the attestation capsule design? (https://github.com/chipsalliance/caliptra-mcu-sw/blob/pbhogaraju/attestation-arch/docs/src/attestation-tock-capsules.md#record-format)

Right now the capsule stores only  `record_count`  and   `attestation_target_fw_id` , but the hitless-update flow needs enough metadata to validate preserved state before using it. 
The documented header includes `magic ,  version ,   header_size ,  record_size ,  record_capacity ,  record_count ,  attestation_target_fw_id , and  attestation_policy_digest`. That would let the capsule/Measurement API confirm the SRAM contents match the expected layout and policy before serving preserved records.

runtime/kernel/capsules/src/dpe_handle_store.rs
```diff
@@ -0,0 +1,504 @@
...
+//! ```
+//!
+//! ## Syscalls (all synchronous)
+//!
+//! | Command | Name                    | Arg1   | Arg2     | Allow           |
+//! |---------|-------------------------|--------|----------|-----------------|
+//! | 0       | EXISTS                  | —      | —        | —               |
+//! | 1       | READ_RECORD             | fw_id  | reserved | RW 0 (output)   |
+//! | 2       | WRITE_RECORD            | fw_id  | reserved | RO 0 (input)    |
+//! | 3       | CLEAR_RECORDS           | —      | —        | —               |
```
parvathib: Can we replace `CLEAR_RECORDS`  with  `INITIALIZE_STORE(attestation_policy_digest)`  and add  `VALIDATE_STORE(attestation_policy_digest)` ? 

- `INITIALIZE_STORE` would be called on cold boot to initialize the record store: write the header, store the digest passed by user, clear the AK target, reset  record count , and zeroize record slots. 

- `VALIDATE_STORE` would be called on hitless update before any reads to validate the record store:  validate header,  record_count <= record_capacity , stored digest matches the digest by passed by use and AK target record.

runtime/kernel/capsules/src/dpe_handle_store.rs
```diff
@@ -0,0 +1,504 @@
...
+//! ```
+//!
+//! ## Record layout (`DPE_HANDLE_RECORD_SIZE` = 32 bytes)
+//!
+//! ```text
+//! offset  0 : fw_id            u32 LE
+//! offset  4 : parent_fw_id     u32 LE  (0xFFFF_FFFF = None)
+//! offset  8 : context_handle   [u8; 16]
+//! offset 24 : tci_tag          u32 LE
+//! offset 28 : flags            u8  (bit 0 = valid, bit 1 = attestation_target)
```
parvathib: I’m thinking we should simplify this from what the current design doc shows. 
Since the store is `append/upsert` and has no delete operation,  `record_count`  can be the source of truth for **valid** records. Similarly, `attestation_target_fw_id`  in the header can be the source of truth for the **attestation target**. 
Can we reserve the record flags byte for now instead of using it for `valid  /  attestation_target` ? That avoids having two sources of truth to keep in sync.

I've updated the documentation now accordingly. (https://github.com/chipsalliance/caliptra-mcu-sw/blob/pbhogaraju/attestation-arch/docs/src/attestation-tock-capsules.md#record-format)

platforms/emulator/runtime/src/board.rs
```diff
@@ -829,6 +836,24 @@ pub unsafe fn main() {
     )
     .finalize(external_otp_component_static!());
 
+    // DPE Handle Store: backed by the persistent storage SRAM reservation
+    // (_sstorage .. _estorage) defined by the firmware-bundler linker script.
+    // Capacity = (region_len - 8) / 32 records (≥ 31 with the default 1 KB region).
+    // When built outside the firmware-bundler (e.g. cargo check), _sstorage ==
+    // _estorage == 0 so capacity is 0 and no records can be stored, which is safe.
+    let dpe_handle_store = {
```
parvathib, line 858: This passes the entire ` _sstorage.._estorage` region to the DPE Handle Store. Since Software PCR Storage will also need part of the persistent SRAM reservation, can we partition this region?

parvathib (changes requested):

platforms/emulator/runtime/src/board.rs
mlvisaya: This will be addressed in PR #1723 (Software PCR Store), which subdivides the `_sstorage.._estorage` region between the DPE Handle Store (first `DPE_STORE_SIZE` bytes) and the Software PCR Store (remainder). The emulator `board.rs` there uses `split_at_mut(DPE_STORE_SIZE)` to partition the region before passing each slice to its respective capsule.

runtime/kernel/capsules/src/dpe_handle_store.rs
mlvisaya: Done — expanded the header from 8 to 72 bytes. The new layout matches the documented format: `magic` (0xD9E4_C7A1), `version` (1), `header_size`, `record_size`, `record_capacity`, `record_count`, `attestation_target_fw_id`, and `attestation_policy_digest` (48 bytes, SHA-384). `VALIDATE_STORE` reads all of these fields to confirm SRAM contents are consistent with the expected layout and policy before serving preserved records on hitless-update boot.

runtime/kernel/capsules/src/dpe_handle_store.rs
mlvisaya: Done — `CLEAR_RECORDS` has been removed and replaced with two new commands:

- `INITIALIZE_STORE(attestation_policy_digest)` (command 3): takes a 48-byte policy digest via `allow_readonly`. Writes the full 72-byte header (magic, version, sizes, capacity, count=0, attestation_target=SENTINEL, digest), then zeros all record slots.
- `VALIDATE_STORE(attestation_policy_digest)` (command 7): takes the same 48-byte digest. Validates magic, version, header/record sizes, record_count ≤ record_capacity, stored digest == provided digest, and that the attestation target fw_id (if set) is within bounds.

runtime/kernel/capsules/src/dpe_handle_store.rs
mlvisaya: Done — `FLAG_VALID` and the `attestation_target` flag have been removed from the per-record `flags` byte. `record_count` is now the sole source of truth for record validity: all records at indices `0..(record_count-1)` are valid by definition. The attestation target is tracked exclusively via `attestation_target_fw_id` in the header. The `flags` byte in each record is now reserved/unused.

runtime/kernel/capsules/src/dpe_handle_store.rs
```diff
@@ -0,0 +1,511 @@
...
+
+use core::cell::RefCell;
+use kernel::grant::{AllowRoCount, AllowRwCount, Grant, UpcallCount};
+use kernel::processbuffer::{
+    ReadableProcessBuffer, ReadableProcessSlice, WriteableProcessBuffer, WriteableProcessSlice,
+};
+use kernel::syscall::{CommandReturn, SyscallDriver};
+use kernel::{ErrorCode, ProcessId};
+
+pub const DPE_STORE_SIZE: usize = 0x400;
```
parvathib: This constant (`DPE_STORE_SIZE`) is a board/storage-layout split value, not a DPE capsule property.
Can we move it out of the capsule and define is as a constant in  `board.rs` near the `_sstorage/_estorage` split logic for now.  

In the long term, we shall provide a platform config file to define/compute these values. I'll add an issue for that.

platforms/emulator/runtime/src/board.rs
```diff
@@ -364,10 +369,12 @@ pub unsafe fn main() {
...
     });
 
-    // Data region (SRAM)
+    // Data region (SRAM) — extends to end of storage region so the kernel
+    // DPE handle store capsule (and any other kernel code) can write to
+    // the persistent storage area at the end of SRAM (_sstorage.._estorage).
     platform_regions.push(PlatformRegion {
         start_addr: addr_of!(_ssram),
-        size: (addr_of!(_esram) as usize + 0x80) - addr_of!(_ssram) as usize,
+        size: addr_of!(_estorage) as usize - addr_of!(_ssram) as usize,
```
parvathib: is extending the normal SRAM data region to ` _estorage`  still needed now that ` _sstorage.._estorage`  is also added as a separate kernel-only RW PMP region below? If not,  keeping the data region as  `_ssram.._esram`  and the storage region as  `_sstorage.._estorage`  would make the separation clearer.

runtime/kernel/capsules/src/dpe_handle_store.rs
mlvisaya: Done — `DPE_STORE_SIZE` has been removed from the capsule. The emulator `board.rs` now owns the storage-layout split with local constants (`DPE_STORE_SIZE = 0x400`, `PCR_STORE_SIZE = 0xC00`) defined near the `_sstorage/_estorage` partitioning logic. Noted on the longer-term platform config file — looking forward to that issue.

platforms/emulator/runtime/src/board.rs
mlvisaya: Done — the data PMP region now covers `_ssram.._esram` only. The persistent storage area is already protected by its own kernel-only RW entry (`_sstorage.._estorage`), so extending the data region to `_estorage` was redundant. The two regions are now non-overlapping and the separation is explicit.

parvathib (approved):

PR #1723, opened 2026-07-01
feat: Software PCR store capsule, syscall layer, and integration test

### LINKED ISSUE #1673: Software PCR Storage capsule, syscall driver, and libtock API
**caliptra-mcu-sw / Tock capsule + libtock syscall API:** implement current/journey PCR-style measurement records Tock capsule backed by the Software PCR Storage subregion. 
Expose `READ_MEASUREMENT, EXTEND_MEASUREMENT, and CLEAR_MEASUREMENTS. EXTEND_MEASUREMENT` updates current and extends journey following the Caliptra current/journey PCR model.

mlvisaya:
Fixes #1673 

Implements the Software PCR (Platform Configuration Register) storage driver for managing SHA-384 measurements in a reserved SRAM subregion. Built on top of the DPE handle store SRAM reservation: the 4 KB storage region is split as DPE [0x000..0x400) and PCR [0x400..0x1000).

Storage layout increase:
- firmware-bundler/reference/\*/\*.toml: storage_size 0x400 → 0x1000 (all emulator and FPGA profiles)

Kernel capsule (driver 0x8000_0021):
- runtime/kernel/capsules/src/pcr_store.rs: PcrStore backed by a &'static mut [u8] slice. Supports 32 SHA-384 PCR slots with a u32 valid-bitmask. Commands: EXISTS, READ_MEASUREMENT, WRITE_MEASUREMENT, CLEAR_MEASUREMENTS
- runtime/kernel/capsules/src/dpe_handle_store.rs: added DPE_STORE_SIZE constant (0x400) for board-level storage partitioning
- runtime/kernel/components/src/pcr_store.rs: PcrStoreComponent + pcr_store_component_static! macro

Userspace syscall layer:
- runtime/userspace/syscall/src/pcr_store.rs: PcrStore<S> with read_measurement, write_measurement, clear_measurements (synchronous) and extend_measurement (async: SHA-384(old||new) via Caliptra mailbox CM_SHA_INIT/CM_SHA_UPDATE/CM_SHA_FINAL, then write_measurement)
- runtime/userspace/syscall/Cargo.toml: added zerocopy dependency

Emulator board wiring:
- platforms/emulator/runtime/src/board.rs: splits _sstorage.._estorage into DPE (first DPE_STORE_SIZE bytes) and PCR (remaining bytes) subslices; instantiates PcrStoreComponent; adds pcr_store field and with_driver match arm

Integration test:
- tests/integration/src/test_pcr_store.rs: boots with feature test-pcr-store, asserts exit 0
- platforms/emulator/runtime/userspace/apps/example/src/test_pcr_store.rs: async test covering exists, write/read, overwrite, extend (SHA-384 via mailbox), clear, and out-of-range index error cases
- builder/src/features.rs: added test-pcr-store to EMULATOR_RUNTIME_TEST_FEATURES


builder/src/all.rs
```diff
@@ -69,6 +69,7 @@ const FEATURES_WITH_EXAMPLE_APP: &[&str] = &[
     "test-external-otp",
     "test-firmware-v2",
     "test-dpe-handle-store",
+    "test-pcr-store",
```
parvathib: to avoid confusion, can we rename this to `test_sw_pcr_store`?

runtime/kernel/capsules/src/dpe_handle_store.rs
```diff
@@ -58,6 +58,9 @@ use kernel::syscall::{CommandReturn, SyscallDriver};
 use kernel::{ErrorCode, ProcessId};
 
 /// Proposed driver number for the DPE Handle Storage capsule.
+/// Total bytes reserved for the DPE Handle Store subregion.
```
parvathib: This should be configured in the board files rather than here. Capsule can compute the capacity from the SRAM slice it received.
The DPE/PCR split should live in the board/platform storage layout code.

runtime/kernel/capsules/src/pcr_store.rs
```diff
@@ -0,0 +1,271 @@
+// Licensed under the Apache-2.0 license
+
+//! Software PCR (Platform Configuration Register) storage capsule.
+//!
+//! Manages up to `PCR_COUNT` (32) SHA-384 measurements in a reserved SRAM
```
parvathib: I don’t think we should hard-code `PCR_COUNT = 32`  in the capsule. 
The Software PCR store capacity should be derived from the assigned SRAM subregion, and the build/config should  validate that this capacity is sufficient for the configured non-TCB component count.

runtime/kernel/capsules/src/pcr_store.rs
```diff
@@ -0,0 +1,271 @@
...
+//! subregion.  All operations are **synchronous**; no upcalls are used.
+//!
+//! ## Driver number
+//!
+//! `0x8000_0021`
+//!
+//! ## SRAM layout
+//!
+//! ```text
+//! offset 0 : pcr_valid_mask  u32 LE  (bit i = PCR index i contains a value)
```
parvathib: This layout does not include a Soft PCR store header. 
For hitless update, we need to validate preserved Software PCR state before using it, similar to DPE Handle Storage. 
Can we add a   `SoftwarePcrStoreHeader`  as described in the document: https://github.com/chipsalliance/caliptra-mcu-sw/blob/pbhogaraju/attestation-arch/docs/src/attestation-tock-capsules.md#record-format-1

runtime/kernel/capsules/src/pcr_store.rs
parvathib, line 1: it might be a good idea to rename this to `soft_pcr_store` to avoid any confusion with hw pcr store.

runtime/kernel/capsules/src/pcr_store.rs
```diff
@@ -0,0 +1,271 @@
...
+//! ...
+//! offset 4 + i*48 : pcr[i]   [u8; 48]
+//! ```
+//!
+//! ## Syscalls (all synchronous)
+//!
+//! | Command | Name                 | Arg0        | Allow           |
+//! |---------|----------------------|-------------|-----------------|
+//! | 0       | EXISTS               | —           | —               |
+//! | 1       | READ_MEASUREMENT     | pcr_index   | RW 0 (output)   |
```
parvathib: These APIs should be passing `fw_id` instead of  `pcr_index`. 
The Measurement API routes non-TCB components to Soft PCR storage by fw_id , so the capsule API should match that (https://github.com/chipsalliance/caliptra-mcu-sw/blob/pbhogaraju/attestation-arch/docs/src/attestation-tock-capsules.md#userspace-syscall-api-1)

runtime/kernel/capsules/src/pcr_store.rs
```diff
@@ -0,0 +1,271 @@
...
+/// Driver number for the Software PCR Store.
+pub const DRIVER_NUM: usize = 0x8000_0021;
+
+/// Size of one PCR measurement (SHA-384 = 48 bytes).
+pub const PCR_MEASUREMENT_SIZE: usize = 48;
+
+/// Maximum number of PCR indices supported.
+pub const PCR_COUNT: usize = 32;
+
+/// Total bytes required in the SRAM subregion:
```
parvathib: these should not be hardcoded in the capsule. The capacity should be derived from the SRAM subregion size configured.

runtime/kernel/capsules/src/pcr_store.rs
```diff
@@ -0,0 +1,271 @@
...
+/// 4-byte valid mask + 32 × 48-byte slots = 1540 bytes.
+/// We reserve 0xC00 (3072) bytes so there is headroom.
+pub const PCR_STORE_SIZE: usize = 0xC00;
+
+const META_VALID_MASK: usize = 0; // u32 LE at [0..4]
+const META_SIZE: usize = 4;
```
parvathib: We should add actual store header (`SoftwarePcrStoreHeader `) instead of only  pcr_valid_mask  metadata.

We need magic/version, record size, capacity, and record count so the preserved Software PCR store can be validated on hitless update before use.   https://github.com/chipsalliance/caliptra-mcu-sw/blob/pbhogaraju/attestation-arch/docs/src/attestation-tock-capsules.md#record-format-1

runtime/kernel/capsules/src/pcr_store.rs
```diff
@@ -0,0 +1,271 @@
...
+mod rw_allow {
+    pub const OUTPUT: usize = 0;
+    pub const COUNT: u8 = 1;
+}
+
+mod cmd {
+    pub const EXISTS: usize = 0;
+    pub const READ_MEASUREMENT: usize = 1;
+    pub const WRITE_MEASUREMENT: usize = 2;
+    pub const CLEAR_MEASUREMENTS: usize = 3;
```
parvathib: Sorry to change this, but I think we should make the Software PCR command contract explicit and align it with the DPE storage model:

- Cold boot:  `INITIALIZE_STORE` 
- Hitless update:  `VALIDATE_STORE` 
- Initial non-TCB measurement record:  `CREATE_MEASUREMENT` 
- Component update:  `EXTEND_MEASUREMENT` , which must fail if the  fw_id  record does not already exist.

Since the SHA extend is done in userspace, `EXTEND_MEASUREMENT` can take the full updated `MeasurementRecord` after userspace reads the old record, computes the new current/journey values, and writes the updated record back. 

This keeps the cold-boot and hitless-update contracts explicit and avoids silently creating missing records during update flows.

parvathib (changes requested):

builder/src/all.rs
mlvisaya: Done — renamed to `test_sw_pcr_store` throughout (feature flag, module, test function, and integration test).

runtime/kernel/capsules/src/dpe_handle_store.rs
mlvisaya: Done — `DPE_STORE_SIZE` has been removed from the DPE capsule and `PCR_STORE_SIZE` was never exported from the new capsule. The emulator `board.rs` now owns both split constants (`const DPE_STORE_SIZE: usize = 0x400`, `const PCR_STORE_SIZE: usize = 0xC00`) and passes each capsule only its assigned subregion.

runtime/kernel/capsules/src/pcr_store.rs
mlvisaya: Done — `PCR_COUNT` and `PCR_STORE_SIZE` have been removed from the capsule. `record_capacity` is now derived at runtime from the SRAM slice length: `(sram_len - HEADER_SIZE) / MEASUREMENT_RECORD_SIZE`.

runtime/kernel/capsules/src/pcr_store.rs
mlvisaya: Done — added `SoftwarePcrStoreHeader` (16 bytes): `magic=0x5350_4352` ("SPCR"), `version=1`, `header_size=16`, `record_size=112`, `record_capacity` (derived from slice), `record_count`. `INITIALIZE_STORE` writes the header on cold boot; `VALIDATE_STORE` checks all structural fields plus `record_count <= record_capacity` on hitless-update boot.

runtime/kernel/capsules/src/pcr_store.rs
mlvisaya: Done — renamed to `soft_pcr_store` throughout: kernel capsule (`SoftPcrStore`), kernel component, userspace syscall module, and all file names.

runtime/kernel/capsules/src/pcr_store.rs
mlvisaya: Done — all APIs now take `fw_id: u32` instead of `pcr_index`. Each `MeasurementRecord` (112 bytes) carries `fw_id`, `current_digest[48]`, `journey_digest[48]`, `svn`, `version`, and `reserved[4]`.

runtime/kernel/capsules/src/pcr_store.rs
mlvisaya: Done — no hardcoded counts remain in the capsule. Capacity is derived from the SRAM slice at construction time (see `SoftPcrStore::derived_capacity`).

runtime/kernel/capsules/src/pcr_store.rs
mlvisaya: Done — replaced the single `pcr_valid_mask` u32 with a full `SoftwarePcrStoreHeader` (16 bytes) containing magic, version, header_size, record_size, record_capacity, and record_count. `VALIDATE_STORE` checks all structural fields on hitless-update boot.

runtime/kernel/capsules/src/pcr_store.rs
mlvisaya: Done — the command contract has been updated to align with the DPE model:

- **`INITIALIZE_STORE` (4)** — cold boot: writes the `SoftwarePcrStoreHeader`, resets `record_count` to 0, zeroizes all record slots.
- **`VALIDATE_STORE` (5)** — hitless-update: validates magic/version/sizes, `record_capacity` against the assigned SRAM slice, and `record_count <= record_capacity`.
- **`CREATE_MEASUREMENT` (2)** — initial load: stores a new `MeasurementRecord` for `fw_id`; fails with `ALREADY` if the record already exists.
- **`UPDATE_MEASUREMENT` (3)** — component update: updates the existing record for `fw_id`; fails with `FAIL` if `fw_id` is not found.
- **`READ_MEASUREMENT` (1)** — reads the record for `fw_id` into the RW allow buffer.

SHA-384 extend is done entirely in userspace; the capsule only stores and retrieves records.

platforms/emulator/runtime/src/board.rs
parvathib, line 1: This seems to be wiring the DPE Handle Store and Soft PCR Store only for the emulator board. Shouldn't we wire this to FPGA also?

mlvisaya: yes. updated now

parvathib (approved):

PR #1727, opened 2026-07-01
Grouping all nightly workflows into a single workflow.

v-focutter:
Grouping all the nightly workflows into a single workflow file to make it easier to manage and trigger. 

Can be triggered via the Github command line with "gh workflow run "Nightly Workflows" --ref v-focutter/nighty-workflow"


parvathib (approved):
LGTM

Issue #966, opened 2026-02-25
[OCP-LOCK]: Firmware flow for setting Perma HEK Fuse


clundin25:
Set the Fuse used by ROM to determine if the HEK perma state is set.

PR #1729, opened 2026-07-01
[OCP-LOCK]: Add mailbox command for setting hek perma state

clundin25:
https://github.com/chipsalliance/caliptra-mcu-sw/issues/966.

### DIFF
--- a/runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
+++ b/runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
@@ -22,8 +22,9 @@ use caliptra_mcu_mbox_common::messages::{
     FuseRevokeVendorPubKeyResp, FuseWriteReq, FuseWriteResp, GetAuthCmdChallengeReq,
     GetAuthCmdChallengeResp, GetLogReq, GetLogResp, MailboxRespHeader, MailboxRespHeaderVarSize,
     McuFeProgReq, McuResponseVarSize, OcpLockRotateHekReq, OcpLockRotateHekResp,
-    ProvisionVendorPkHashReq, ProvisionVendorPkHashResp, RevokeVendorPubKeyType, DEVICE_CAPS_SIZE,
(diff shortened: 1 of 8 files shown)


runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -926,6 +932,29 @@ impl<'a> CmdInterface<'a> {
...
+    async fn handle_ocp_lock_set_perma_hek<'r>(
+        &self,
+        req: &[u8],
+        resp_buf: &'r mut [u8],
+    ) -> McuResult<(&'r mut [u8], MbxCmdStatus)> {
+        let mut _full_req = OcpLockSetPermaHekReq::default();
+        if req.len() > size_of::<OcpLockSetPermaHekReq>() {
+            return Err(errors::INVALID_PARAMS);
+        }
+        _full_req.as_mut_bytes()[..req.len()].copy_from_slice(req);
```
zhalvorsen: Should we drop this if it isn't used?

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -926,6 +932,29 @@ impl<'a> CmdInterface<'a> {
...
+
+        let status = if self.otp.set_hek_perma().is_err() {
```
zhalvorsen, line 944: Should this check all slots are zeroized before setting this value?

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
clundin25: Oh whoops, yes!

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
clundin25: Good catch, yes this should

zhalvorsen (approved):

PR #1733, opened 2026-07-01
Switch reference command authorizer to use asymmetric keys

zhalvorsen:
The old reference used HMAC (symmetric) keys which are difficult to keep secure because they need to exist in both an HSM or similar infrastructure and be referenced by the firmware. Instead of giving an example of provisioning this key into fuses and worrying about key hygiene we have switched to using asymmetric keys so a public keys can be stored directly in firmware.

This example uses both ECDSA and ML-DSA to verify the authorization to maintain PQC safety.


RaunakGu:
hi @zhalvorsen,

I had previously filed a proposal for this use case, where the key is stored as part of the SoC Manifest, enabling a fully rooted chain of trust with Caliptra Core(Instead of MCU) with support for key replacement.

Can you please check this?

https://github.com/chipsalliance/caliptra-sw/issues/3928

zhalvorsen:
> hi @zhalvorsen,
> 
> I had previously filed a proposal for this use case, where the key is stored as part of the SoC Manifest, enabling a fully rooted chain of trust with Caliptra Core(Instead of MCU) with support for key replacement.
> 
> Can you please check this?
> 
> [chipsalliance/caliptra-sw#3928](https://github.com/chipsalliance/caliptra-sw/issues/3928)

I agree, we should have a better long term solution for these commands. In the meantime, I think it would be easier to for integrators to take this approach over the example symmetric key solution. I like your RFC option, there are some details I think still need to be ironed out, but lets collaborate on a final approach.

zhalvorsen:
Closing this to merge into `main` instead

PR #1734, opened 2026-07-02
Fix VirtualMuxAlarm timer reprogramming bug on VeeR EL2 FPGA

mlvisaya:
Replace Tock's VirtualMuxAlarm with a custom CaliptraVirtualAlarm that removes the within_range + next_tick_vals optimization. This optimization fails on VeeR EL2 because stale timer state after disable/re-enable cycles causes the guard to skip hardware reprogramming.

Changes:
- Add caliptra-mcu-virtual-alarm crate as drop-in replacement for capsules_core::virtualizers::virtual_alarm
- Update all imports across the codebase to use the new crate
- Add hardware timer polling (has_timer0_expired) in chip.rs to detect timer expiry when ISR can't run (mstatus.MIE=0 in kernel mode)
- Skip wfi when timer is armed since VeeR's wfi stalls the core clock and stops the internal timer counter
- Fix get_alarm() wrapping arithmetic so expired timers return past time
- Fix now() with race-free 64-bit read (mcycleh/mcycle retry loop)
- Remove ProcessConsole (incompatible with custom MuxAlarm, debug-only)

Tested on FPGA: test_firmware_update_streaming_fpga passes (534s).
Tested on emulator: test_streaming_soc_boot_successful passes (212s).

### DIFF
--- a/runtime/kernel/virtual-alarm/src/lib.rs
+++ b/runtime/kernel/virtual-alarm/src/lib.rs
@@ -1,0 +1,251 @@
+// Licensed under the Apache-2.0 license.
+//
+// Caliptra-specific virtual alarm multiplexer.
+#![no_std]
+//
+// This is a replacement for Tock's `capsules_core::virtualizers::virtual_alarm`
+// that removes the `within_range` + `next_tick_vals` optimization which fails
+// on VeeR EL2 FPGA (stale timer state after disable/re-enable cycles).
+//
+// API-compatible drop-in replacement: same types, same trait implementations.
+
+use core::cell::Cell;
+
+use kernel::collections::list::{List, ListLink, ListNode};
+use kernel::hil::time::{self, Alarm, Ticks, Time};
+use kernel::utilities::cells::OptionalCell;
+use kernel::ErrorCode;
+
+#[derive(Copy, Clone)]
+struct TickDtReference<T: Ticks> {
+    reference: T,
+    dt: T,
+    extended: bool,
+}
+
+impl<T: Ticks> TickDtReference<T> {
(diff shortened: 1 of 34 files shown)


runtime/kernel/virtual-alarm/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+        };
+        self.dt_reference.set(dt_reference);
+        let dt = dt_reference.dt;
+
+        if !self.armed.get() {
+            self.mux.enabled.set(enabled + 1);
+            self.armed.set(true);
+        }
+
+        // First alarm: always program hardware.
```
mhatrevi, line 129: Could we reprogram the hardware timer with the earliest currently armed virtual alarm instead of unconditionally using the alarm being set here?

As written, a later alarm can overwrite an earlier hardware deadline. For example, if alarm A expires in 10 ticks and alarm B is then set for 100 ticks, the physical timer is reprogrammed for B. The expired-alarm scan will not run until B fires, so A may be delivered roughly 90 ticks late.

Please preserve the earliest active deadline—such as by scanning the armed virtual alarms and programming the nearest one—while still avoiding the stale-state optimization that caused the original FPGA issue. A regression test with an earlier alarm followed by a later alarm would also help validate this behavior.

mhatrevi (changes requested):

mlvisaya:
i'm sorry, this PR is no longer valid. It has been superseded by https://github.com/chipsalliance/caliptra-mcu-sw/pull/1836. This other PR has a simpler and more correct solution. I thought I already closed this PR

PR #1735, opened 2026-07-02
build: allow overriding Caliptra ROM source ref

ArthurHeymans:
Build the Caliptra ROM through a generated helper crate so the ROM can
come from a different caliptra-sw ref than the image/update crates pinned
in this workspace.

Usage:
  CALIPTRA_ROM_REF=<commit> cargo xtask rom
  CALIPTRA_ROM_REF=<tag> cargo xtask rom

When CALIPTRA_ROM_REF is unset, the ROM still builds from the current
caliptra-sw hash resolved by Cargo.toml.

### DIFF
--- a/.github/workflows/attestation.yml
+++ b/.github/workflows/attestation.yml
@@ -8,6 +8,11 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+env:
+  # Pin the Caliptra core ROM to a release tag; image/update crates continue
+  # to use the Cargo.toml caliptra-sw rev. main-2.1 pins rom-2.1.2 instead.
+  CALIPTRA_ROM_REF: rom-2.0.3-1
+
 jobs:
   attestation_e2e:
     runs-on: ubuntu-24.04
--- a/builder/src/caliptra.rs
+++ b/builder/src/caliptra.rs
@@ -4,7 +4,7 @@
 //! the ROM, firwmare, and SoC manifest.
 
 use crate::target_dir;
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use caliptra_auth_man_gen::{
     AuthManifestGenerator, AuthManifestGeneratorConfig, AuthManifestGeneratorKeyConfig,
 };
@@ -24,7 +24,12 @@ use caliptra_image_types::{
 use caliptra_mcu_flash_image::MCU_RT_IDENTIFIER;
 use cargo_metadata::MetadataCommand;
 use hex::ToHex;
-use std::{num::ParseIntError, path::PathBuf, str::FromStr};
+use std::{
+    num::ParseIntError,
+    path::{Path, PathBuf},
+    process::Command,
+    str::FromStr,
+};
 use zerocopy::{transmute, FromBytes, IntoBytes};
 
 /// A wrapper for raw firmware bytes that implements ImageGeneratorExecutable.
@@ -417,6 +422,10 @@ impl CaliptraBuilder {
     }
 
     fn compile_caliptra_rom_cached(fpga: bool) -> Result<PathBuf> {
+        if let Ok(rom_ref) = std::env::var("CALIPTRA_ROM_REF") {
+            return Self::compile_caliptra_rom_ref_cached(&rom_ref, fpga);
+        }
+
         let platform = if fpga { "fpga" } else { "emulator" };
         if let Some(version) = Self::caliptra_version() {
             let path = target_dir().join(format!("caliptra-rom-{}-{}.bin", version, platform));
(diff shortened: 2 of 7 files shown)


.github/workflows/attestation.yml
```diff
@@ -8,6 +8,11 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+env:
+  # Keep the Caliptra core ROM on the release line for the target branch,
+  # while image/update crates continue to use the Cargo.toml caliptra-sw rev.
+  CALIPTRA_ROM_REF: ${{ (github.base_ref == 'main-2.1' || github.ref_name == 'main-2.1') && 'rom-2.1.2' || 'rom-2.0.3-1' }}
```
clundin25: nit: I think for main we can just assume a 2.0 ROM

builder/src/caliptra.rs
```diff
@@ -417,6 +422,10 @@ impl CaliptraBuilder {
     }
 
     fn compile_caliptra_rom_cached(fpga: bool) -> Result<PathBuf> {
+        if let Ok(rom_ref) = std::env::var("CALIPTRA_ROM_REF") {
+            return Self::compile_caliptra_rom_ref_cached(&rom_ref, fpga);
```
clundin25, line 426: Does it make sense to reduce `compile_caliptra_rom_cached` to call `compile_caliptra_rom_ref_cached` with the active ref to reduce duplicate code?

builder/src/caliptra.rs
```diff
@@ -428,26 +437,181 @@ impl CaliptraBuilder {
...
+                )?
+            } else {
+                caliptra_builder::rom_for_fw_integration_tests()?.to_vec()
+            };
+            std::fs::write(&path, rom_bytes)?;
+            Ok(path)
+        })
+    }
+
+    fn caliptra_rom_cc_wrapper_path() -> Result<PathBuf> {
```
clundin25: This looks a bit funky, what issue did you have to workaround?

builder/src/caliptra.rs
```diff
@@ -428,26 +437,181 @@ impl CaliptraBuilder {
...
+            .join("caliptra-rom-build")
+            .join(cache_key)
+            .join(if fpga { "fpga" } else { "emulator" });
+        let src_dir = build_dir.join("src");
+        std::fs::create_dir_all(&src_dir)?;
+        std::fs::write(
+            build_dir.join("Cargo.toml"),
+            format!(
+                r#"[package]
+name = "caliptra-rom-build"
```
clundin25, line 492: Do you think we should just check in a crate for this? Seems the xtask got a bit complex

.github/workflows/attestation.yml
ArthurHeymans: Are the main yml files not also used on main-2.1 branch?

builder/src/caliptra.rs
ArthurHeymans: Using the already resolved caliptra-builder should have shorter build times and is a bit less confused on the developer part (using just the ref from Cargo.toml and not the actual builder code could be confusing?).

builder/src/caliptra.rs
ArthurHeymans: so caliptra-sw for main-2.0 uses that patched openssl. I added a comment, but maybe we can bump openssl there (assuming the mldsa code is now upstream)? It should not change the ROM hopefully.

builder/src/caliptra.rs
ArthurHeymans: So that would be a crate outside the workspace. I'll try to do this.

.github/workflows/attestation.yml
clundin25: They are but this means each branch has dead code

builder/src/caliptra.rs
clundin25: IMO Let's just bump the openssl version if means we can skip all this

builder/src/caliptra.rs
clundin25: My main concern is that the build code branches quite a bit. Probably outside of the scope of this PR but it would be nice to clean it up

builder/src/caliptra.rs
ArthurHeymans: Sorry this was to fix my local nixos mess. Removed.

builder/src/caliptra.rs
ArthurHeymans: So we need to dynamically generate it anyway for the ref.

clundin25 (approved):

builder/src/caliptra.rs
clundin25: Do you have a flake with a working devshell? It would be pretty cool to check in :)

PR #1739, opened 2026-07-06
[doc] Update integrator guide

### LINKED ISSUE #1687: Integration guidance documentation
To help integrators know where to start and what pieces need to be added for their specific integration, it would be good to have an integration guide (this might already exist, but I haven't seen one).

This could follow a similar pattern to the [hardware integration guide](https://github.com/chipsalliance/caliptra-rtl/blob/main/docs/CaliptraIntegrationSpecification.md).

An example of something that should be configured is the `MockCommandAuthorizer` in https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/platforms/emulator/runtime/userspace/apps/user/src/mcu_mbox/cmd_auth_mock.rs which needs to be overridden by each integrator.

mhatrevi:


docs/src/integrator-guide.md
```diff
@@ -65,6 +138,169 @@ the integration-specific fuse map.
...
+SoC manifest digest into PCR31. If an integrator needs that specific digest in
+PCR31, add an explicit Caliptra RT-side stash after manifest verification rather
+than asking MCU ROM to stash unverified manifest bytes.
+
+MCU ROM does stash ROM-owned measurements that only it can observe. When
+`RomParameters::stash_rom_digest` is set, ROM stashes its own ROM digest. ROM
+also hashes and stashes `FIELD_ENTROPY_STATE` after Caliptra RT is ready for
+runtime mailbox commands.
+
+## 2.1 Security-State Integration
```
parvathib: Would it make sense to refer to this as "MCU ROM 2.1" to make it clearer?

docs/src/integrator-guide.md
```diff
@@ -65,6 +138,169 @@ the integration-specific fuse map.
...
+
+The 2.1 ROM has several security-state hooks that require integrator policy:
+
+| Feature | Integrator responsibility | ROM behavior |
```
parvathib, line 317: are any of these applicable to 2.0 as well?

docs/src/integrator-guide.md
```diff
@@ -65,6 +138,169 @@ the integration-specific fuse map.
...
+formats supported MCU mailbox commands through a platform-provided
+`MailboxDriver`; it does not give an external BMC native access to the MCI
+mailbox, and not every runtime MCI command is wrapped by the current host
+utility dispatch tables. If an OOB BMC must initiate MCI-only operations such
+as PK provisioning, PK revocation, or direct Caliptra SVN increment, the
+platform must provide a trusted SoC-side service or bridge that owns the MCI
+access, exposes the intended command wrappers, and enforces the deployment
+policy.
+
+## Measurement And PCR31 Responsibilities
```
parvathib, line 290: isn't this applicable to 2.0 as well?

parvathib (commented):
The content for 2.1 looks very similar to 2.0, so the comments from #1740 largely apply here as well.
I have a couple of additional comments specific to the differences in 2.1.

docs/src/integrator-guide.md
mhatrevi: Updated.

docs/src/integrator-guide.md
mhatrevi: Some of them are; they will be documented accordingly in the main branch.

docs/src/integrator-guide.md
mhatrevi: Makes sense; also removing the 2.1 wording as this integration guide is for 2.1 ROM only.

parvathib (approved):

PR #1744, opened 2026-07-06
Switch reference command authorizer to use asymmetric keys

zhalvorsen:
The old reference used HMAC (symmetric) keys which are difficult to keep secure because they need to exist in both an HSM or similar infrastructure and be referenced by the firmware. Instead of giving an example of provisioning this key into fuses and worrying about key hygiene we have switched to using asymmetric keys so a public keys can be stored directly in firmware.

This example uses both ECDSA and ML-DSA to verify the authorization to maintain PQC safety.


common/command-auth-challenge-signer/src/lib.rs
```diff
@@ -1,54 +1,85 @@
...
+        let mut padded_mldsa_sig = mldsa_sig.to_vec();
+        padded_mldsa_sig.resize(MLDSA87_SIGNATURE_BYTE_SIZE, 0u8); // Pad to 4628 bytes
+
+        Ok(HybridSignature {
+            ecc_sig_r: ecc_sig_bytes[..48]
+                .try_into()
+                .map_err(|_| anyhow::anyhow!("Failed to convert ECC signature r"))?,
+            ecc_sig_s: ecc_sig_bytes[48..96]
+                .try_into()
+                .map_err(|_| anyhow::anyhow!("Failed to convert ECC signature s"))?,
```
clundin25, line 79: nit: Does it make sense to use a single buffer for the EC signature? A different direction to the existing Caliptra APIs but makes EC less special-case

clundin25 (approved):

RaunakGu:
hi @zhalvorsen and @clundin25 ,

Since this functionality is now being covered by the new RFC (https://github.com/chipsalliance/caliptra-sw/issues/3928) and its associated implementation PR, perhaps we should consider closing this PR and consolidating our efforts around the RFC-based approach to keep the implementation aligned and avoid code conflict, below is the PRs.

**Here is the PRs:**
https://github.com/chipsalliance/caliptra-mcu-sw/pull/1811
https://github.com/chipsalliance/caliptra-sw/pull/4003

zhalvorsen:
> hi @zhalvorsen and @clundin25 ,
> 
> Since this functionality is now being covered by the new RFC ([chipsalliance/caliptra-sw#3928](https://github.com/chipsalliance/caliptra-sw/issues/3928)) and its associated implementation PR, perhaps we should consider closing this PR and consolidating our efforts around the RFC-based approach to keep the implementation aligned and avoid code conflict, below is the PRs.
> 
> **Here is the PRs:** #1811 [chipsalliance/caliptra-sw#4003](https://github.com/chipsalliance/caliptra-sw/pull/4003)

I apologize, I didn't look closely enough at your RFC initially to realize the extent of the changes required for `caliptra-sw`.

We are currently having internal discussions about how we want to proceed with this, and we will pull you into the conversation as soon as we have a clearer picture of our requirements. 

In the meantime, we want to avoid using symmetric secrets for authorization. We plan to merge this PR while we continue those discussions.

I am sorry you already started on the implementation. For context, Caliptra RFCs generally require review and approval from each of the Caliptra TAC member companies.

RaunakGu:
@zhalvorsen, 

Sure, no problem. Please go ahead with this MCU-side changes; we can work on consolidating things later.

In the meantime, how should we follow up with the other TAC members to review and approve the RFC?
I'm targeting to land the implementation in the Caliptra FW 2.1.2 release and the MCU RT 2.1 release.

Issue #1557, opened 2026-06-03
Undefined Behavior in pointer-to-reference conversion


DavidSkrundz-Solidigm:
The following function triggers Undefined Behavior in the pointer-to-reference conversion because `&[u8; _]` has an alignment of 1 but `&[u32; _]` has an alignment of 4.  Rust requires all references to be [properly aligned](https://doc.rust-lang.org/std/ptr/index.html#pointer-to-reference-conversion).

https://github.com/chipsalliance/caliptra-mcu-sw/blob/62c3bade80aa55c2961b65e722846533479abc45/rom/src/mailbox.rs#L76-L81

A possible fix is to wrap the u8 array in an aligned newtype:

```rust
#[repr(align(4))]
struct Aligned4([u8; CMB_SHA_CONTEXT_SIZE]);

/// Interpret an aligned `[u8; CMB_SHA_CONTEXT_SIZE]` as `&[u32; CTX_DWORDS]`. 
fn ctx_as_u32(ctx: &Aligned4([u8; CMB_SHA_CONTEXT_SIZE])) -> &[u32; CTX_DWORDS] { 
    // Safety: CMB_SHA_CONTEXT_SIZE is a multiple of 4 and the array has alignment 4
    unsafe { &*(ctx.0.as_ptr() as *const [u32; CTX_DWORDS]) } 
} 
```

Enabling the `clippy::cast-ptr-alignment` lint catches this issue.

PR #1749, opened 2026-07-07
Updating the ctx_as_u32 function to remove undefined behavior.

v-focutter:
This change removes the undefined behavior from the ctx_as_u32 function when the passed in u8 array is not dword aligned. Where previously the code returned a u32 pointer to the u8 array by doing an unsafe cast, the updated code returns a new u32 array. 

Task [#1557](https://github.com/chipsalliance/caliptra-mcu-sw/issues/1557)

### DIFF
--- a/rom/src/mailbox.rs
+++ b/rom/src/mailbox.rs
@@ -77,8 +77,11 @@ fn u32_byte_sum(data: &[u32]) -> u32 {
 fn ctx_as_u32(ctx: &[u8; CMB_SHA_CONTEXT_SIZE]) -> [u32; CTX_DWORDS] {
     let mut ctx_u32 = [0u32; CTX_DWORDS];
 
-    for (dst, chunk) in ctx_u32.iter_mut().zip(ctx.chunks_exact(4)) {
-        *dst = u32::from_le_bytes([chunk[0], chunk[1], chunk[2], chunk[3]]);
+    // Read each 4-byte lane directly to avoid bounds-checked indexing paths.
+    for (i, dst) in ctx_u32.iter_mut().enumerate() {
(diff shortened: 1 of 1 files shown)


rom/src/mailbox.rs
```diff
@@ -73,11 +73,18 @@ fn u32_byte_sum(data: &[u32]) -> u32 {
         .fold(0u32, |acc, &w| acc.wrapping_add(word_byte_sum(w)))
 }
 
-/// Interpret a `[u8; CMB_SHA_CONTEXT_SIZE]` as `&[u32; CTX_DWORDS]`.
-fn ctx_as_u32(ctx: &[u8; CMB_SHA_CONTEXT_SIZE]) -> &[u32; CTX_DWORDS] {
-    // Safety: CMB_SHA_CONTEXT_SIZE is a multiple of 4 and the array is
-    // stack-allocated with natural alignment.
-    unsafe { &*(ctx.as_ptr() as *const [u32; CTX_DWORDS]) }
+/// Convert a `[u8; CMB_SHA_CONTEXT_SIZE]` to `[u32; CTX_DWORDS]`.
```
mhatrevi, line 76: Could this use ctx.chunks_exact(4) and u32::from_le_bytes([bytes[0], bytes[1], bytes[2], bytes[3]])? That preserves the little-endian conversion while eliminating the remaining raw-pointer arithmetic and unsafe blocks.

mhatrevi (changes requested):

rom/src/mailbox.rs
v-focutter: That was one of the first things I tried, but it still generated the 'panic_is_possible' failure.

PR #1756, opened 2026-07-10
Fixing the current failures in the size-history workflow

v-focutter:
The Enable all features (#1691) PR added a new option in CaliptraBuildArgs, no_default_features. This flag was added to most of the uses of this structure for executing build commands, but the code for size-history was not updated. Without this flag the resulting roms are too big to be properly linked.


mhatrevi (approved):

mhatrevi (changes requested):
Size seems incorrect; please investigate.

Issue #963, opened 2026-02-25
[OCP-LOCK]: Implement “ENDORSE_HPKE_PUBKEY”


clundin25:
MCU needs to endorse Caliptra’s HPKE pub keys. We need to support ML-DSA & ECDSA certs, for all three HPKE algorithms.

MCU should use its DPE context for the endorsement.

You would need to:

- Retrieve the HPKE public key from Caliptra
- Generate an x509 certificate based on the one from Caliptra.
   - Do we need to support HPKE Identifiers in caliptra-dpe?
- Endorse it with the MCU Firmware DPE context
- Embed Caliptra’s HPKE pub key into the certificate as its subject key.
- Return it over the MCU Firmware mailbox

daeyoungkim-87:
Hi, @clundin25 
I have a question about this change.

I noticed that the DPE leaf certificate has BasicConstraints: CA=FALSE, but it is also used to sign another X.509 certificate. Was this intentional?

Should I understand this as an OCP LOCK–specific attestation PKI design (rather than a standard RFC 5280 CA hierarchy), where the verifier does not rely on the usual CA semantics?

Thanks!

clundin25:
@daeyoungkim-87,

Thanks for your question. That's a miss, and I believe you are correct. We should most likely instead use the export-cdi feature to create a cert + key that can be used as a CA.

I opened https://github.com/chipsalliance/caliptra-mcu-sw/issues/1765 to track this

clundin25:
Initial implementation is in.

PR #1757, opened 2026-07-10
[OCP-LOCK]: HPKE Endorsement Certificates

clundin25:
* Added Mailbox commands for
    * Enumerating HPKE keys
    * Endorsing an HPKE key
* Added tests to validate the HPKE certificates.
* Migrated OCP LOCK integration tests to a self-standing folder.
* Split OCP LOCK Runtime config trait into a kernel and userspace.

This resolves: https://github.com/chipsalliance/caliptra-mcu-sw/issues/963


platforms/emulator/runtime/userspace/apps/example/src/main.rs
```diff
@@ -439,3 +439,16 @@ mod test {
         drop(kernel);
     }
 }
+
+#[cfg(feature = "test-ocp-lock")]
+pub mod ocp_lock_config {
```
zhalvorsen, line 444: Should we add info to the integration guide for this config?

zhalvorsen (approved):

platforms/emulator/runtime/userspace/apps/example/src/main.rs
clundin25: Ah yea, let me do that in a follow up.

Tracking: https://github.com/chipsalliance/caliptra-mcu-sw/issues/1816

PR #1758, opened 2026-07-13
Use measurement API for DPE handles

parvathib:
Route attestation cert, kid, and signing operations through the library-owned Measurement API instance. Measurement API now owns the read-handle, issue-command, write-rotated-handle sequence for DPE operations that can rotate handles.

Keep caliptra-api-lite as the low-level DPE mailbox layer, while SPDM cert store and OCP EAT use the Measurement API surface instead of calling DPE helpers directly.


runtime/userspace/api/caliptra-api-lite/src/slice.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#[inline]
+pub(crate) fn checked_slice_mut(src: &mut [u8], offset: usize, len: usize) -> McuResult<&mut [u8]> {
+    let end = offset.checked_add(len).ok_or(INVARIANT)?;
+    src.get_mut(offset..end).ok_or(INVARIANT)
+}
+
+#[inline]
+pub(crate) fn copy_bytes(dst: &mut [u8], src: &[u8]) {
+    for (dst_byte, src_byte) in dst.iter_mut().zip(src.iter()) {
```
mlvisaya, line 29: nit: might be good to assert or return an error if src.len != dst.len

runtime/userspace/api/caliptra-api-lite/src/sha.rs
```diff
@@ -65,18 +66,23 @@ impl<B: Deref<Target = [u8]> + DerefMut> HashState<B> {
 
     /// Deep-copy this running hash state into a new buffer.
     pub fn clone_into<B2: Deref<Target = [u8]> + DerefMut>(&self, mut buf: B2) -> HashState<B2> {
```
mlvisaya: not: Can this return Result<HashState<B2>>? If for some reason, buf,get_mut or self.inner.get below returns None, the HashState returned is uninitialized. We can probably return an error if one of it returns None.

mlvisaya (approved):
some nits, otherwise, I think it's good.

Issue #1572, opened 2026-06-04
MCU ROM Version Reporting


RaunakGu:
https://github.com/chipsalliance/caliptra-ss/blob/a5a2445ef8b65685f7eec0a9008172742dd6640d/src/mci/rtl/mci_reg.rdl#L109

Can we have the MCU ROM version maintained and reporting implemented?

cc @zhalvorsen

RaunakGu:
@mhatrevi, if MCU requires to be in FIPS boundary(which mostly will be required), Then MCU ROM needs to have very similar Versioning command as Caliptra?

RaunakGu:
# Proposal: MCU ROM Firmware Version Reporting

**Status:** Proposed for adoption &nbsp;·&nbsp; **Target:** Caliptra SS MCU ROM

## Motivation

The existing firmware-version mechanisms — MCI-mailbox `MFWV`, MCTP-VDM `FirmwareVersion`, and PLDM — all live in the Tock **runtime** and cannot answer during the **ROM boot window**. This proposal adds a standardized way for a SoC/host agent to read the MCU **ROM** version *before the runtime is up*, for:

- Boot triage and field diagnostics
- Provenance / trust binding
- Recovery and streaming-boot handshakes

> [!NOTE]
> This is build/diagnostic provenance only. The MCU is **outside** the FIPS 140-3 cryptographic boundary (Caliptra Core is the sole FIPS module).

## Design

### 1. Version encoding

`Major.Minor.Patch` compile-time constants packed 5/5/6 into a `u16`:

```rust
version = (Major & 0x1F) << 11 | (Minor & 0x1F) << 6 | (Patch & 0x3F)
```

| Field | Bits      | Width | Max |
| ----- | --------- | ----- | --- |
| Major | `[15:11]` | 5     | 31  |
| Minor | `[10:6]`  | 5     | 31  |
| Patch | `[5:0]`   | 6     | 63  |

Current value `2.1.0` = `0x1040`. This mirrors the encoding of the read-only companion `HW_REV_ID` register.

### 2. Register publication

The packed version is written (as `u32`) into MCI **`FW_REV_ID[0]`** — the RTL-defined *"MCU FW RevID"* register (`mci_reg_fw_rev_id`, offset `0x10`, `sw=rw`, HW never overwrites) — early in ROM start. It is then SoC-readable over AXI with no mailbox handshake.

### 3. Mailbox command `MCUV` (`0x4D435556`)

[...]

PR #1760, opened 2026-07-13
MCU ROM Version Reporting

v-focutter:
Summary
* Extending the existing firmware version request function to return the MCU Rom version.
* Update the tests to also test the MCU Rom version 
* Documentation update


RaunakGu:
@v-focutter 

Original Idea was to use mci reg for this purpose
https://github.com/chipsalliance/caliptra-ss/blob/a5a2445ef8b65685f7eec0a9008172742dd6640d/src/mci/rtl/mci_reg.rdl#L109

Can you please check the proposal in the below filed issue. https://github.com/chipsalliance/caliptra-mcu-sw/issues/1572

cc @mhatrevi

PR #1768, opened 2026-07-15
Add erase_size() to FlashStorage trait

### LINKED ISSUE #1511: runtime: FlashStorageToPages erase operates at page granularity (256B) instead of sector granularity (4KB+)
## Problem

`FlashStorageToPages::erase()` in `runtime/kernel/drivers/flash/src/flash_storage_to_pages.rs` splits erase operations into 256-byte page-sized chunks via repeated calls to `hil::flash::Flash::erase_page()`.

https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/runtime/kernel/drivers/flash/src/flash_storage_to_pages.rs#L167-L202

However, real SPI flash devices have a minimum erase granularity of **4KB (sector)**, with larger options of 32KB/64KB (block). The 256-byte "page" is only meaningful for read/write (program) operations.

This means:
1. **Redundant erase operations**: Erasing 4KB calls `erase_page()` 16 times (4096/256), but on real hardware all 16 calls would erase the same 4KB sector.
2. **Increased flash wear**: Repeated erases on the same sector accelerate wear-out.
3. **Incorrect partial-erase handling**: The current code does read-modify-write for "partial page" erases, but the actual atomic erase unit on real flash is a sector, not a page.

## Possible Solutions

1. **Introduce erase granularity to the Flash trait**: Add an `erase_size()` or similar method to `hil::flash::Flash` that returns the minimum erase unit (e.g., 4096), separate from the page size used for read/write.

2. **Add an erase_sector/erase_block API**: Extend the flash trait with sector/block-level erase methods that map directly to SPI flash commands.

mlvisaya:
Addresses issue #1511 where FlashStorageToPages erases at page granularity (256B) instead of sector granularity (4KB+). On real SPI flash hardware, the minimum erase unit is a sector (typically 4KB), not a page.

Changes:
- Add erase_size() method to FlashStorage<'a> trait returning the minimum erase granularity in bytes
- Update FlashStorageToPages to accept erase_size parameter and use it for erase operations: one erase_page() call per sector instead of per page
- Require erase address alignment to erase_size; non-aligned lengths are rounded up to the next sector boundary (matching real flash behavior)
- Update emulator flash HW model (DummyFlashCtrl) to erase a full 4KB sector per erase_page() call
- Add ERASE_SIZE = 4096 constants to emulator and FPGA flash controllers
- Expose erase_size to userspace via new GET_ERASE_SIZE syscall command (6) in FlashPartition capsule
- Add get_erase_size() to userspace SpiFlash API and FakeFlashDriver
- Update FlashPartitionComponent and instantiate_flash_partitions! macro to propagate erase_size
- Update tests to use erase-aligned operations

Fixes: #1511

### DIFF
--- a/emulator/periph/src/flash_ctrl.rs
+++ b/emulator/periph/src/flash_ctrl.rs
@@ -104,7 +104,7 @@ impl DummyFlashCtrl {
     pub const PAGE_SIZE: usize = 256;
 
     /// Erase sector size. The minimum erase unit on real flash hardware.
-    pub const ERASE_SIZE: usize = 4096;
+    pub const ERASE_SECTOR_SIZE: usize = 4096;
 
     /// Maximum number of pages in the flash storage connected to the controller.
     /// This is a dummy value, the actual value should be set based on the flash storage size.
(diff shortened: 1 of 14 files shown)


emulator/periph/src/flash_ctrl.rs
```diff
@@ -103,6 +103,9 @@ impl DummyFlashCtrl {
     /// Page size for the flash storage connected to the controller.
     pub const PAGE_SIZE: usize = 256;
 
+    /// Erase sector size. The minimum erase unit on real flash hardware.
+    pub const ERASE_SIZE: usize = 4096;
```
helloxiling: nit: we may name it as `ERASE_SECTOR_SIZE` for clarity.

helloxiling (approved):
Looks good to me.

emulator/periph/src/flash_ctrl.rs
mlvisaya: thanks. i'll rename it to ERASE_SECTOR_SIZE

PR #1771, opened 2026-07-16
Adding nightly workflows to main-2.1

v-focutter:
Adding the nightly workflow changes to the main-2.1 branch. 

This will not trigger the workflow, as scheduled workflows will only execute on the default branch.

### DIFF
--- a/.github/workflows/attestation.yml
+++ b/.github/workflows/attestation.yml
@@ -3,10 +3,9 @@
 name: End-to-End Attestation
 
 on:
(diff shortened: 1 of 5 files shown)


mlvisaya (commented):
I think the cherry-pick is good. We could probably include in this PR the way for the workflow to run it on main-2.1 even though the workflow is targeted for main (by default).

mlvisaya (approved):

PR #1773, opened 2026-07-16
Add on_fw_update_request hook to prevent flash wear-out attacks

### LINKED ISSUE #1528: Requirement: firmware update rate limiting to prevent flash wear-out attacks
## Background

The MCU firmware update path currently has no protection against repeated update attempts that could exhaust flash P/E cycle endurance. An attacker or misbehaving Update Agent could trigger continuous update cycles, wearing out flash partitions.

## Request

Provide a **callback/notification hook** in the FW update path (e.g., at the PLDM FD layer) that is invoked when a firmware update request is received, **before any flash I/O occurs**.

The callback should allow SoC-specific logic to:
- Accept or reject the update request
- Return an error code or delay hint to the requester

### Proposed Interface (conceptual)

```rust
/// Called before processing a FW update request.
/// Return Ok(()) to proceed, or Err to reject.
fn on_fw_update_request(&mut self, component_id: u32) -> Result<(), FwUpdateRejectReason>;

mlvisaya:
Add a new FirmwareUpdateHooks::on_fw_update_request() method that is invoked from handle_component when the PLDM UA sends an UpdateComponent request. This gives platforms the ability to accept or reject firmware updates before any flash I/O occurs, protecting against flash wear-out from repeated update cycles.

The hook receives the FirmwareComponent being requested so platforms can make decisions based on component classification, identifier, or size. Returning Err rejects the update with CompPrerequisitesNotMet.

Changes:
- Add on_fw_update_request(&self, &FirmwareComponent) to FirmwareUpdateHooks trait with a default no-op implementation
- Store hooks in a FW_UPDATE_HOOKS static so pldm_fdops can access them
- Call the hook from UpdateFdOps::handle_component on UpdateComponent
- Add ExitOnRejectActions UA state machine actions for testing
- Add test-firmware-update-reject feature and integration test

Fixes: #1528


helloxiling (approved):
Looks good to me.

mlvisaya:
thanks!

PR #1776, opened 2026-07-17
builder: Use image_id from ImageCfg for SOC image flash TOC identifiers

### LINKED ISSUE #1580: FirmwareUpdater::verify() fails when image_id differs from SOC_IMAGES_BASE_IDENTIFIER
### Problem

When building a flash image with `flash_image_create()`, SOC image identifiers in the flash TOC are always assigned as `SOC_IMAGES_BASE_IDENTIFIER` (0x1000) + index, regardless of the `image_id` configured for each SOC image.

However, the auth manifest generated by `CaliptraBuilder::get_soc_manifest_metadata()` uses `image_id` from the config as `fw_id` in `AuthManifestImageMetadata`.

At runtime, `FirmwareUpdater::verify()` iterates flash TOC entries and passes `image_header.identifier` to `get_image_metadata()`, which searches the auth manifest by `fw_id`. When `image_id` in the config is not equal to `SOC_IMAGES_BASE_IDENTIFIER + index`, the lookup fails because the flash TOC identifier and the manifest `fw_id` are different values.

### Example

Given two SOC images built with `ImageCfg` where `image_id` does not match `SOC_IMAGES_BASE_IDENTIFIER + index`:

ImageCfg { image_id: 12, component_id: 4096, path: "soc_image_0.bin", ... } ImageCfg { image_id: 6, component_id: 4097, path: "soc_image_1.bin", ... }

Build result:
- `flash_image_create()` assigns flash TOC identifiers as 4096, 4097 (hardcoded from `SOC_IMAGES_BASE_IDENTIFIER`)
- `CaliptraBuilder::get_soc_manifest_metadata()` sets auth manifest `fw_id` as 12, 6 (from `image_id`)

Runtime:
- `FirmwareUpdater::verify()` → `get_image_metadata(4096)` → searches `fw_id == 4096` → not found → verify fails

mlvisaya:
flash_image_create() was always assigning SOC_IMAGES_BASE_IDENTIFIER + index as the flash TOC identifier for SOC images. However, the auth manifest generated by get_soc_manifest_metadata() uses image_id from ImageCfg as fw_id. This mismatch caused FirmwareUpdater::verify() to fail when image_id differs from SOC_IMAGES_BASE_IDENTIFIER + index, because the flash TOC identifier and manifest fw_id wouldn't match.

Use image_id from args.soc_images when available. Fall back to SOC_IMAGES_BASE_IDENTIFIER + index when only soc_image_paths is provided (no ImageCfg) for backward compatibility.

Fixes: #1580

### DIFF
--- a/builder/src/flash_image.rs
+++ b/builder/src/flash_image.rs
@@ -234,10 +234,15 @@ pub fn flash_image_create(args: &CaliptraBuildArgs) -> Result<()> {
     }
 
     // Generate FirmwareImage from soc image buffer
-    let mut soc_image_identifer = SOC_IMAGES_BASE_IDENTIFIER;
-    for soc_img in soc_img_buffers.iter() {
-        images.push(FirmwareImage::new(soc_image_identifer, soc_img)?);
-        soc_image_identifer += 1;
+    // Use image_id from soc_images config when available; otherwise fall back to
+    // SOC_IMAGES_BASE_IDENTIFIER + index for backward compatibility.
+    for (i, soc_img) in soc_img_buffers.iter().enumerate() {
+        let identifier = if let Some(ref soc_images) = args.soc_images {
+            soc_images[i].image_id
+        } else {
+            SOC_IMAGES_BASE_IDENTIFIER + i as u32
+        };
+        images.push(FirmwareImage::new(identifier, soc_img)?);
     }
 
     let image_info = generate_image_info(images.clone());
@@ -379,7 +384,7 @@ pub fn write_partition_table(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::PROJECT_ROOT;
+    use crate::{ImageCfg, PROJECT_ROOT};
     use std::fs::{self, File};
     use std::io::Write;
     use tempfile::NamedTempFile;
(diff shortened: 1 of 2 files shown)


builder/src/flash_image.rs
```diff
@@ -234,10 +234,15 @@ pub fn flash_image_create(args: &CaliptraBuildArgs) -> Result<()> {
     }
 
     // Generate FirmwareImage from soc image buffer
-    let mut soc_image_identifer = SOC_IMAGES_BASE_IDENTIFIER;
-    for soc_img in soc_img_buffers.iter() {
-        images.push(FirmwareImage::new(soc_image_identifer, soc_img)?);
-        soc_image_identifer += 1;
+    // Use image_id from soc_images config when available; otherwise fall back to
+    // SOC_IMAGES_BASE_IDENTIFIER + index for backward compatibility.
+    for (i, soc_img) in soc_img_buffers.iter().enumerate() {
```
mhatrevi, line 239: This branch does not appear to be reached by the main builder flow. `builder/src/all.rs::create_flash_image()` passes only `soc_image_paths`, even though the auth manifest is generated from `feature_soc_images`. As a result, this code still takes the fallback path and produces `SOC_IMAGES_BASE_IDENTIFIER + index`, so non-default `image_id` values remain mismatched with the manifest `fw_id`.

Could we propagate the corresponding `Vec<ImageCfg>` through `create_flash_image()` and add a regression test using nonsequential IDs such as `12` and `6`?

mhatrevi (changes requested):

mhatrevi (approved):

PR #1777, opened 2026-07-17
fix(pldm): propagate image authorization failure to PLDM Update Agent

### LINKED ISSUE #1636: Streaming boot: image authorization failure is not reported to the PLDM Update Agent (UA always sees VerifySuccess)
hi @mlvisaya, Can you take a look at this issue, thanks

## Summary

In the streaming-boot image-loading path, when `AUTHORIZE_AND_STASH` rejects an
image, the failure is **not** propagated to the PLDM Update Agent (UA). The
Firmware Device (FD) still reports `VerifyResult::VerifySuccess` in its
`VerifyComplete` (0x17) request, so from the UA's perspective a failed update
looks successful — or surfaces only as a dropped/timed-out session. A UA/BMC
therefore cannot distinguish a genuinely verified image from one that failed
local authorization.

**Affected component:** `runtime/userspace/api/caliptra-api/src/image_loading`
**Observed on:** `main-2.1` @ `87fa7d82a`

### 1. `authorize_image()` correctly turns a bad verdict into an error
`runtime/userspace/api/caliptra-api/src/image_loading/mod.rs:293`

```rust
if resp.auth_req_result != IMAGE_AUTHORIZED {   // != 0xDEADC0DE
    return Err(ErrorCode::Fail);
}
```

### 2. `PldmImageLoader::load_and_authorize()` finalizes with success on the error path
`runtime/userspace/api/caliptra-api/src/image_loading/mod.rs:199`

```rust
if result.is_err() {
    self.finalize()?;            // finalize() passes VerifyResult::VerifySuccess
    return Err(ErrorCode::Fail);
}
```

### 3. `finalize()` is hardcoded to `VerifySuccess`
`runtime/userspace/api/caliptra-api/src/image_loading/mod.rs:167`

```rust
pub fn finalize(&self) -> Result<(), ErrorCode> {
    pldm_client::finalize(VerifyResult::VerifySuccess)   // always Success
}
```

`pldm_client::finalize()` then stores it (`pldm_client.rs:202`):

```rust
pub fn finalize(verify_result: VerifyResult) -> Result<(), ErrorCode> {
    // ...
    ctx.download_complete = true;
    ctx.verify_result = verify_result;   // = VerifySuccess
```

### 4. The FD `verify()` callback reports that stored result verbatim
`runtime/userspace/api/caliptra-api/src/image_loading/pldm_fdops.rs`

```rust
let verify_result = DOWNLOAD_CTX.lock(|ctx| ctx.borrow().verify_result);
Ok(verify_result)   // VerifyComplete carries VerifySuccess even on failure
```

## Behavior / impact

- The actual image verification (Caliptra `AUTHORIZE_AND_STASH`, SHA vs. SoC-manifest
  digest) happens during the **Download** phase, inside the FD. The PLDM
  `VerifyComplete` is purely a *report* of that result.
- On the failure path the report is hardcoded `VerifySuccess`, so:
  - the UA receives `VerifyComplete = VerifySuccess` even though the image was
    rejected, and
  - the only signal of failure is the FD returning `Err` upward (the app then
    halts, e.g. `System::exit(1)`), which the UA observes as a **dropped/
    timed-out session**, not a PLDM-level failure code.
- **Net effect:** a UA cannot tell, over PLDM, that authorization failed.

## Note: failure result codes already exist but are unused on this path

`common/pldm/src/message/firmware_update/verify_complete.rs:12`

```rust
pub enum VerifyResult {
    VerifySuccess = 0x00,
    VerifyErrorVerificationFailure = 0x01,
    VerifyErrorVersionMismatch = 0x02,
    VerifyFailedFdSecurityChecks = 0x03,
    VerifyErrorImageIncomplete = 0x04,
    VerifyTimeOut = 0x09,
    VerifyGenericError = 0x0a,
}
```

mlvisaya:
When AUTHORIZE_AND_STASH rejects an image during streaming boot, the VerifyResult reported to the PLDM Update Agent (UA) was always VerifySuccess, because finalize() was hardcoded to that value.

This change:
- Makes PldmImageLoader::finalize() accept a VerifyResult parameter instead of hardcoding VerifySuccess.
- On the authorization failure path in load_and_authorize(), passes VerifyResult::VerifyFailedFdSecurityChecks so the UA receives the correct failure code in the VerifyComplete message.
- Restructures the user-app streaming boot flow so that wait_for_service_stopped() is always called (via async block), ensuring the PLDM service has time to send VerifyComplete before the process exits.
- Changes CancelUpdateComponent responder action to Complete so the PLDM service properly shuts down after the UA cancels a failed component (prevents wait_for_service_stopped() from hanging).

Fixes #1636

### DIFF
--- a/platforms/emulator/runtime/userspace/apps/user/src/image_loader/mod.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/image_loader/mod.rs
@@ -48,6 +48,15 @@ use caliptra_mcu_measurement_api::{ImageHashSource, ImageMetadata};
     feature = "test-pldm-fw-update-e2e",
     feature = "test-pldm-streaming-boot"
 ))]
+use caliptra_mcu_pldm_common::message::firmware_update::verify_complete::VerifyResult;
+#[allow(unused)]
+#[cfg(any(
+    feature = "streaming-boot",
+    feature = "test-pldm-discovery",
+    feature = "test-pldm-fw-update",
+    feature = "test-pldm-fw-update-e2e",
+    feature = "test-pldm-streaming-boot"
+))]
 use caliptra_mcu_pldm_lib::daemon::PldmService;
 #[cfg(any(feature = "streaming-boot", feature = "flash-boot"))]
 use caliptra_mcu_spdm_pal::{BitmapAllocator, BITMAP_SLOT_SIZE};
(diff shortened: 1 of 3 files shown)


mhatrevi (approved):

helloxiling (approved):
Looks good. 
Thanks for fixing it.

PR #1778, opened 2026-07-18
fix(pldm): resolve race condition between responder and initiator tas…

### LINKED ISSUE #1764: FW Update: FD responder may process ActivateFirmwareRequest before initiator processes ApplyComplete response handling (race condition)
## Summary
pldm_responder_task and pldm_initiator_task in daemon.rs run as two concurrent Embassy tasks sharing the same CmdInterface (and thus fd_ctx state). After the FD sends ApplyComplete(Success) as an initiator request and the UA responds, the UA immediately sends ActivateFirmwareRequest. If the executor polls the responder task before the initiator task has processed the ApplyComplete response (which transitions FD state from Apply to ReadyXfer), the responder handles ActivateFirmwareRequest while FD state is still Apply, returning InvalidStateForCommand (0x84).

The UA cannot parse this short error response (4 bytes instead of expected 6-byte ActivateFirmwareResponse), logs it as "Unhandled packet", and eventually times out.

## Log
2026-07-15T10:35:35.962Z INFO  [caliptra_mcu_pldm_ua::update_sm] Apply success
2026-07-15T10:35:35.962Z INFO  [caliptra_mcu_pldm_ua::update_sm] Num components downloaded: 1
**2026-07-15T10:35:36.061Z ERROR [caliptra_mcu_pldm_ua::daemon] Unhandled packet: RxPacket { src: EndpointId(0), payload: Payload { data: [4, 5, 26, 132], len: 4 } }**
2026-07-15T10:35:36.061Z ERROR [caliptra_mcu_pldm_ua::daemon] Error handling packet: Err(())
2026-07-15T10:35:38.962Z INFO  [caliptra_mcu_pldm_ua::update_sm] Retrying request, attempt: 1
2026-07-15T10:35:41.962Z INFO  [caliptra_mcu_pldm_ua::update_sm] Retrying request, attempt: 2
2026-07-15T10:35:44.962Z INFO  [caliptra_mcu_pldm_ua::update_sm] Retrying request, attempt: 3
2026-07-15T10:35:47.962Z ERROR [caliptra_mcu_pldm_ua::update_sm] Max retry count reached, giving up on request
Packet Decode
[4, 5, 26, 132] = [0x04, 0x05, 0x1A, 0x84]:

## Packet Decode
ERROR [caliptra_mcu_pldm_ua::daemon] Unhandled packet: RxPacket { src: EndpointId(0), payload: Payload { **data: [4, 5, 26, 132],** len: 4 } } 

Byte 0 (0x04): instance_id=4, rq=0 (Response)
Byte 1 (0x05): PLDM Type 5 (Firmware Update)
Byte 2 (0x1A): Command 26 = ActivateFirmware
Byte 3 (0x84): Completion code = InvalidStateForCommand

## Root Cause
In daemon.rs, PldmService::start() spawns two concurrent tasks:

```
self.spawner.spawn(pldm_responder_task(cmd_interface, ...)).unwrap();
self.spawner.spawn(pldm_initiator_task(cmd_interface, ...)).unwrap();
```
During the Apply→Activate transition:

1. Initiator task sends ApplyComplete(Success) to UA, then awaits transport.receive_response()
2. UA responds to ApplyComplete and immediately sends ActivateFirmwareRequest (back-to-back, per PLDM spec this is valid)
3. MCTP layer delivers the response to initiator's receive_response() and the request to responder's receive_request() — both tasks become ready
4. If executor polls responder first: activate_firmware_rsp() in fd_context.rs checks get_fd_state() != ReadyXfer — state is still Apply because initiator  hasn't executed process_apply_complete_rsp() yet → returns InvalidStateForCommand

Branch: main-2.1
commit id : 47fd0460c05425a15e

mlvisaya:
…ks during FW activate

The responder task may process ActivateFirmwareRequest before the initiator task processes the ApplyComplete response (which transitions FD state from Apply to ReadyXfer). This causes the responder to reject the request with InvalidStateForCommand, leading to UA timeout.

Fix by detecting the pending Apply→ReadyXfer transition in activate_firmware_rsp: if the state is Apply and the ApplyComplete request was sent with a success result, perform the state transition inline since the UA has already acknowledged it (evidenced by the ActivateFirmwareRequest arriving).

Also make process_apply_complete_rsp tolerant of the state having already advanced past Apply, treating it as a no-op when the responder already performed the transition.

Fixes: chipsalliance/caliptra-mcu-sw#1764

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1828,6 +1828,7 @@ dependencies = [
  "embassy-executor",
  "embassy-sync",
  "embedded-alloc",
+ "futures",
 ]
 
 [[package]]
--- a/runtime/userspace/api/pldm-lib/Cargo.toml
+++ b/runtime/userspace/api/pldm-lib/Cargo.toml
@@ -24,6 +24,9 @@ caliptra-mcu-userlog.workspace = true
 [target.'cfg(not(target_arch = "riscv32"))'.dependencies]
 caliptra-mcu-libtock_unittest.workspace = true
 
+[dev-dependencies]
+futures.workspace = true
+
 [target.'cfg(target_arch = "riscv32")'.dependencies]
 embassy-executor = { version = "0.9.1", features = ["arch-riscv32"] }
 embedded-alloc.workspace = true
--- a/runtime/userspace/api/pldm-lib/src/cmd_interface.rs
+++ b/runtime/userspace/api/pldm-lib/src/cmd_interface.rs
@@ -99,24 +99,35 @@ impl<'a> CmdInterface<'a> {
             return Ok(());
         }
 
-        // Send the request
-        transport
-            .send_request(ua_eid, &msg_buf[..req_len + reserved_len])
-            .await
-            .map_err(|_| errors::TRANSPORT_ERROR)?;
+        // Mark that a response is pending. The responder task will wait
+        // on the signal in wait_for_initiator() before processing commands
+        // that depend on state transitions the initiator has not performed yet.
+        self.fd_ctx.set_pending_response(true);
 
-        // Wait for and process the response
-        transport
-            .receive_response(msg_buf)
-            .await
-            .map_err(|_| errors::TRANSPORT_ERROR)?;
+        let result = async {
+            // Send the request
+            transport
+                .send_request(ua_eid, &msg_buf[..req_len + reserved_len])
+                .await
+                .map_err(|_| errors::TRANSPORT_ERROR)?;
 
-        let payload = extract_pldm_msg(msg_buf).map_err(|_| errors::UTIL_ERROR)?;
+            // Wait for and process the response
+            transport
+                .receive_response(msg_buf)
+                .await
+                .map_err(|_| errors::TRANSPORT_ERROR)?;
 
-        // Handle the response
-        self.fd_ctx.handle_response(payload).await?;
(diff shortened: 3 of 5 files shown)


helloxiling (commented):
@mlvisaya, the analysis and code changes are sound. Can we add the integration test to repro the issue? Then we can prove that the fix solved the problem.

mlvisaya:
@helloxiling Done — I've pushed a follow-up commit that adds:

**Unit tests emulating the two-task race condition:**
- Tests simulate both execution orderings (responder polled first vs initiator polled first) to reproduce the exact race from issue #1764.
- Verified that commenting out the fix causes the race condition test to fail with `InvalidStateForCommand` (0x84), proving the test catches the bug.

**Additional race condition fix identified from the PLDM spec (DSP0267 v1.3.0, Table 15):**
- `update_component_rsp` has the same vulnerability in multi-component flows: after `ApplyComplete(Success)` for component N, the UA immediately sends `UpdateComponent` for component N+1. Same race, same fix applied.
- Updated `process_apply_complete_rsp` to also tolerate `Download` state (reachable when the responder processes `UpdateComponent` before the initiator).

All 16 tests pass (`cargo test -p caliptra-mcu-pldm-lib`).

helloxiling (approved):
Looks good to me.

PR #1782, opened 2026-07-19
Measurement API: Add production attestation evidence path

### LINKED ISSUE #1677: Measurement API: encode_measurement_evidence() and read_measurement()
**caliptra-mcu-sw / Measurement API:** implement `encode_measurement_evidence(buffer)` so the OCP EAT encoder can provide a buffer and receive the CBOR-encoded concise evidence triple array. 
Measurement API owns integrator config iteration and DPE/PCR reads. Also keep `read_measurement(fw_id)` as a test/diagnostic/internal helper.

### LINKED ISSUE #1678: Measurement API - read_attestation_target_handle() and OCP EAT generation/transport integration
**caliptra-mcu-sw / Measurement API + OCP EAT encoder:** implement `read_attestation_target_handle()` to select the configured AK context handle (defaulting to MCU Runtime). 
Update OCP EAT generation to call `encode_measurement_evidence(buffer)`, embed the concise evidence, and use the selected target handle to trigger Caliptra Core's signing path. 
SPDM is one transport path.

parvathib:
Add the Measurement API evidence-generation boundary, including DPE tagged-TCI reads, normalized measurement values, selected-AK lineage filtering, generated concise-evidence templates, and buffer-too-small behavior.

Add a shared attestation-evidence crate that wraps Measurement API concise evidence in OCP EAT claims, reads MCI debug status at runtime, signs via Measurement API selected-AK wrappers, and exposes PCR quote as a transport-neutral evidence format.

Wire SPDM GET_MEASUREMENTS through spdm-pal providers for signed OCP EAT and PCR quote formats, keep user-app provider selection at initialization, and allow SPDM measurement records to use provider-returned actual lengths within advertised maximum sizes.

Keep optional/private EAT claims and legacy caliptra-api evidence cleanup as follow-up work; emulator/end-to-end coverage remains in the next PR.

Fixes #1677 
Fixes #1678


runtime/userspace/api/attestation-evidence/src/ocp_eat.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn write_bstr(&mut self, bytes: &[u8]) -> McuResult<()> {
+        self.write_type_value(2, bytes.len() as u64)?;
+        self.write(bytes)
+    }
+
+    fn write_uint(&mut self, value: u8) -> McuResult<()> {
+        self.write_type_value(0, u64::from(value))
+    }
+
+    fn write_type_value(&mut self, major_type: u8, value: u64) -> McuResult<()> {
```
mhatrevi, line 206: This branch appears to encode CBOR values from 24 through 255 incorrectly. The first `write()` already emits both the additional-info byte and the value byte, but the second `write()` appends the value again:

```suggestion
            self.write(&[major | 24, value as u8])
```

This can produce malformed OCP EAT claims when the concise-evidence byte string is 24–255 bytes long. Could we also add a decoding test with evidence in that size range to cover this CBOR boundary?

runtime/userspace/api/attestation-evidence/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+    let payload_len = ocp_eat::encode_claims_payload(eat_payload, nonce, concise_evidence)?;
+    let payload = eat_payload
+        .get(..payload_len)
+        .ok_or(mcu_error::codes::INTERNAL_BUG)?;
+
+    let signed_eat = SignedEat::new();
+    let mut sig_digest = [0u8; ATTESTATION_P384_DIGEST_SIZE];
+    signed_eat
+        .sig_context_digest(alloc, payload, &mut sig_digest)
+        .await?;
```
helloxiling: Can we evaluate if a potential race condition can occur?  
`kid` and evidence generation release the Measurement API lock before this await, and `sign()` reacquires it afterward. Could measurement state change during that gap once SPDM responder and hitless component updates run concurrently? If so, the token could contain a `kid` for one AK state but be signed by another. Should this sequence be made atomic or explicitly gated until measurement updates complete?

runtime/userspace/api/attestation-evidence/src/ocp_eat.rs
parvathib: Fixed the  24..=255  encoding to emit only  `[major | 24, value]`  and added a 24-byte concise-evidence decode/round-trip unit test.

runtime/userspace/api/attestation-evidence/src/lib.rs
parvathib: Agree! Added `measure_and_sign_evidence()` so signed OCP EAT holds the Measurement API lock across `kid`, evidence encoding, digest construction, and signing.

helloxiling (approved):
Looks good to me. 
Please look into the spdm integration test failure.

runtime/userspace/api/spdm-lib/stack/src/measurements.rs
```diff
@@ -143,12 +142,12 @@ pub(crate) async fn handle_get_measurements_req<'a, Pal: SpdmPal>(
     } else {
         0
     };
-    let spdm_len = SpdmMsgHdrPdu::SIZE + body_len_without_sig + signature_len;
+    let spdm_max_len = SpdmMsgHdrPdu::SIZE + body_len_without_sig + signature_len;
```
mhatrevi, line 145: Since `MeasurementInfo::value_size` is now an advertised maximum, should the normal-vs-chunked response decision be based on the provider’s actual encoded length instead?

Currently, `spdm_max_len` can exceed `effective_data_transfer_size()` even when the actual evidence would fit in a normal response. In that case, we enter the large-response path before calling the provider and may reject a requester without chunking support unnecessarily.

Could we generate the measurement record into maximum-capacity storage first, then select the normal or chunked response path using the resulting actual length? A test where the advertised maximum exceeds the transfer size but the returned value fits would help cover this case.

mhatrevi (approved):

runtime/userspace/api/spdm-lib/stack/src/measurements.rs
parvathib: Agree! I'll address this in the follow up PR.

PR #1784, opened 2026-07-20
[ci]: Signal that the smoke test is optional

clundin25:


zhalvorsen (approved):

zhalvorsen:
Is the reason we don't make this test blocking because it is flaky?

clundin25:
> Is the reason we don't make this test blocking because it is flaky?

I think it could lead to a deadlock, if we can't update either MCU or SW independently

PR #1824, opened 2026-07-22
Fix VirtualMuxAlarm stale guards race condition

mlvisaya:
Add CaliptraVirtualAlarm crate as a drop-in replacement for Tock's
VirtualMuxAlarm. The fix addresses two issues:

1. InternalTimers::get_alarm() now stores and returns the absolute fire
   time (fire_at field). This fixes Guard 1: expired timers correctly
   report a past time instead of a bogus future time (now + bound).

2. Guard 2 (next_tick_vals within_range check) is removed. It cannot
   distinguish 'alarm expired and being handled' from 'alarm expired
   but nobody noticed' when MIE=0 during kernel syscall handling.

Guard 1 is preserved — it correctly prevents overwriting a sooner
pending alarm with a later one, relying on MuxAlarm::alarm()'s rescan
to catch the deferred alarm.

Also adds test-alarm-race-condition FPGA integration test that verifies
the hardware is correctly reprogrammed when a second alarm is set while
a first alarm's timer has expired without being processed.

veer/timers: Fix race in 64-bit cycle counter read

The now() implementation reads mcycleh and mcycle in two separate CSR
reads. If mcycle wraps between the reads, the combined 64-bit value
is incorrect (hi is stale relative to lo). Use the standard retry
loop: read mcycleh, read mcycle, re-read mcycleh; if it changed,
retry.


mlvisaya:
replaced by: https://github.com/chipsalliance/caliptra-mcu-sw/pull/1836

PR #1828, opened 2026-07-22
[provisioning] add Manuf fuse provisioning binary

### LINKED ISSUE #872: [provisioning] develop reference fuse configuration MCU programs

amit-kumarh:
and an accompanying integration test


provisioning/fuses/test.hjson
```diff
@@ -44,19 +44,19 @@
...
   CPTRA_SS_OWNER_PK_HASH: "0x0",
   CPTRA_SS_OWNER_PQC_KEY_TYPE: "0x0",
   CPTRA_SS_OWNER_PK_HASH_VALID: "0x0",
-  CPTRA_CORE_VENDOR_PK_HASH_1: "0x0",
-  CPTRA_CORE_PQC_KEY_TYPE_1: "0x0",
-  CPTRA_CORE_VENDOR_PK_HASH_2: "0x0",
-  CPTRA_CORE_PQC_KEY_TYPE_2: "0x0",
-  CPTRA_CORE_VENDOR_PK_HASH_3: "0x0",
-  CPTRA_CORE_PQC_KEY_TYPE_3: "0x0",
+  CPTRA_CORE_VENDOR_PK_HASH_1: "0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f30",
```
timothytrippel, line 54: could you make use of the key files i checked in here: https://github.com/chipsalliance/caliptra-sw/pull/4025?

provisioning/fuses/test.hjson
```diff
@@ -44,19 +44,19 @@
   // VENDOR_TEST_PARTITION - Skipped; for integrator use.
 
   // VENDOR_HASHES_MANUF_PARTITION
-  CPTRA_CORE_VENDOR_PK_HASH_0: "0x0",
-  CPTRA_CORE_PQC_KEY_TYPE_0: "0x0",
+  CPTRA_CORE_VENDOR_PK_HASH_0: "0x0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f30",
```
timothytrippel, line 47: see comment below

provisioning/fuses/test.hjson
```diff
@@ -30,7 +30,7 @@
   CPTRA_SS_TEST_UNLOCK_TOKEN_5: "0x5",
   CPTRA_SS_TEST_UNLOCK_TOKEN_6: "0x6",
   CPTRA_SS_TEST_UNLOCK_TOKEN_7: "0x7",
-  CPTRA_SS_TEST_EXIT_TO_MANUF_TOKEN: "0x8",
+  CPTRA_SS_TEST_EXIT_TO_MANUF_TOKEN: "0x6b530a63566dd030f97a4360506f5f9c",
```
timothytrippel: did you just randomly generate this?

tests/integration/src/jtag/test_provisioning.rs
```diff
@@ -63,8 +86,105 @@ mod test {
...
+            ProvisioningStage::TestUnlocked,
+        );
+
+        let jtag_params = JtagParams {
+            openocd: PathBuf::from("openocd"),
+            adapter_speed_khz: 1000,
+            log_stdio: true,
+        };
+
+        let mut core_tap = model
```
timothytrippel, line 125: do you think we should wrap the manuf debug unlock operation in a helper function like we do with the prod debug unlock code in `hw/model/src/debug_unlock.rs`?

timothytrippel (approved):

tests/integration/src/jtag/test_provisioning.rs
amit-kumarh: good idea, done.

provisioning/fuses/test.hjson
amit-kumarh: ah, yeah I did, was trying stuff in the process of debugging that maybe 8 didn't work for some reason. reverted.

provisioning/fuses/test.hjson
amit-kumarh: Spoke offline - this will require some extra work to make happen since we need to assemble this into a PK hash. Perhaps we have some tooling that allows specifying the key files in this JSON, and then assembles the binary from that. Regardless, outside the scope of the bare metal binary, so filed an issue for it

PR #1836, opened 2026-07-23
Fix Veer Timer Alarm

mlvisaya:
The Tock OS `VirtualMuxAlarm` virtualizer has an optimization in `set_alarm()` (Guard 1) that skips reprogramming the hardware timer when it believes an existing alarm will fire sooner. On VeeR EL2 FPGA, this optimization fails because VeeR's `get_alarm()` implementation returns a **bogus value** for expired timers:

- When the internal timer expires, `mitcnt0` halts at `mitb0` (counter == bound).
- The original `get_alarm()` computes the fire time as `now + bound`. For any timer (expired or not), this returns a value **slightly in the future** (`now + bound`). For an expired timer with a small bound (e.g., 1000 ticks), this is `now + 1000` — which always falls within the new alarm's `[reference, expiration)` range.
- Guard 1 interprets this as "the existing alarm fires sooner than ours" and **skips reprogramming**.

The fix stores the absolute fire time (`reference + dt`) in a `fire_at` field when `set_alarm()` is called. `get_alarm()` returns this stored value, which correctly represents a **past** time for expired timers. Guard 1 then sees it's not within the new alarm's range and allows the reprogram.


firmware-bundler/reference/fpga/user-app.toml
```diff
@@ -34,5 +34,5 @@ stack = 0x0_7000
 
 [[app]]
 name = "user-app"
-stack = 0x8000
+stack = 0x10000
```
parvathib: This is a huge bump for the user-app stack. why do we need it?

runtime/kernel/veer/src/timers.rs
```diff
@@ -168,27 +180,36 @@ impl<'a> time::Alarm<'a> for InternalTimers<'a> {
...
         self.mitcnt0.set(0);
         self.mitb0.set(val);
         self.enable_timer0();
     }
 
     fn get_alarm(&self) -> Self::Ticks {
-        let bound = self.mitb0.read(mitb0::bound) as u64;
-        let now = self.now().into_u64();
-        (now + bound).into()
+        // Return the absolute fire time stored by set_alarm().
```
parvathib, line 208: NIT: can we also clear `fire_at` in `disarm()` so that the subsequent `get_alarm()` doesn't get a stale value? This is not look like a correctness bug, but to keep the invariant consistent.

firmware-bundler/reference/fpga/user-app.toml
mlvisaya: This was because firmware update test was never included in the FPGA test, so it was fine working with a 0x8000 stack. Once the firmware update test was enabled, the PLDM firmware update code path exercises its full stack depth.

Currently emulator user-app also has 0x10000 stack allocation.

runtime/kernel/veer/src/timers.rs
mlvisaya: that makes sense. I'll add it.

firmware-bundler/reference/fpga/user-app.toml
parvathib: Makes sense for the PLDM path. Is 0x10000 (64K) a measured peak or are we just matching to the emulator? On main branch the FPGA user-app is tuned to 0x9e00 (~39.5 KB) and the same fw-update path fits, so are we may be over-allocating ~24 KB here?

Not blocking for this PR, but with the image-loader/firmware-update move to the scratch allocator coming, many of these buffers could come from scratch instead of the stack, so we should be able to trim this back and reduce stack usage in this path where we can.

mlvisaya: You're right. After testing with 0x9e00, the firmware update still passes. I'll bring it down to 0x9e00.

mlvisaya: by the way, looks like another PR increased the stack size to 0x10000 (https://github.com/chipsalliance/caliptra-mcu-sw/pull/1843). I'll still see if 0x9e00 works.

parvathib: Sounds good. Going forward, let's see if using the scratch can reduce stack size and thus keep the bss size under control.

parvathib (approved):

mlvisaya:
thanks

PR #1840, opened 2026-07-23
[update] rom: recover locked DOT blob from backup before reset

### LINKED ISSUE #1737: feat: Attempt DOT locked recovery (backup blob) before entering reset-recovery flow
## Current Behavior
In rom/src/cold_boot.rs, when a DOT blob is empty/corrupt and dot_fuses.is_locked(), the ROM currently chooses between two exclusive modes controlled by params.dot_recovery_reset_flow:
- true: maybe_enter_dot_recovery_reset_failure_flow sets DEVICE_STATUS_0 and calls fatal_error. attempt_dot_locked_recovery (interactive I3C) is never reached.
- false: maybe_enter_dot_recovery_reset_failure_flow is a no-op, and attempt_dot_locked_recovery is tried.

## Limitation
When dot_recovery_reset_flow = true, the ROM skips attempt_dot_locked_recovery entirely, including the BackupBlobRecoveryHandler which reads a pre-stored backup DOT blob from a known flash offset. That backup blob is already present in flash and could repair the device on the same boot, but the ROM fatals without ever consulting it.
The result is an unnecessary extra boot cycle: ROM fatals -> BMC cold-resets -> BMC sets RESET_CTRL -> ROM boots again. If the backup blob is valid, all of that is avoidable.

mhatrevi:
Attempt local backup recovery before entering the BMC reset flow for an empty or corrupt locked DOT blob. Revalidate the restored active blob and continue the current cold boot on success, while preserving failure status reporting when recovery fails.

### DIFF
--- a/rom/src/cold_boot.rs
+++ b/rom/src/cold_boot.rs
@@ -80,6 +80,91 @@ fn maybe_enter_dot_recovery_reset_failure_flow(
     fatal_error(err);
 }
 
+/// Attempts a configured, noninteractive backup-blob recovery.
+///
+/// On success, re-reads the active copy and runs the normal DOT flow so the
+/// current cold boot can continue with the recovered owner.
+fn attempt_dot_backup_recovery(
+    env: &mut RomEnv,
+    dot_fuses: &crate::DotFuses,
+    params: &RomParameters,
+    dot_flash: &dyn crate::hil::FlashStorage,
+    key_type: CmStableKeyType,
+) -> Option<caliptra_mcu_error::McuResult<Option<OwnerPkHash>>> {
+    if params.dot_recovery_policy != crate::DotRecoveryPolicy::BackupBlob {
+        return None;
+    }
+    let recovery_handler = params.dot_recovery_handler?;
+
+    Some(
+        device_ownership_transfer::dot_recovery_flow(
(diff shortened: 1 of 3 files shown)


rom/src/cold_boot.rs
```diff
@@ -80,6 +80,91 @@ fn maybe_enter_dot_recovery_reset_failure_flow(
...
+            let blob: DotBlob = transmute!(blob_bytes);
+            device_ownership_transfer::dot_flow(env, dot_fuses, &blob, key_type)
+        }),
+    )
+}
+
+/// Handles an empty or corrupt DOT blob while the device is locked.
+///
+/// With reset recovery enabled, a configured local backup is attempted first.
+/// If it cannot restore a valid active blob, ROM reports the failure to the BMC
```
helloxiling, line 121: nit: I think it might be better to use  “OCP Recovery Agent” instead of “BMC” here to keep common ROM code documentation platform-neutral.

helloxiling (approved):
Looks good to me.

rom/src/cold_boot.rs
mhatrevi: I'll update this in my upcoming DOT doc update PR.

Issue #1842, opened 2026-07-24
feat: hand off derived stable owner key CMD across ROM to RT


wlibo-mrvl:
# Summary
When the stable-owner-key ROM feature is enabled, `derive_stable_owner_key()` (`rom/src/stable_owner_key.rs`) reads a personalization seed from OTP and issues CM_DERIVE_STABLE_KEY to Caliptra Core, which returns an opaque 128-byte CMK (Caliptra Cryptographic Mailbox Key) handle. Currently that handle is discarded at the call site (`rom/src/cold_boot.rs`) — nothing hands it to Runtime, so RT has no way to reuse the same derived key.

# Proposed change
Extend the ROM→RT handoff table with the CMK, gated behind the stable-owner-key feature (mirroring the existing ocp-lock/hek_state pattern):

1. romtime/src/handoff.rs — add `stable_owner_cmk: [u32; 32]` and `stable_owner_cmk_valid: u32` to `RomHandoffTable`, each with a size-matched `#[cfg(not(feature = "stable-owner-key"))]` reserved companion so the struct layout — and the runtime block's offset — stays identical regardless of whether the feature is enabled. Add `HandoffData::store_stable_owner_cmk(&[u32; 32])`, called after the table's initial write() (the CMK isn't known yet at that point — it's derived later in cold boot), patching in just the CMK + validity flag without disturbing the rest of the table.
2. rom/src/cold_boot.rs — capture `Ok(cmk)` from `derive_stable_owner_key()` (currently discarded) and call the new updater.
3. Feature plumbing — forward stable-owner-key from `rom/Cargo.toml` down to caliptra-mcu-romtime, and from each runtime crate (platforms/common, plus whichever platform runtime(s) need it) down to romtime/platforms-common, since ROM and RT are separate build invocations and must agree on the feature to agree on struct layout.

PR #1844, opened 2026-07-24
[rom]: Hand off derived owner key CMK from ROM to Runtime

wlibo-mrvl:
Resolve https://github.com/chipsalliance/caliptra-mcu-sw/issues/1842

Extend the shared ROM/RT handoff table so the CMK derived by derive_stable_owner_key() during cold boot survives past ROM instead of being discarded, letting Runtime observe it via the .handoff DCCM region.

- romtime: add stable_owner_cmk/stable_owner_cmk_valid to RomHandoffTable, gated on the stable-owner-key feature with size-matched reserved companions (mirrors the existing ocp-lock/hek_state pattern) so the struct layout is identical whether the feature is on or off. Add HandoffData::store_stable_owner_cmk(), called after write() since the CMK isn't known until later in cold boot.
- rom: capture the derived Cmk at the cold_boot.rs call site instead of discarding it, and call store_stable_owner_cmk() on success.
- Wire the stable-owner-key feature through romtime -> rom -> platforms/common -> emulator/fpga runtime Cargo.tomls (ROM and Runtime are separate builds, so both need it enabled to agree on layout).
- Add a println in both emulator and fpga board.rs reading stable_owner_cmk_valid on boot (not the raw key material).
- Add a test-cmk-handoff feature + verification block in both board.rs files, and a new integration test (test_cmk_handoff_integrity) confirming Runtime observes a valid CMK after ROM derivation.


wlibo-mrvl:
duplicate with https://github.com/chipsalliance/caliptra-mcu-sw/pull/1852

PR #1845, opened 2026-07-24
[ci]: Fix main-2.1 nightly dispatch

v-focutter:
Consolidate the main-2.1 nightly workflows into the existing scheduler on the default branch.

The existing `nightly-build-test-2-1.yml` loaded `build-test.yml` from `main-2.1` as a reusable workflow, but `actions/checkout` still checked out the caller's `main` commit. The scheduled runs therefore mixed the main-2.1 workflow definition with the main workspace and failed while building xtask.

This change:
- dispatches both `build-test.yml` and `nightly.yml` with `ref: main-2.1`;
- monitors both downstream runs and propagates failures;
- removes the need for a duplicate `nightly-2-1.yml` dispatcher;
- uses `actions/github-script@v9`.

Validation:
- `actionlint .github/workflows/nightly-build-test-2-1.yml`
- `cargo xtask precheckin`

### DIFF
--- a/.github/workflows/nightly-build-test-2-1.yml
+++ b/.github/workflows/nightly-build-test-2-1.yml
@@ -1,12 +1,121 @@
 # Licensed under the Apache-2.0 license
 
-name: Nightly Build and Test (main-2.1)
+name: Nightly Workflows (main-2.1)
 
 on:
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:
 
+permissions:
+  actions: write
+  contents: read
+
 jobs:
-  build-test:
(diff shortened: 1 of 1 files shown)


mlvisaya (approved):

clundin25:
Duplicate of https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/.github/workflows/nightly-build-test-2-1.yml?

ArthurHeymans:
Addressed the duplicate-workflow concern by consolidating this into `.github/workflows/nightly-build-test-2-1.yml`. The existing reusable-workflow call was also checking out `main`, so the updated scheduler now dispatches both workflows with `ref: main-2.1` and monitors their results.

Issue #1825, opened 2026-07-22
[OCP LOCK]: Add command to get DPE endorsement certificate


clundin25:
We need to update the tests to verify the OCP LOCK endorsement cert is signed by the Caliptra DPE command.

MCU needs to supply this cert since it's the SoC is not allowed to call DPE directly or export CDI

clundin25:
This is completed

Issue #1732, opened 2026-07-01
[OCP-LOCK]: Add ML-DSA support


clundin25:
We need to support wielding ML-DSA keys from DPE.

clundin25:
Just need to sign the attestation token too

PR #1850, opened 2026-07-27
[OCP-LOCK]: Add MCU command to get DPE signer certificate

clundin25:
This is needed to verify the signature on the endorsement certificate and the epoch key report.

This resolves: https://github.com/chipsalliance/caliptra-mcu-sw/issues/1825

### DIFF
--- a/firmware-bundler/reference/fpga/user-app.toml
+++ b/firmware-bundler/reference/fpga/user-app.toml
@@ -34,5 +34,5 @@ stack = 0x0_7000
 
 [[app]]
 name = "user-app"
-stack = 0x9e00
-grant_space = 0x4000
+stack = 0xE000
+grant_space = 0x3000
--- a/runtime/userspace/api/caliptra-api/src/certificate.rs
+++ b/runtime/userspace/api/caliptra-api/src/certificate.rs
@@ -18,11 +18,12 @@ use caliptra_mcu_libsyscall_caliptra::mailbox::{Mailbox, MailboxError, PayloadSt
 use caliptra_mcu_libsyscall_caliptra::DefaultSyscalls;
 use caliptra_mcu_libtock_platform::ErrorCode;
 use dpe::commands::{
-    CertifyKeyCommand, CertifyKeyFlags, CertifyKeyP384Cmd, Command, CommandHdr,
-    GetCertificateChainCmd, SignCommand, SignFlags, SignP384Cmd,
+    CertifyKeyCommand, CertifyKeyFlags, CertifyKeyP384Cmd, Command, CommandHdr, DeriveContextCmd,
+    DeriveContextFlags, GetCertificateChainCmd, SignCommand, SignFlags, SignP384Cmd,
 };
 use dpe::context::ContextHandle;
-use dpe::response::SignP384Resp;
+use dpe::response::{DeriveContextExportedCdiResp, SignP384Resp};
+use dpe::tci::TciMeasurement;
 use zerocopy::{FromBytes, FromZeros, IntoBytes, TryFromBytes};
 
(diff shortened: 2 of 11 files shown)


tests/integration/src/runtime/test_mcu_mailbox.rs
```diff
@@ -142,3 +143,134 @@ fn test_fe_prog_authorized_req() -> Result<()> {
...
+    let issuer_cn_str = std::str::from_utf8(issuer_cn.data().as_slice())
+        .expect("Issuer Common Name should be valid UTF-8");
+    assert_eq!(
+        issuer_cn_str, "Caliptra 2.1 Ecc384 Rt Alias",
+        "DPE leaf certificate Issuer CN should match expected 'Caliptra 2.1 Ecc384 Rt Alias'"
+    );
+
+    // Verify leaf certificate was signed by a key in the DPE certificate chain
+    let mut verified = false;
+    for chain_cert in &chain_certs {
```
zhalvorsen: Can we assume it was signed by the Runtime alias instead of iterating through all of them?

runtime/userspace/api/caliptra-api/src/certificate.rs
```diff
@@ -186,6 +187,49 @@ impl CertContext {
         }
     }
 
+    // TODO(clundin): Retain the exported_cdi handle.
+    pub async fn dpe_signer_context_cert(&mut self, cert: &mut [u8]) -> CaliptraApiResult<usize> {
+        let cmd = DeriveContextCmd {
+            handle: ContextHandle::default(),
```
zhalvorsen: Should this function take a context handle? Or do we have a trait integrators can fill in the context handle for? I'm not sure we can guarantee it will always be the default context

runtime/userspace/api/caliptra-api/src/certificate.rs
```diff
@@ -186,6 +187,49 @@ impl CertContext {
         }
     }
 
+    // TODO(clundin): Retain the exported_cdi handle.
+    pub async fn dpe_signer_context_cert(&mut self, cert: &mut [u8]) -> CaliptraApiResult<usize> {
```
zhalvorsen, line 195: Do we need a similar function but for ML-DSA?

clundin25: ML-DSA tracked in: https://github.com/chipsalliance/caliptra-mcu-sw/issues/1732

runtime/userspace/api/caliptra-api/src/certificate.rs
clundin25: Good point, should make it a function param, but I'd prefer to do in a follow up to keep the diff churn down

zhalvorsen: That's cool with me

tests/integration/src/runtime/test_mcu_mailbox.rs
clundin25: Will do!

firmware-bundler/reference/fpga/user-app.toml
```diff
@@ -34,5 +34,5 @@ stack = 0x0_7000
 
 [[app]]
 name = "user-app"
-stack = 0x9e00
-grant_space = 0x4000
+stack = 0xE000
```
zhalvorsen, line 37: Can you expand the emulator as well so stay consistent?

runtime/userspace/api/caliptra-api/src/certificate.rs
clundin25: Tracking in: https://github.com/chipsalliance/caliptra-mcu-sw/issues/1900

zhalvorsen (approved):

PR #1852, opened 2026-07-27
[update] rom: hand off stable owner key CMK to runtime

### LINKED ISSUE #1842: feat: hand off derived stable owner key CMD across ROM to RT

mhatrevi:
Store the derived encrypted CMK in an append-only ROM-owned handoff extension and publish it with a validity marker. Add version-aware Runtime access, ABI layout checks, integration coverage, and handoff documentation.

### DIFF
--- a/docs/src/handoff.md
+++ b/docs/src/handoff.md
@@ -1,0 +1,75 @@
+# Firmware Handoff
+
+MCU ROM passes boot state to MCU Runtime through `HandoffData`, a shared
+structure in a `NOLOAD` DCCM region. The platform reserves 1 KiB for this
+structure at a fixed address supplied to both ROM and Runtime linker scripts.
+The handoff is volatile: Runtime may use it after the ROM-to-Runtime reset, but
+software must not treat it as persistent storage across a power cycle.
+
+## Ownership
+
+`HandoffData` begins with two fixed-size tables:
+
+| Offset | Size | Owner | Purpose |
+|---:|---:|---|---|
+| 0 | 64 bytes | ROM | Data produced by ROM and consumed by Runtime |
+| 64 | 64 bytes | Runtime | Data produced or updated by Runtime |
+| 128 | 132 bytes | ROM | Stable owner key CMK extension |
+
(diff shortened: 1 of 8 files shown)


clundin25 (approved):

docs/src/handoff.md
clundin25, line 1: We should probably document the OCP LOCK data in this too

PR #1859, opened 2026-07-28
fix(size): reduce .bss by moving MldsaVerifyReq to shared static

mlvisaya:
The MldsaVerifyReq struct (~11 KB) was previously a local variable in the async verify_authorized_signatures() function. Embassy stores async task futures as static allocations, causing this buffer to be duplicated in every task that calls it (MCU mailbox + SPDM VDM), inflating .bss by ~18 KB.

Move MldsaVerifyReq to a shared static Mutex<RefCell<MaybeUninit<...>>> that is populated inside a critical section and referenced across the await point. This is safe because the cooperative single-threaded executor guarantees no re-entrancy between the fill and the send.

Also pass mldsa_pub key by reference (&[u8; 2592]) instead of by value to avoid copying 2.6 KB into the async future state.

Additionally fix caliptra-util-host build: update fe_prog() to accept &HybridSignature (matching the FeProgRequest struct change from #1744) and add the caliptra-mcu-mbox-common dependency.

Measured .bss reduction: 126,228 -> 107,604 bytes (-18.2 KB). All-features release overflow reduced from 44 KB to 24 KB.

### DIFF
--- a/platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/device_ops.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/device_ops.rs
@@ -12,6 +12,8 @@ use caliptra_mcu_libsyscall_caliptra::mailbox::{Mailbox, MailboxError};
 use caliptra_mcu_libsyscall_caliptra::DefaultSyscalls;
 use caliptra_mcu_libtock_platform::ErrorCode;
 use caliptra_mcu_mbox_common::messages::{HybridSignature, AUTH_CMD_NONCE_LEN};
+use core::cell::RefCell;
+use critical_section::Mutex;
 use mcu_caliptra_api_lite::{
     fe_prog, get_attested_csr_ecc384, get_attested_csr_mldsa87, get_idev_csr_ecc384,
     request_debug_unlock_challenge, rng_generate, ApiAlloc, McuErrorCode,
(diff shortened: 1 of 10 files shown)


platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/device_ops.rs
```diff
@@ -148,28 +150,49 @@ pub async fn verify_authorized_signatures(
...
+        req.message = [0u8; caliptra_api::mailbox::MAX_CMB_DATA_SIZE];
+        req.message[..message.len()].copy_from_slice(message.as_slice());
+        req.as_bytes_partial_mut()
+            .map(|b| b.as_ptr() as usize..b.as_ptr() as usize + b.len())
+    });
+    let mldsa_req_range = mldsa_req_bytes.map_err(|_| CaliptraCompletionCode::OperationFailed)?;
+    // SAFETY: the static buffer is not mutated while we hold this slice —
+    // cooperative executor ensures no re-entrancy between the critical
+    // section above and the await below.
+    let mldsa_req_slice = unsafe {
```
helloxiling: A hardening suggestion:  Can we keep explicit ownership of  `MLDSA_REQ`  across the mailbox  `.await`  instead of reconstructing a mutable slice after the  `RefCell`  borrow has ended? Although this executor is single-threaded,  .await  still permits task interleaving. 

The current implementation is safe only because the separate global  MAILBOX_MUTEX  prevents another path from reaching this buffer. That is a non-local invariant, and a future mailbox/control-flow change could allow overlapping mutable slices or overwrite the request while Caliptra is consuming it.

mlvisaya: Good catch — agreed the non-local invariant is fragile. Addressed in 0c287f98a: replaced `critical_section::Mutex<RefCell<...>>` with `embassy_sync::mutex::Mutex<CriticalSectionRawMutex, ...>`. The async `MutexGuard` is now held across the `.await`, so:

- No more unsafe raw pointer-to-slice reconstruction
- No reliance on a separate MAILBOX_MUTEX for safety
- The borrow checker enforces exclusive access across the await point

Thanks for the review!

helloxiling (approved):
Looks good to me.

PR #1860, opened 2026-07-28
ci(size-history): track all-features in release size measurements

mlvisaya:
Update cargo xtask ci size-history to build with all-features enabled (--features all-features,release --no-default-features) so the CI size report reflects the full production feature set (SPDM, PLDM, streaming-boot, flash-boot, firmware-update, DOE, userspace-log).

Add passthrough feature flags (all-features, streaming-boot, flash-boot, firmware-update, doe, userspace-log) to both emulator and FPGA kernel Cargo.toml so the bundler can pass --features all-features without 'feature not found' errors on the kernel crate.

The linker layout step may fail if the binary currently overflows the 512 KB SRAM budget — this is tolerated because the individual ELFs are still produced by cargo and the size-history tool only needs those to measure .text/.stack section sizes.


mlvisaya:
CI size-history is already compiling with all features

PR #1862, opened 2026-07-28
spdm: raise DOE DataTransferSize floor to 4096 bytes

TimMarkey:
## Summary

The DOE transport (`runtime/userspace/api/spdm-lib/transports/src/doe.rs`) capped `MAX_SPDM_MTU` at **1024 bytes**. That sizes the SPDM responder's advertised `DataTransferSize` below:

- what post-quantum signatures require — an ML-DSA-87 signature is ~4.6 KiB, and
- the OCP 2.7 SPDM-14 minimum `DataTransferSize` of 4 KiB.

This raises the cap to **4096**.

It also upgrades the driver-reported `max_message_size` check from `debug_assert!` to `assert!`, so a driver that reports a sub-4-KiB size fails loudly in **release** builds rather than silently under-sizing `DataTransferSize` (the `debug_assert!` is stripped out of release).

## Change

```rust
const MAX_SPDM_MTU: usize = 4096; // was 1024
assert!(
    max >= MAX_SPDM_MTU + DOE_HEADER_SIZE,
    "DOE driver max_message_size must be >= 4 KiB + DOE header for DataTransferSize"
);
```

## Testing

`cargo clippy` and `rustfmt` are clean on the touched crate (`caliptra-mcu-spdm-transports`). The full `cargo xtask precheckin` could not complete in my local environment due to an unrelated `provisioning/fuses` build-script tooling gap (not touched by this change) — relying on CI as the authoritative gate.


TimMarkey:
This is my first PR to this repo, so the CI workflow needs a maintainer to approve the run. Could one of you kick it off when you have a moment? Thanks! @clundin25 @swenson @zhalvorsen

TimMarkey:
Amended the commit: author email is now the id-based GitHub noreply address (resolves to @TimMarkey) and the unresolvable co-author trailer has been removed. recheckcla

clundin25:
@TimMarkey triggered CI!

RaunakGu (approved):

RaunakGu:
@parvathib, Could you please review this one, thank you

parvathib (changes requested):
Two changes are needed before this should merge:

1. `runtime/userspace/api/spdm-lib/transports/src/doe.rs:91`: Please make the SPDM transfer size an integrator/profile configuration rather than a universal DOE transport invariant. `mtu()` becomes `DataTransferSize`, while the stack independently advertises `MaxSPDMmsgSize` from `large_capacity()` and already supports CHUNK_SEND/CHUNK_GET. Generic SPDM integrations can therefore use a smaller transfer buffer and chunk larger messages; OCP NVMe v2.7 integrations can select 4096. The current `assert!` turns a valid smaller-buffer integration (and any max-size syscall failure) into a release panic. Prefer passing a configured payload limit into `McuSpdmDoeTransport`, bounding it by the driver-reported payload capacity, and validating an OCP-specific minimum once during initialization.

2. `runtime/userspace/api/spdm-lib/transports/src/doe.rs:99`: Raising `mtu()` to 4096 also needs corresponding allocator sizing/accounting. The DOE responder has a 12 KiB shared pool. During CHUNK_GET, a persistent large-message buffer must coexist with the request, hash state, and a `DOE_HEADER_SIZE + mtu` response allocation (~4.1 KiB). An 8 KiB buffered response therefore cannot allocate its next chunk, even though CAPABILITIES can advertise `MaxSPDMmsgSize` near the allocator's full `large_capacity()`. Please reserve transfer-buffer headroom when deriving the advertised large-message capacity, or increase/configure the pool, and cover the 4 KiB-MTU chunking path.

Chunking does not waive the OCP SPDM-14 requirement for an OCP-compliant profile, but that profile policy should not be hard-coded into the reusable DOE transport.

TimMarkey:
@parvathib, thanks for the review — addressed both in the latest commit.

DataTransferSize as configuration: the reusable DOE transport no longer carries any OCP policy and never panics. mtu() reports the configured DataTransferSize bounded by the driver max via a pure bound_mtu() (unit-tested for small/zero driver maxima). The OCP 4 KiB minimum is now an integrator choice, selected and enforced once at init in the app behind a Cargo feature (spdm-doe-ocp-transfer-size), matching how skip-doe-spdm gates other integrator/profile decisions.

Allocator headroom: the advertised/effective MaxSPDMmsgSize now reserves room for one in-flight CHUNK_GET exchange (two header + mtu frames + scratch) via usable_large_capacity(), so a buffered large response can always allocate its next chunk. When the pool can't hold one frame after that reservation, CHUNK is stripped from the advertised capabilities rather than mis-sized.

Rather than bumping the pool unconditionally, the larger 20 KiB pool is a second opt-in feature (spdm-large-scratch). The reference build stays compact by default — no added BSS for integrators who don't need 4 KiB chunking — and the two features are enabled together for an OCP build. The 4 KiB-MTU chunking path is now covered by tests.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
```diff
@@ -178,11 +194,45 @@ async fn spdm_mctp_responder() {
 async fn spdm_doe_responder() {
     let mut cw = Console::<DefaultSyscalls>::writer();
 
-    let doe_transport = McuSpdmDoeTransport::new(doe::driver_num::DOE_SPDM);
-    if !doe_transport.exists() {
-        crate::log_info!(cw, "SPDM_DOE: No DOE device, exiting");
-        return;
-    }
+    // OCP 2.7 SPDM-14 requires a 4 KiB DataTransferSize. That minimum is
```
parvathib: Thanks for the detailed comment. Could you also include a link to the OCP Datacenter NVMe specification in this comment?

runtime/userspace/api/spdm-lib/transports/src/doe.rs
```diff
@@ -45,22 +45,46 @@ const DOE_TYPE_SPDM: u8 = 1;
 /// Data object type for SPDM Secured Messages.
 const DOE_TYPE_SECURE_SPDM: u8 = 2;
 
+/// Default SPDM `DataTransferSize` (payload bytes) when a caller uses
```
parvathib, line 48: Thanks for finding the hard-coded transfer limit and for addressing the allocator/headroom concern. I agree this should be configurable by the integrator, with `OCP NVMe v2.7` being one integration profile.

Could we keep the default SPDM payload size at 1024 bytes and allow integrators to request a larger size through  with_transfer_size() ? The effective MTU would be:

`min(configured_transfer_size, DOE capsule max_message_size - DOE header size)`

An OCP NVMe v2.7 integration can configure 4096 and verify that the DOE capsule supports at least 4104 total bytes. This keeps the generic default unchanged while ensuring the configured size never exceeds the capsule capacity

runtime/userspace/api/spdm-lib/stack/src/stack.rs
```diff
@@ -230,6 +230,32 @@ impl<S, L> ConnectionState<S, L> {
     }
 }
 
+/// Scratch the responder must keep free alongside the large-message buffer for
+/// non-chunk allocations (peak ~2.4 KB during certify_key kid computation).
+const LARGE_MSG_SCRATCH_RESERVE: usize = 2 * 1024;
```
parvathib: Thanks for adding the allocator headroom. 
Since each integrator configures the memory budget through the user app/PAL, could we remove the fixed `LARGE_MSG_SCRATCH_RESERVE`,  `large_msg_headroom()`, and `usable_large_capacity()` policy from the generic stack? 
Instead, we should have the user-app configure the safe persistent large-message capacity in the PAL and make `pal.large_capacity()` return that capacity bounded by the allocator’s available space and use it directly for CHUNK advertisement and validation. 

Perhaps the reference user-app should include an allocator test showing that its configured maximum large response can complete `CHUNK_GET` without allocation failure. Integrators should apply equivalent validation to their customized configurations.

platforms/emulator/runtime/userspace/apps/user/Cargo.toml
```diff
@@ -82,6 +82,16 @@ ocp-lock = [
     "mcu-mbox-service",
 ]
 doe = []
+# spdm-doe-ocp-transfer-size: configure the DOE responder for OCP 2.7 SPDM-14's
```
parvathib: Since this `user-app` is only a reference implementation and integrators are expected to customize its transport and memory configuration, the  `spdm-doe-ocp-transfer-size`  and  `spdm-large-scratch  feature` flags are unnecessary. Each integrator can configure the intended transfer size and scratch allocation directly in its application, while the reusable libraries expose the necessary configuration APIs. 

If the reference app needs to demonstrate an OCP NVMe 2.7 configuration, could we group all relevant settings/configs under a single user-app profile feature instead?

runtime/userspace/api/spdm-lib/stack/src/stack.rs
TimMarkey: Hi parvathib, I've moved the large-message capacity policy out of the generic stack. The pal.large_capacity() now returns the safe, allocator-bounded value (raw largest-free-run minus the CHUNK_GET headroom: 2*(header+mtu) + 2 KiB scratch reserve), and the stack consumes it directly for CHUNK advertisement and buffered large-response/request validation. I removed LARGE_MSG_SCRATCH_RESERVE, large_msg_headroom(), and usable_large_capacity() from the stack — the headroom reservation is now an integrator/PAL memory-budget decision.

The stack keeps only the protocol math (.max(mtu) flooring and min(local, peer) negotiation). By construction large_capacity() returns the same value the old usable_large_capacity(pal) computed for a given allocator + transport geometry, so the advertised MaxSPDMmsgSize and CHUNK advertise/strip thresholds are byte-identical to before the PR.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
TimMarkey: Update: we opted to drop the OCP spec link. Since this transport library isn't OCP-SSD-specific, the comment now states the ≥ 4 KiB DataTransferSize as a generic buffer-sizing requirement, with the OCP justification tracked in the ticket instead. The ocp-nvme-profile reference-app feature still documents the OCP intent for that integration.

runtime/userspace/api/spdm-lib/transports/src/doe.rs
TimMarkey: Done. The generic default is back to 1024 bytes (DEFAULT_TRANSFER_SIZE); new() uses it, integrators opt into a larger size via with_transfer_size(). The effective MTU is computed as you suggested — min(configured_transfer_size, DOE capsule max_message_size − DOE header size). The OCP NVMe v2.7 profile configures 4096 and checks the capsule supports at least 4104 total bytes at init.

platforms/emulator/runtime/userspace/apps/user/Cargo.toml
TimMarkey: I dropped the separate spdm-doe-ocp-transfer-size and spdm-large-scratch flags and grouped all the OCP NVMe v2.7 reference settings (4 KiB DataTransferSize + the larger scratch pool) under a single ocp-nvme-profile user-app feature. The reusable libraries expose the configuration APIs (with_transfer_size(), the allocator pool size) directly, so an integrator configures its own transport/memory without these flags — the feature is purely the reference app's OCP demonstration profile.

TimMarkey:
Rebased this branch onto the latest main-2.1 (cd29575f) to clear the failing analyze check.

The failure wasn't from this change — the branch was forked before main-2.1 gained the xtask ci size-history command, so the analyze job's cargo xtask ci size-history step was failing with error: unrecognized subcommand 'ci' at argv-parse time (before any size measurement runs). Rebasing onto current main-2.1 picks up xtask/src/ci.rs and resolves it.

The rebase was conflict-free. No source changes beyond the replay; the only difference in commit count is that my earlier smoke-test workflow fix dropped out as a no-op since main-2.1 already carries that exact change. Host tests (spdm-lib stack, pal) and fmt are green on the rebased tree.

TimMarkey:
@parvathib I've addressed all four review items — would you mind taking another look 
when you get a chance?  Thanks!

TimMarkey:
Hi @parvathib, following up on PR #1862. I addressed the four review items, rebased onto current main-2.1, and required CI is green. Could you please take another look when convenient? Thanks!

parvathib:
> Hi @parvathib, following up on PR #1862. I addressed the four review items, rebased onto current main-2.1, and required CI is green. Could you please take another look when convenient? Thanks!

Thanks, @TimMarkey. I’ll make sure to review this PR by Monday (08/17) at the latest.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
TimMarkey: Providing the spec reference here rather than in the code comment. On the review feedback that the reusable spdm-lib comments should stay vendor-neutral, the OCP-specific naming was removed from the shared code and the profile behavior is gated behind the `ocp-nvme-profile` cargo feature. The DataTransferSize requirement it satisfies is SPDM-14 in the OCP Datacenter NVMe SSD Specification: https://www.opencompute.org/documents/datacenter-nvme-ssd-specification-v2-7-final-pdf-1 — the PR description carries the same rationale.

TimMarkey:
Hi @parvathib — friendly ping on this one. Since your last review the branch has been reconciled onto upstream `main-2.1` (adopting the configurable `max_spdm_msg_size` / `large_capacity()` model), all four earlier review items are addressed, and the OCP-specific naming was moved out of the reusable spdm-lib code per the vendor-neutral feedback (spec link provided in the mod.rs thread above). CI is running on the current head now. When you get a chance, could you take another look and, if the concerns are resolved, dismiss the earlier `CHANGES_REQUESTED`? Thanks!

TimMarkey:
Hi @parvathib, your changes requested is blocking this PR, could you re-approve?  Thanks.

TimMarkey:
CI status — the red checks are not caused by this PR.

spdm_validator_tests: failures are DMTF conformance case 5.9 (spdm_test_case_certificate_large_invalid_request) — the responder must answer a LargeCertChain GET_CERTIFICATE with ERROR(UnsupportedRequest) when LARGE_RESP_CAP wasn't advertised. This PR only raises the DOE DataTransferSize and touches no certificate-handler code, so it doesn't affect 5.9. It's a pre-existing main-2.1 gap fixed by #2089, and this check will pass here once #2089 merges.

Smoke Test (Optional) group: non-blocking. test_artifacts_subsystem (1–4) were cancelled (shown as "fail" in the checks list), upstream builds passed, and print_test_results failed only on the missing junit.xml — a cascade of the cancellation, not a real test failure.

Net: no functional failure in this PR; the 5.9 validator red is external and resolves with #2089.

PR #1863, opened 2026-07-28
ci(size-history): add App .bss and SRAM overflow to size report

mlvisaya:
Add two new columns to the size-history CI report:

- App .bss size: tracks the .bss section of user-app, which contains static buffers (SPDM scratch, async task futures, crypto key storage) that are the main contributor to SRAM pressure.

- SRAM overflow: reports how many bytes the combined kernel + user-app exceeds the 512 KB FPGA SRAM budget (0 when it fits). This makes it immediately visible whether a PR pushes the binary over budget.

Bump CACHE_FORMAT_VERSION to v5 so historical commits are re-measured with the new metrics.

### DIFF
--- a/xtask/src/ci.rs
+++ b/xtask/src/ci.rs
@@ -10,7 +10,7 @@ use size_history::{
 use std::path::{Path, PathBuf};
 use std::{env, error::Error, fs, io};
 
-const CACHE_FORMAT_VERSION: &str = "v4";
+const CACHE_FORMAT_VERSION: &str = "v5";
 
 pub const MCU_KERNEL_FPGA: FwId = FwId {
     crate_name: "mcu-runtime-fpga",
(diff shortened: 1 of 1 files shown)


xtask/src/ci.rs
```diff
@@ -189,3 +207,99 @@ impl ArtifactBuilder for CaliptraElfSizeGenerator {
...
+        let target_dir = workspace.join("target");
+
+        match build_runtime(&target_dir) {
+            Ok(_) => Ok(0), // Fits in SRAM
+            Err(e) => {
+                let msg = format!("{:?}", e);
+                // Parse "Bytes <needed> would exceed remaining memory space <available>"
+                if let Some(overflow) = parse_overflow_from_error(&msg) {
+                    Ok(overflow)
+                } else {
```
helloxiling: If the build error is not the recognized SRAM-overflow error, should this return `None`  instead of reading ELFs? The shared target directory may contain stale artifacts, and this approximation omits bundler alignment/TBF overhead, so an unrelated failure could be cached as a valid but incorrect overflow value.

mlvisaya: Good point. Removed the ELF-based fallback — `measure()` now returns `None` (reported as "build error" in the table) when the failure isn't the recognized SRAM-overflow message. This avoids caching an incorrect value from stale artifacts.

v-focutter (approved):

helloxiling (approved):
Looks good to me.

Issue #1833, opened 2026-07-22
[xtask] Enable offline signing flows for SoC auth manifests


timothytrippel:
Currently the xtask flow for building SoC auth manifests uses fake keys hardcoded in the repo. To support downstream integrations that would use offline signing flows, we need to enable building unsigned auth manifests that can be signed offline with an HSM.

timothytrippel:
# Offline Auth Manifest Signing Flow

This document specifies the CLI extensions and offline signing workflow for the
Caliptra Authorization Manifest (`AuthorizationManifest`).

## Overview

The [`AuthorizationManifest`](./image_loading.md) is a critical security
structure verified by Caliptra ROM/runtime during boot. It contains metadata
and SHA-384 cryptographic digests for the MCU Runtime binary and optional
SoC firmware components.

This design decouples manifest creation from private key signing. It enables
building an unsigned manifest in CI, extracting signature targets, computing
signatures in an isolated environment (such as an offline signing workstation
or key vault), and reattaching signatures to produce a final signed binary.

## Manifest Signature Targets

An `AuthorizationManifest` contains four signature blocks in its preamble
over three distinct payload regions:

- **Vendor Public Keys Signature**:
  - **Signer**: Vendor FW Key (Root Key anchored in hardware OTP fuses).
  - **Payload Region**: `AuthManifestPreamble::vendor_signed_data_range()`
    (covers `version`, `svn`, `flags`, and `vendor_pub_keys`).
  - **Digest Target**: `SHA384(version || svn || flags || vendor_pub_keys)`.
  - **Preamble Field**: `preamble.vendor_pub_keys_signatures`.
- **Owner Public Keys Signature**:
  - **Signer**: Owner FW Key (Root Key anchored in hardware OTP fuses/registers).
  - **Payload Region**: `AuthManifestPreamble::owner_pub_keys_range()`
    (covers `owner_pub_keys`).
  - **Digest Target**: `SHA384(preamble.owner_pub_keys)`.
  - **Preamble Field**: `preamble.owner_pub_keys_signatures`.
- **Image Metadata Collection (IMC) Signatures**:
  - **Signer**: Vendor Manifest Key (Operational key in `vendor_pub_keys`)
    and Owner Manifest Key (Operational key in `owner_pub_keys`).
  - **Payload Region**: `image_metadata_col` (Image Metadata Collection).
  - **Digest Target**: `SHA384(image_metadata_col)`.
[...]

PR #1866, opened 2026-07-29
[builder] support unsigned manifest generation and offline signing

timothytrippel:
Extend CaliptraBuilder to support loading public keys (ECC, LMS, and ML-DSA) via AuthManifestPubKeysPaths and generating unsigned authorization manifest templates with fallback key warnings.

Also add caliptra_sw_workspace_root helper in builder/src/lib.rs to locate test keys checked into caliptra-sw for unit testing.

This partially addresses #1833.

### DIFF
--- a/builder/src/caliptra.rs
+++ b/builder/src/caliptra.rs
@@ -3,6 +3,7 @@
 //! Wrappers around the Caliptra builder library to make it easier to build
 //! the ROM, firwmare, and SoC manifest.
 
+use crate::offline_signing::{create_signing_request, SigningRequestJson};
 use crate::target_dir;
 use anyhow::{bail, Context, Result};
 use caliptra_auth_man_gen::{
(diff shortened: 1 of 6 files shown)


builder/src/caliptra.rs
```diff
@@ -893,6 +911,181 @@ fn main() -> Result<()> {
...
+
+        let gen_config = AuthManifestGeneratorConfig {
+            vendor_fw_key_info: Some(vendor_fw_key_info),
+            vendor_man_key_info: Some(vendor_man_key_info),
+            owner_fw_key_info: Some(owner_fw_key_info),
+            owner_man_key_info: Some(owner_man_key_info),
+            image_metadata_list,
+            version: 1,
+            flags: AuthManifestFlags::VENDOR_SIGNATURE_REQUIRED,
+            pqc_key_type: FwVerificationPqcKeyType::LMS,
```
amit-kumarh: Should we be able to select MLDSA keys as well? or is that coming later down the pipeline

builder/src/offline_signing.rs
```diff
@@ -51,6 +51,67 @@ pub struct SignatureEntry {
     pub pqc_sig: Option<String>,
 }
 
+fn sha384_hex(data: &[u8]) -> String {
+    use sha2::{Digest, Sha384};
```
amit-kumarh: nit: move to top of file? (same throughout)

runtime/userspace/api/caliptra-api/src/crypto/aes_gcm.rs
```diff
@@ -443,7 +443,7 @@ impl AesGcm {
         let mut req = CmAesGcmDecryptFinalReq {
             hdr: MailboxReqHeader::default(),
             context,
-            tag_len: tag.len() as u32,
+            tag_len: (tag.len() * 4) as u32,
```
amit-kumarh: consider `size_of` ?

amit-kumarh:
also - might be worth separating into one commit that bumps caliptra-sw (and performs all the necessary refactorings) and one that actually does the unsigned manifest and signing request generation?

timothytrippel:
> also - might be worth separating into one commit that bumps caliptra-sw (and performs all the necessary refactorings) and one that actually does the unsigned manifest and signing request generation?

That is how it is currently implemented.

builder/src/caliptra.rs
timothytrippel: Its coming later; but made an option here to better allude to that

runtime/userspace/api/caliptra-api/src/crypto/aes_gcm.rs
timothytrippel: This is no longer relevant; accidently revved caliptra-sw commit to a 2.1 commit instead of a 2.0 commit.

builder/src/offline_signing.rs
timothytrippel: done

clundin25 (approved):

amit-kumarh:
Ah - I'm just dumb. thought I was viewing the latest commit and was actually viewing the whole PR. thanks!

PR #1874, opened 2026-07-29
[OCP LOCK]: epoch key attestation report command

clundin25:


tests/integration/src/test_ekp.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    const MAILBOX_TIMEOUT_CYCLES: u64 = 20_000_000;
+
+    /// Execute a mailbox command with a large timeout and return the raw
+    /// response bytes.
+    fn mailbox_execute_with_timeout(
+        hw: &mut impl McuHwModel,
+        cmd: u32,
+        payload: &[u8],
+    ) -> Result<Option<Vec<u8>>, String> {
```
clundin25: TODO: This should be moved to the test integration lib.

runtime/userspace/api/caliptra-api/src/ocp_lock.rs
```diff
@@ -515,6 +520,124 @@ impl<'a> OcpLock<'a> {
...
+            HekSeedState::Unused,
+            HekSeedState::Unused,
+            HekSeedState::Unused,
+            HekSeedState::Unused,
+            HekSeedState::Unused,
+            HekSeedState::Unused,
+            HekSeedState::Unused,
+        ];
+
+        for i in 0..num_slots {
```
zhalvorsen: We should probably do some sanity checks on `num_slots`

runtime/userspace/api/caliptra-api/src/ocp_lock.rs
```diff
@@ -515,6 +520,124 @@ impl<'a> OcpLock<'a> {
...
+            let reg = caliptra_mcu_libsyscall_caliptra::otp::reg::LOCK_HEK_PROD_ALL[i];
+            let mut slot_bytes = [0u8; 48];
+            for word_idx in 0..12 {
+                let word = otp_driver
+                    .read(reg, word_idx as u32)
+                    .map_err(CaliptraApiError::Syscall)?;
+                slot_bytes[word_idx * 4..(word_idx + 1) * 4].copy_from_slice(&word.to_le_bytes());
+            }
+            hek_state_list[i] = HekSeedState::from_seed_bytes(&slot_bytes, active_state);
+        }
```
zhalvorsen: nit: can we use an iterator here instead:


```suggestion
        for slot in hek_state_list.get(0..num_slots)?.iter_mut() {
            let reg = caliptra_mcu_libsyscall_caliptra::otp::reg::LOCK_HEK_PROD_ALL[i];
            let mut slot_bytes = [0u8; 48];
            for word_idx in 0..12 {
                let word = otp_driver
                    .read(reg, word_idx as u32)
                    .map_err(CaliptraApiError::Syscall)?;
                slot_bytes[word_idx * 4..(word_idx + 1) * 4].copy_from_slice(&word.to_le_bytes());
            }
            *slot = HekSeedState::from_seed_bytes(&slot_bytes, active_state);
        }
```

zhalvorsen (approved):

runtime/userspace/api/caliptra-api/src/ocp_lock.rs
clundin25: Yes good catch, updated!

runtime/userspace/api/caliptra-api/src/ocp_lock.rs
clundin25: Done!

zhalvorsen (approved):

PR #1883, opened 2026-07-30
ci(xtask): add stack-guard to catch user-app stack-frame growth

### LINKED ISSUE #1885: Add CI stack-guard for user-app frame growth
CI cannot currently catch user-app stack-frame growth: `ci.rs::elf_stack_size` reads the `.stack` section size (the linker budget), not actual usage.

Add `cargo xtask stack-guard`: build with `-Z emit-stack-sizes`, read per-function frames, fail if the largest frame exceeds the `[[app]].stack` budget minus a reserve. Wired into precheckin. Fix in PR #1883.

RaunakGu:
Closes #1885

## What

Add `cargo xtask stack-guard`: fail CI if the user-app's stack usage approaches its budget — the
regression class (a large on-stack buffer creeping into an async task future) that caused the original
SPDM stack overflow. Wired into `precheckin`.

## Why

`ci.rs::elf_stack_size` reads the `.stack` **section** size — i.e. the budget the linker reserved
(`[[app]].stack`). It reports that number regardless of actual usage, so it can't catch per-function
frame growth, only an explicit budget edit.

## How

1. Build the bundle with `-Z emit-stack-sizes` (via `runtime_build_with_apps`, so the app is fully
   linked — a bare `cargo rustc -p user-app` would DCE it).
2. Read exact per-function frames from `.stack_sizes`.
3. Fail if the **largest single frame** > `[[app]].stack` − `STACK_RESERVE`.

Largest-frame + reserve (not a call-graph sum): a precise additive walk is unreliable here — the
executor dispatches task polls through indirect/`dyn` edges a static graph can't follow, and
`.stack_sizes` omits asm frames. The reserve bounds the frames stacked above the largest one plus an
undercount cushion.

## Calibration (measured, current main)

| | B |
|---|--:|
| Deepest additive chain (peak) | 13,520 |
| Largest frame (`spdm_mctp_responder::poll`) | 6,960 |
| `STACK_RESERVE` (gap 6,560 + ~3,680 cushion) | 10,240 (`0x2800`) |

Budget is read from the toml (single source of truth: the fpga manifest, with emulator required to
match), so the guard **self-adjusts** to whatever `[[app]].stack` is — no hard-coded number, works in
any merge order relative to a budget change.

## Tests

- 5 unit tests: ULEB128 decode + truncated/overlong rejection, record layout, `frame_limit`, and the
  gate (6,960 B passes; re-adding an 11 KB on-stack buffer trips the 10,240 B limit).
- Live self-test: passes on current main (frame 6,960 B ≤ limit); fails with exit 1 + actionable
  message when the budget is lowered below the frame.
- `fmt`/`clippy` clean. Touches only `xtask/src/{stack_guard.rs, main.rs, precheckin.rs}`.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -27,6 +27,7 @@ mod registers;
 mod rom;
 mod runtime;
 mod sizes;
+mod stack_guard;
 mod test;
 mod vertex_ai;
 
@@ -153,6 +154,10 @@ enum Commands {
         #[arg(long)]
         variants: Option<String>,
     },
+    /// Fail if the user-app's worst-case stack usage (from `.stack_sizes`)
+    /// approaches its `[[app]].stack` budget. Unlike the section-size check,
+    /// this catches per-function frame growth.
+    StackGuard,
     /// Build emulator binary and package it in emulators.zip
     EmulatorBuild {
         #[arg(long)]
@@ -674,6 +679,7 @@ fn main() {
                 )
             })
         }
+        Commands::StackGuard => stack_guard::run(),
         Commands::FlashImage { subcommand } => match subcommand {
             FlashImageCommands::Create {
                 caliptra_fw,
--- a/xtask/src/precheckin.rs
+++ b/xtask/src/precheckin.rs
@@ -51,5 +51,9 @@ pub(crate) fn precheckin(rom_variants: &[RomVariant]) -> Result<()> {
     crate::test::e2e_tests()?;
     crate::test::test_hello_c_emulator()?;
     crate::rom::build_all_variants(rom_variants)?;
+
+    // Fail if the user-app's largest stack frame approaches its budget. Catches
+    // the per-function frame growth the section-size check cannot see.
+    crate::stack_guard::run()?;
     Ok(())
 }
--- a/xtask/src/stack_guard.rs
+++ b/xtask/src/stack_guard.rs
@@ -1,0 +1,251 @@
+// Licensed under the Apache-2.0 license
+
+//! `cargo xtask stack-guard` — fail CI if the user-app's largest stack frame
+//! approaches its budget.
+//!
+//! The existing size check (`ci.rs::elf_stack_size`) reads the `.stack`
+//! *section* size, which is just the budget carved out by the linker — it
+//! reports the budget no matter how much stack the code actually uses, so it
+//! cannot catch frame growth. This guard instead builds the app with
+//! `-Z emit-stack-sizes`, reads the exact LLVM-emitted per-function frames, and
+//! fails if the largest frame exceeds `[[app]].stack` budget − `STACK_RESERVE`.
+//!
+//! Why the largest single frame (not a full call-graph sum): the user-app's
+//! peak is dominated by one monolithic async `poll` frame (the SPDM responder
+//! coroutine embeds its whole handler tree inline), so max-frame is the peak
+//! driver. A precise additive walk is unreliable here anyway — the executor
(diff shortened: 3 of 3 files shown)


RaunakGu:
@timothytrippel, @clundin25, Can you please review and help to merge this, thanks

mlvisaya:
Thanks for adding this—the `.stack_sizes` approach is useful and addresses a real blind spot in the existing `.stack` section-size check. I do have one concern with the threshold calibration before merge.

`frame_limit()` derives the allowed frame size as `manifest budget - 0x2800`, while the regression test hard-codes a `0x5000` budget. On this PR head, the emulator/FPGA manifests are `0xae80`, so the live guard reports:

- budget: 44,672 B
- limit: 34,432 B
- largest frame: 6,960 B

Reintroducing the motivating 11,324 B contribution would produce roughly 18,284 B (`6,960 + 11,324`), which is still well below the actual 34,432 B limit. Thus the regression passes the real guard even though the unit test says it should fail; the test only trips because it assumes the separate `0x5000` budget reduction.

After merging current `main`, the manifests are `0x6600`, giving a 15,872 B limit. Using the latest reported 4,384 B SPDM frame, the same modeled growth is 15,708 B, so it still narrowly passes. More generally, increasing the configured stack budget silently weakens the frame-regression detector.

Could we separate these into two checks?

1. A budget-independent maximum-frame baseline/growth threshold, intended specifically to catch large frame regressions.
2. A separate safety check comparing estimated/measured peak usage plus margin against the configured stack budget.

A secondary maintainability concern is that the instrumented build manually mirrors codegen flags from `.cargo/config.toml`; future target flag changes could silently make this build differ from production. Deriving those flags or validating that the mirror stays synchronized would make the check more robust.

Validation performed locally: all five focused `stack_guard` tests pass, and `cargo xtask stack-guard` passes with the values above.

mlvisaya:
Hi @RaunakGu , please find my review comments above. Let me know if they make sense or not. Thanks.

RaunakGu:
Thanks @mlvisaya, you're right on both counts.

On the calibration: I checked on current `main` — the largest frame is 4,320 B against a 15,872 B limit, so re-adding the 11,324 B buffer comes to 15,644 B and still passes. The unit test only disagreed because it hard-coded a `0x5000` budget from a companion PR that I later closed as superseded.

Both points are now fixed:
- The threshold is a fixed `MAX_SINGLE_FRAME` of 8 KiB and the manifest is no longer read, so raising the stack budget can't weaken the check. Verified the gate fires and that the 11 KB case now trips it.
- The codegen flags are read from `.cargo/config.toml` instead of copied by hand, so they can't drift from the shipping build.

I did not add the second check. The bundler already fails the build when the app doesn't fit ("Bytes N would exceed remaining memory space M") and `ci.rs` reports the overflow size, so a stack peak estimate would be the weakest of those signals. If an earlier margin warning is wanted, the size report seems like the better place for it.

Also rebased onto latest `main`. Let me know if you have further comments.

mlvisaya (approved):
looks good to me. please fix the CI failures. It might need rebase.

RaunakGu:
Thanks! All the CI failures are the Caliptra ROM helper resolving `textwrap@0.16.4` (rustc 1.88) on `main`'s 1.85 toolchain, so any PR on `main` hits them. #2122 fixes it; I checked locally that with it the helper resolves `textwrap 0.16.2` and builds clean on 1.85. I'll rebase once #2122 lands.

RaunakGu:
@mlvisaya - Please help to click merge, We don't have merge access right, Thanks

PR #1747, opened 2026-07-07
Add Measurement API and gated cold-boot/hitless boot init

parvathib:
Fixes #1674 
Implements measurement_boot_init on caliptra-mcu-measurement-api: cold boot initializes DPE/PCR stores and writes the MCU-RT root record; hitless update validates preserved state and fails closed. Adds dpe_rotate_context_default and dpe_tag_tci to caliptra-api-lite, and an ATTESTATION error domain.

App boot wiring is gated behind the off-by-default `measurement-boot-init` feature: cold boot rotates MCU-RT off the DPE default handle, which the current SPDM path still signs with, so it stays disabled until that path is rewired to track the rotated handle.

PR #1895, opened 2026-07-31
Cherry picks from main phase 1

mlvisaya:
Cherry-picked PRs:

- #1627 Size Tracking for Image and Stack sizes
- #1153 Already present in `main-2.1` (empty cherry-pick)
- #1649 Add tool to check vendor and owner PK hashes on firmware bundles
- #1726 Update GitHub Actions
- #1710 Use chunked CertifyKey for DPE leaf certificate
- #1750 Trigger bitstream build only on TCL file changes
- #1735 Allow overriding Caliptra ROM source ref
- #1754 Fix attestation workflow bundle build
- #1747 Add Measurement API and gated cold-boot/hitless boot initialization
- #1694 Implement streaming DebugUnlock and SET_CERTIFICATE
- #1761 Move MCU mailbox to scratch-backed lite APIs
- #1758 Use Measurement API for DPE handles


mhatrevi (approved):

.github/workflows/bitstream-build.yml
```diff
@@ -25,7 +26,7 @@ jobs:
         segmented: [true, false]
     steps:
       - name: Checkout repo
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
+        uses: actions/checkout@v7
```
helloxiling, line 29: Why do we change the checkout from an immutable SHA to `actions/checkout@v7`?

mlvisaya: Hi, this is from https://github.com/chipsalliance/caliptra-mcu-sw/pull/1726. I think the SHA was for v.6.0.2 which is not tagged since it's a patch. v7 is a tag so it points directly to a specific commit, therefore no need to specify SHA

clundin25: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
clundin25, line 229: These commands were unintentionally deleted. If possible chunk cherry-picks into smaller patches to improve odds of catching this

mlvisaya: that's a good catch! i'll re-place it back on the next set of cherry-picks

clundin25: No worries, I've added it back in: https://github.com/chipsalliance/caliptra-mcu-sw/pull/1874

PR #1906, opened 2026-08-05
Add production support for FirmwareVersion and DeviceCapabilities commands

helloxiling:
Source Caliptra Core version and capabilities from the Core mailbox, report MCU runtime version and the compiled capability bitmap over MCI mailbox and MCTP VDM, and retire the mock command handlers.


common/config/src/version.rs
```diff
@@ -0,0 +1,25 @@
+// Licensed under the Apache-2.0 license
+
+pub const MCU_RUNTIME_VERSION_MAJOR: u32 = 2;
+pub const MCU_RUNTIME_VERSION_MINOR: u32 = 1;
```
mlvisaya: for main branch, would the version be 2.0.0 ? and for main-2.1 branch, this would be 2.1.0?

common/config/src/capabilities.rs
```diff
@@ -0,0 +1,77 @@
+// Licensed under the Apache-2.0 license
+
+bitflags::bitflags! {
+    #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+    pub struct McuRuntimeCapabilities: u64 {
+        const FLASH_BOOT = 1 << 0;
+        const STREAMING_BOOT = 1 << 1;
+        const FIRMWARE_UPDATE = 1 << 2;
+        const SPDM_RESPONDER = 1 << 3;
+        const MCTP_VDM_RESPONDER = 1 << 4;
```
mlvisaya, line 10: should we include DOE in the capabilities?

mlvisaya (commented):
added some nits, but otherwise it looks good.

common/config/src/capabilities.rs
helloxiling: Good catch! We should add it.

common/config/src/version.rs
helloxiling: Yes, that's correct, and I misunderstood it earlier. Fixed it in latest commit.

mlvisaya (approved):
LGTM

caliptra-util-host/apps/mctp-vdm/config/src/lib.rs
```diff
@@ -50,9 +50,6 @@ pub struct ServerConfig {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct DeviceCapabilitiesConfig {
     pub capabilities: u32,
```
mlvisaya: nit: do we also need to extend this to [u8, 36] ?

mlvisaya (approved):
looks good to me

caliptra-util-host/apps/mctp-vdm/config/src/lib.rs
helloxiling: Good catch. This config field was left at the old scalar shape when the response changed to a 36-byte payload. Updated both MCTP VDM and mailbox configs to `[u8; 36]`; the mailbox mock server now emits the configured payload unchanged and its validator compares all 36 bytes. Added fixture parsing tests for both configs.

Issue #1849, opened 2026-07-27
MCTP Get Endpoint UUID (0x03) not implemented


yujieh-mrvl:
## Summary

The MCTP control message handler (`runtime/kernel/capsules/src/mctp/control_msg.rs`)
does not implement Get Endpoint UUID (command code `0x03`, DSP0236 Section 12.6).
When a BMC sends this command during endpoint enumeration, it falls through to the
`Unsupported` branch and returns completion code `0x05`. The BMC therefore cannot
persistently identify the endpoint across EID reassignments or hot-plug events.

## Proposed Change

Add Get Endpoint UUID support to the MCTP control command handler, sourcing the
16-byte UUID (RFC 4122) from a platform-configurable location:

1. **`runtime/kernel/capsules/src/mctp/control_msg.rs`** — Add `GetEndpointUUID = 3`
   variant to `MCTPCtrlCmd`, implement handler returning 1-byte completion code +
   16-byte UUID per DSP0236 Section 12.6. Add `uuid: &[u8; 16]` parameter to
   `process_mctp_control_msg`.

   <details><summary>Diff</summary>

   ```diff
    pub fn process_mctp_control_msg(
        msg_buf: &[u8],
        local_eid: u8,
        supported_msg_types: &[MessageType],
   +    uuid: &[u8; 16],
        resp_buf: &mut [u8],
    ) -> Result<MCTPCtrlMsgResp, ErrorCode> {
   ```

   ```diff
            MCTPCtrlCmd::GetEID => mctp_ctrl_cmd
                .process_get_endpoint_id(local_eid, rsp_payload)
                .map(|_| mctp_ctrl_cmd.resp_data_len()),
   +        MCTPCtrlCmd::GetEndpointUUID => mctp_ctrl_cmd
   +            .process_get_endpoint_uuid(uuid, rsp_payload)
   +            .map(|_| mctp_ctrl_cmd.resp_data_len()),
            MCTPCtrlCmd::GetVersionSupport => {
   ```

   ```diff
    pub enum MCTPCtrlCmd {
        SetEID = 1,
        GetEID = 2,
   -    GetMsgTypeSupport = 5,
   +    GetEndpointUUID = 3,
        GetVersionSupport = 4,
   +    GetMsgTypeSupport = 5,
        Unsupported = 0xFF,
    }
   ```

   ```diff
    impl From<u8> for MCTPCtrlCmd {
        fn from(val: u8) -> Self {
            match val {
[...]

PR #1907, opened 2026-08-05
Adding support for the MCTP Get Endpoint UUID message.

v-focutter:
As per task [1849](https://github.com/chipsalliance/caliptra-mcu-sw/issues/1849) adding support for the MCTP Get Endpoint UUID message.


mlvisaya (approved):
LGTM

PR #1910, opened 2026-08-05
common-commands: share authorized fuse operations across transports

ArthurHeymans:
## Summary

Refactor only — no wire-format or mailbox behavioral changes.

- Move vendor PK-hash provisioning, minimum Caliptra SVN increase, vendor public-key revocation, and vendor PK-hash-slot revocation out of private MCU mailbox handlers into the transport-neutral common command handler interface (`caliptra-common-commands`), alongside `program_field_entropy`.
- Keep MCU mailbox authorization and wire framing in the mailbox layer; delegate fuse policy to the emulator Caliptra command backend.
- All existing fuse policy is preserved unchanged: slot validity, idempotency, min-SVN bounds (nonzero, ≤128, ≤ running `FW_INFO.fw_svn`, no decrease), thermometer encoding, current-boot key/slot protection, and last-key-per-algorithm protection.

This is PR 1 of a 3-PR stack adding authorized fuse commands over SPDM VDM:
1. **#1910 (this PR)** — common policy extraction
2. #1911 — SPDM VDM responder subcommands
3. #1912 — host tooling + end-to-end validation

### Memory delta (`measure-memory.sh`, devel profile, vs `main`)

| Section | `main` | Stack head | Delta |
|---|---:|---:|---:|
| Kernel `.text`/`.data`/`.bss` | 150,392 / 36 / 24,536 | 150,392 / 36 / 24,536 | 0 |
| User-app FLASH | 260,384 | 267,252 | **+6,868 B** |
| User-app RAM | 78,136 | 78,776 | **+640 B** |
| SPDM task FLASH | 104,142 | 111,598 | **+7,456 B** |
| SPDM task RAM | 29,853 | 30,501 | **+648 B** |
| Runtime bundle | 412,160 | 419,072 | **+6,912 B** |

PR #1911, opened 2026-08-05
spdm-lite: add authorized fuse VDM commands

ArthurHeymans:
## Summary

Adds the authorized fuse subcommands and their host tooling to the SPDM VDM `AuthorizedCommand` path.

| Subcommand | ID | Command payload |
|---|---|---|
| ProvisionVendorPkHash | `0x5056_504B` (`PVPK`) | `slot:u32 | hash:[u8;48]` |
| FuseIncreaseCaliptraMinSvn | `0x4D43_4D53` (`MCMS`) | `flags:u32 | svn:u32` |
| FuseRevokeVendorPubKey | `0x4D52_564B` (`MRVK`) | `reserved:u32 | slot:u32 | key_type:u32 | key_index:u32` |
| FuseRevokeVendorPkHash | `0x5256_4B48` (`RVKH`) | `reserved:u32 | slot:u32` |

Requests use the command payload followed by the one-use challenge, ECC and ML-DSA public keys, and hybrid signature. The signature covers `cmd_id(BE) || payload || challenge`.

The responder verifies authorization before dispatching to the transport-neutral fuse operations. The host stack adds typed requests, explicit little-endian encoders, completion-code preservation, bounded large-request reassembly, and isolated validator suites for destructive operations.

Review feedback addressed:

- Advertise all six implemented authorized subcommands through `DeviceCapabilities`.
- Document the complete authorization trailer and byte ranges.
- Require MCMS `flags` and MRVK/RVKH `reserved` fields to be zero.
- Exercise real signed requests and reject tampered payloads, command IDs, public-key anchors, ECC signatures, and ML-DSA signatures.

### Memory delta (`measure-memory.sh`, devel profile, vs `main`)

| Section | `main` | Stack head | Delta |
|---|---:|---:|---:|
| Kernel `.text`/`.data`/`.bss` | 150,392 / 36 / 24,536 | 150,392 / 36 / 24,536 | 0 |
| User-app FLASH | 260,384 | 267,252 | **+6,868 B** |
| User-app RAM | 78,136 | 78,776 | **+640 B** |
| SPDM task FLASH | 104,142 | 111,598 | **+7,456 B** |
| SPDM task RAM | 29,853 | 30,501 | **+648 B** |
| Runtime bundle | 412,160 | 419,072 | **+6,912 B** |

### DIFF
--- a/docs/src/caliptra_spdm_vdm_cmds.md
+++ b/docs/src/caliptra_spdm_vdm_cmds.md
@@ -36,7 +36,6 @@ For the unified software architecture shared between OOB (SPDM VDM) and in-band 
 
 Caliptra commands assigned to SPDM VDM are carried within SPDM `VENDOR_DEFINED_REQUEST` and `VENDOR_DEFINED_RESPONSE` messages using the OCP-assigned Vendor ID (`42623`). The command range `0x01`-`0x20` is [reserved in the OCP registry](https://github.com/opencomputeproject/ocp-registry/blob/main/command-registry.md) and defined by the Caliptra Working Group.
 
-
 ### OCP VDM Header
 
 The SPDM VDM standard header identifies the vendor organization:
@@ -83,7 +82,7 @@ These command codes are assigned from the Caliptra range reserved in the [OCP co
 | `0x06`       | RequestDebugUnlock        | O   | Request debug unlock in production environment.                            |
 | `0x07`       | AuthorizeDebugUnlockToken | O   | Send debug unlock token to device for authorization.                       |
 | `0x08`       | ExportAttestedCsr         | O   | Export attested CSR for a Caliptra device identity key.                    |
-| `0x12`       | AuthorizedCommand         | O   | Carry authorization-gated subcommands. The SPDM authorization flow is TBD. |
+| `0x12`       | AuthorizedCommand         | O   | Carry challenge-authorized provisioning and fuse subcommands.              |
 
 R = Required, O = Optional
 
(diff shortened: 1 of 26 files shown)


ArthurHeymans:
This PR is part of a stack of 3 bookmarks:

1. `trunk()`
1. [fuse-ops-common](https://github.com/chipsalliance/caliptra-mcu-sw/pull/1910)
1. **spdm-vdm-fuse-cmds ← this PR**
1. [util-host-fuse-vdm](https://github.com/chipsalliance/caliptra-mcu-sw/pull/1912)

---
*Created with [jj-stack](https://github.com/keanemind/jj-stack)*

docs/src/caliptra_spdm_vdm_cmds.md
```diff
@@ -83,22 +82,54 @@ These command codes are assigned from the Caliptra range reserved in the [OCP co
...
 | `0x07`       | AuthorizeDebugUnlockToken | O   | Send debug unlock token to device for authorization.                       |
 | `0x08`       | ExportAttestedCsr         | O   | Export attested CSR for a Caliptra device identity key.                    |
-| `0x12`       | AuthorizedCommand         | O   | Carry authorization-gated subcommands. The SPDM authorization flow is TBD. |
+| `0x12`       | AuthorizedCommand         | O   | Carry challenge-authorized provisioning and fuse subcommands.              |
 
 R = Required, O = Optional
 
 Implemented commands are advertised in the `external_commands` field returned by `DeviceCapabilities`. Implemented operations under `AuthorizedCommand` are advertised individually in `authorized_subcommands`. Recognized but unimplemented commands are not advertised. See [Device Capabilities](caliptra_common_commands.md#device-capabilities).
 
 ## Authorization-Gated Subcommands
```
mhatrevi, line 91: The dispatcher now reports PVPK, MCMS, MRVK, and RVKH as `Supported`, but the runtime capability mask still advertises only `GET_AUTH_CHALLENGE` and `PROGRAM_FIELD_ENTROPY` when the `spdm` feature is enabled. As a result, a requester using `DeviceCapabilities.authorized_subcommands` will not discover these newly implemented commands.

Please update `authorized_subcommand_capabilities()` and its tests to advertise all four new subcommands, or keep their documented status as planned until capability discovery is updated.

docs/src/caliptra_spdm_vdm_cmds.md
```diff
@@ -83,22 +82,54 @@ These command codes are assigned from the Caliptra range reserved in the [OCP co
...
+4. Append `HybridSignature` (`ecc_sig_r[48] || ecc_sig_s[48] || mldsa_sig[4628]`) to the payload and submit it under `AuthorizedCommand`.
+
+The challenge is consumed by the verification attempt and cannot be reused. Requests must have the exact documented size; missing, truncated, and oversized signatures are rejected with `InvalidPayloadSize`, while failed authorization returns `AccessDenied`.
+
+### Implemented Subcommand Payloads
+
+Byte offsets below begin immediately after the four-byte `subcommand_id`.
+
+| Subcommand | Bytes | Field | Encoding |
+| ---------- | ----- | ----- | -------- |
```
mhatrevi, line 125: The payload table currently shows `HybridSignature` immediately after each command-specific payload, but the handler parses the request as:

```text
payload || nonce(48) || ecc_pub_x(48) || ecc_pub_y(48) || mldsa_pub(2592) || HybridSignature
```

For example, the PVPK signature is parsed after the nonce and public-key fields, not at byte offset 52 as implied by this table. Please document the complete authorization trailer and correct the byte ranges, or change the parser if the intended wire format is actually `payload || HybridSignature`.

The host and device implementations must use the same exact layout because the signature covers the raw payload and challenge.

platforms/emulator/runtime/userspace/apps/user/src/spdm/caliptra_vdm.rs
```diff
@@ -128,6 +127,68 @@ impl CaliptraVdmAuthorization for CaliptraVdmAuthorizationHook {
...
+            )
+            .await
+            .map_err(|_| CaliptraCompletionCode::AccessDenied)?;
+        CaliptraCmdBackend
+            .provision_vendor_pk_hash(slot, hash)
+            .await
+            .map_err(map_common_completion)
+    }
+
+    async fn increase_caliptra_min_svn<A: SpdmPalAlloc>(
```
mhatrevi, line 161: `flags` is part of the authenticated MCMS payload, but it is ignored by the platform hook and never passed to the transport-neutral operation.

Please either validate that `flags` is zero/reserved and reject unsupported bits, or propagate it to the operation. Silently accepting signed values that have no effect makes the wire contract ambiguous and could create policy surprises if flags gain semantics later.

platforms/emulator/runtime/userspace/apps/user/src/spdm/caliptra_vdm.rs
```diff
@@ -161,6 +222,71 @@ impl CaliptraVdmAuthorization for CaliptraVdmAuthorizationHook {
...
+            .verify_signatures(
+                REVOKE_VENDOR_PUB_KEY_CMD_ID,
+                payload,
+                nonce,
+                ecc_pub_x,
+                ecc_pub_y,
+                mldsa_pub,
+                sig,
+            )
+            .await
```
mhatrevi, line 255: The `reserved` fields for MRVK and RVKH are parsed and included in the signed payload, but arbitrary nonzero values are accepted and then discarded. The tests currently use nonzero values, which locks in this behavior.

If these fields are reserved, please require them to be zero and return `InvalidParameter` otherwise, with negative tests for nonzero values. If they are intended to carry semantics, please document those semantics instead.

runtime/userspace/api/spdm-lib/vdm-handler/src/iana/ocp/caliptra_vdm/mod.rs
```diff
@@ -22,6 +22,10 @@ pub use caliptra_mcu_spdm_codec::vendor_defined::iana::ocp::caliptra::{
     CaliptraCompletionCode, CaliptraVdmCmdResult, CaliptraVdmCommand, CaliptraVdmResult,
     CALIPTRA_VDM_COMMAND_VERSION, CALIPTRA_VENDOR_ID,
 };
+pub use commands::authorized_command::{
```
mhatrevi, line 25: The handler tests provide good coverage for parsing, request sizes, command binding, and challenge reuse, but they use a synthetic signature checker rather than the production ECC P-384 and ML-DSA-87 verification path.

Before merging, could we add at least one end-to-end SPDM/emulator test that obtains a real challenge, submits a real signed request for the new commands, and verifies rejection after tampering with the payload, command ID, public-key anchor, or either signature leg? It would also be useful for that test to assert that the corresponding authorized-subcommand capability bits are advertised.

mhatrevi (changes requested):

ArthurHeymans:
Addressed the review feedback and folded the host/validator commit from #1912 into this PR. The updated series advertises the new capabilities, documents the full authorization trailer, rejects nonzero reserved fields, and adds real-signature tamper coverage. `cargo xtask precheckin` passes.

mhatrevi (approved):

PR #1916, opened 2026-08-05
[dpe]: Add exported-cdi handle to dpe store

clundin25:
We need to sign OCP LOCK endorsement certificates with an exported-cdi since it can create a CA certificate

### DIFF
--- a/runtime/kernel/capsules/src/dpe_handle_store.rs
+++ b/runtime/kernel/capsules/src/dpe_handle_store.rs
@@ -11,26 +11,54 @@
 //!
 //! ## SRAM layout
 //!
-//! ### Header (72 bytes)
+//! ### Header (104 bytes)
 //!
 //! ```text
 //! offset  0 : magic                     u32 LE  (0xD9E4_C7A1)
 //! offset  4 : version                   u8      (1)
 //! offset  5 : _pad                      u8
-//! offset  6 : header_size               u16 LE  (72)
+//! offset  6 : header_size               u16 LE  (104)
 //! offset  8 : record_size               u16 LE  (DPE_HANDLE_RECORD_SIZE = 32)
 //! offset 10 : record_capacity           u16 LE
 //! offset 12 : record_count              u32 LE
 //! offset 16 : attestation_target_fw_id  u32 LE  (0xFFFF_FFFF = none)
 //! offset 20 : _pad                      [u8; 4]
 //! offset 24 : attestation_policy_digest [u8; 48]
-//! offset 72 : records[0..capacity]      DPE_HANDLE_RECORD_SIZE bytes each
+//! offset 72 : exported_cdi_handle       [u8; 32]
+//! offset 104: records[0..capacity]      DPE_HANDLE_RECORD_SIZE bytes each
 //! ```
 //!
 //! Records at indices 0..(record_count-1) are all valid.  The `flags` byte
 //! within each record (offset 28) is **reserved** — the capsule does not
 //! interpret it for validity.  Record validity is determined solely by
 //! position relative to `record_count`.
+//!
+//! ## Exported CDI Lifecycle
+//!
+//! The `exported_cdi_handle` field (offset 72) is used to retain the handle
+//! of the exported Compound Device Identifier (CDI) derived from DPE.
+//!
+//! ### Creation
+//! - Derived via a DPE `DeriveContext` command with `EXPORT_CDI` flag.
+//! - DPE generates the CDI and returns a 32-byte handle.
+//! - MCU Runtime stores this handle in this capsule.
+//!
+//! ### Usage
+//! - Subsequent signing operations (e.g., OCP LOCK) retrieve this handle and
+//!   pass it to Caliptra via `SignWithExportedEcdsa` to sign data using the CDI.
+//!
+//! ### Destruction, Invalidation & Revocation
+//! - **Cold Boot / Power Cycle**: SRAM is cleared/reinitialized, destroying the stored handle.
+//!   Caliptra internal state is also lost, invalidating the CDI.
+//! - **Caliptra Reset**: If Caliptra undergoes a cold reset, its internal state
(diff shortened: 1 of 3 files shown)


runtime/kernel/capsules/src/dpe_handle_store.rs
```diff
@@ -333,6 +341,30 @@ impl DpeHandleStore {
             .ok_or(ErrorCode::FAIL)?;
         self.copy_record_to_slice(index, slice)
     }
+
+    fn do_read_exported_cdi(&self, slice: &WriteableProcessSlice) -> Result<(), ErrorCode> {
```
zhalvorsen, line 372: unrelated comment: Why are all of these function names prepended with `do_`?

zhalvorsen (approved):

runtime/kernel/capsules/src/dpe_handle_store.rs
```diff
@@ -55,6 +56,8 @@
 //! | 5   | MARK_ATTESTATION_TARGET | fw_id | —             |
 //! | 6   | READ_ATTESTATION_TARGET | —     | RW 0 (output) |
 //! | 7   | VALIDATE_STORE          | —     | RO 0 (digest) |
+//! | 8   | READ_EXPORTED_CDI       | —     | RW 0 (output) |
+//! | 9   | WRITE_EXPORTED_CDI      | —     | RO 0 (input)  |
```
parvathib, line 87: Could you clarify the lifecycle of the exported CDI? Where is `write_exported_cdi()` called, and under what circumstances is the value updated or rotated?

clundin25: Done!

parvathib: Thanks! 
Is the plan for the Measurement API to own this handle and expose only purpose-specific operations, rather than allowing application tasks to call `READ_EXPORTED_CDI / WRITE_EXPORTED_CDI` directly?

It would be good to keep the raw handle internal to the Measurement API implementation.

clundin25: That sounds right. I need to do a bit more research on the Measurement API but it makes sense to me OCP LOCK would be a downstream consumer of it

clundin25: One thing to call out, it's a privileged API for PL0 and has features like acting as a self-signed CA.

With that context, does it still fit in the measurement API? Or should we be doing something for just MCU?

parvathib: Good point. I think the `exported-CDI lifecycle`, including its `internal read/write operations`, fits within the Measurement API because it already owns DPE Handle Storage. 

If CDI export changes any MCU-managed DPE handles, the Measurement API can update the store consistently. The only consumer would be the MCU OCP LOCK orchestration running within the Caliptra-core-authenticated MCU Runtime. The raw handle would remain internal, and the functionality can be gated by the 2.1  `ocp-lock` feature.

clundin25: Perfect, thanks! I will tag you in the PR that wires this to the OCP LOCK module once the work is completed

runtime/kernel/capsules/src/dpe_handle_store.rs
```diff
@@ -11,27 +11,55 @@
 //!
 //! ## SRAM layout
 //!
-//! ### Header (72 bytes)
+//! ### Header (104 bytes)
 //!
 //! ```text
 //! offset  0 : magic                     u32 LE  (0xD9E4_C7A1)
 //! offset  4 : version                   u8      (1)
 //! offset  5 : _pad                      u8
```
parvathib, line 19: For the DPE store header, should we repurpose `byte 5 (_pad)`  as compile-time feature flags? 
For example, bit 0 could be `EXPORTED_CDI_SUPPORTED`, with the remaining bits reserved. 
This would let us keep a common layout while distinguishing configurations if 2.0 and 2.1 share the codebase later.
```
flags = 0x00 // Exported CDI unsupported; field reserved/zero (for 2.0 /2.1 without OCP LOCK)
flags = 0x01 // Exported CDI enabled and used by this build/configuration
```

clundin25: Is it correct to say that a breaking change to this layout is only impactful to a hitless update?

parvathib: Yes, the layout compatibility concern currently applies when retained state is reused across a hitless firmware update, since cold boot initializes the DPE store. 
The flag also makes the common layout self-describing: when clear, the `exported_cdi_handle`  bytes are reserved and must be ignored; when set, the field is supported and may be interpreted. This prevents firmware with exported-CDI support from interpreting reserved bytes from a configuration without that support.

clundin25: How about binding the schema shape to the version? https://github.com/chipsalliance/caliptra-mcu-sw/blob/main-2.1/runtime/kernel/capsules/src/dpe_handle_store.rs#L72

That mirrors how it is handled in caliptra-sw, which uses the same layout irrelevant of what features are enabled.

parvathib: That makes sense.  We can bind each schema shape to a header version and keep the layout identical regardless of enabled features. The current 72-byte schema remains version 1, and the new 104-byte schema becomes version 2. 
Any future field that affects the header size or record offsets would require another version. Hitless updates across schema versions would fail closed unless explicit migration support is added.

clundin25: Great! I bumped the version. Also I want to clarify that the exported-cdi feature isn't really tied to any feature flags / build commands. It's also already present in 2.0, and we may want to support it in order to sign other certificates

parvathib (approved):

PR #1918, opened 2026-08-05
[ci] fix smoke-test workflow: drop continue-on-error on reusable-workflow job

### LINKED ISSUE #1915: caliptra-sw-smoke-test.yml fails to parse: continue-on-error is invalid on a reusable-workflow-call job
The Caliptra SW Smoke Test (Optional) workflow no longer parses, so PR fails with:

Invalid workflow file: .github/workflows/caliptra-sw-smoke-test.yml#L1
(Line: 13, Col: 5): Unexpected value 'uses'
(Line: 14, Col: 5): Unexpected value 'with'
(Line: 11, Col: 5): Required property is missing: runs-on

Commit 64649d15 ("[ci]: Signal that the smoke test is optional", #1784) added continue-on-error: true to the caliptra-sw-mcu-rom-smoke-test job. That job calls a reusable workflow via uses:, and continue-on-error is not an allowed key on a reusable-workflow-call job. GitHub permits only name, uses, with, secrets, needs, if, strategy, permissions, and concurrency there. Encountering the illegal key, the parser stops treating it as a uses: job and falls back to normal-job validation — which is why it then reports uses/with as unexpected and runs-on as missing.

jobs:
  caliptra-sw-mcu-rom-smoke-test:
    name: Smoke Test (Optional)
    continue-on-error: true        # <-- not valid on a `uses:` job
    uses: chipsalliance/caliptra-sw/.github/workflows/fpga-subsystem.yml@main
    with:
      ...

The file is on main, so this parse error surfaces on all open PRs regardless of their changes.

Suggested fix

Remove the continue-on-error: true line. The "optional / non-blocking" intent from #1784 is better expressed by not marking this check as required in branch-protection settings, rather than by a job-level key (which has no valid equivalent for a reusable-workflow call).

jobs:
  caliptra-sw-mcu-rom-smoke-test:
    name: Smoke Test (Optional)
    uses: chipsalliance/caliptra-sw/.github/workflows/fpga-subsystem.yml@main
    with:
      ...

Example failures:

https://github.com/TimMarkey/caliptra-mcu-sw/actions/runs/30947453197
https://github.com/TimMarkey/caliptra-mcu-sw/actions/runs/30957138595

TimMarkey:
## Problem

The `Caliptra SW Smoke Test (Optional)` workflow fails to parse, so it errors on every PR against `main-2.1`:

```
Invalid workflow file: .github/workflows/caliptra-sw-smoke-test.yml#L1
(Line: 13, Col: 5): Unexpected value 'uses'
(Line: 14, Col: 5): Unexpected value 'with'
(Line: 11, Col: 5): Required property is missing: runs-on
```

## Cause

Commit 64649d15 ("[ci]: Signal that the smoke test is optional", #1784) added `continue-on-error: true` to the `caliptra-sw-mcu-rom-smoke-test` job. That job calls a reusable workflow via `uses:`, and `continue-on-error` is not an allowed key on a reusable-workflow-call job (only `name`, `uses`, `with`, `secrets`, `needs`, `if`, `strategy`, `permissions`, `concurrency` are). The parser hits the illegal key, stops treating it as a `uses:` job, and falls back to normal-job validation, hence the "unexpected `uses`/`with`" and "missing `runs-on`" errors.

## Fix

Remove the `continue-on-error: true` line. The "optional / non-blocking" intent from #1784 is expressed by not marking this check required in branch protection; there is no valid job-level equivalent for a reusable-workflow call.

Fixes #1915

### DIFF
--- a/.github/workflows/caliptra-sw-smoke-test.yml
+++ b/.github/workflows/caliptra-sw-smoke-test.yml
@@ -9,7 +9,6 @@ on:
 jobs:
   caliptra-sw-mcu-rom-smoke-test:
     name: Smoke Test (Optional)
-    continue-on-error: true
     uses: chipsalliance/caliptra-sw/.github/workflows/fpga-subsystem.yml@main
     with:
       mcu-commit: ${{ github.event.pull_request.head.sha || github.sha }}


clundin25 (approved):
Thanks !

TimMarkey:
CI status — two unrelated failures, neither caused by this change:

build-xtask (cancelled): hit the job's 30-minute timeout-minutes limit — it started 22:15:59Z and was killed at 22:46:14Z, ~30m exactly. The entire budget was spent in "Install required packages" (apt-get + rustup toolchain install 1.85 with clippy/rust-src/llvm-tools/rustc-dev across two extra targets); Prebuild xtask never started. This is a cold-runner/uncached-toolchain timeout, not a code issue.

print_test_results (failure): a downstream effect of #1 — with build-xtask cancelled, test_emulator was skipped, so no caliptra-test-results-* artifacts were produced. But print_test_results runs with if: always(), so it ran anyway; the empty /tmp/junit/junit-*.xml glob passed the literal pattern to test-printer, which panicked with Unable to read junit.xml: NotFound.

Both trace back to the transient timeout. A re-run of the failed jobs on a warm runner should clear them. Longer term (separate infra issue, not this PR): Infra fixes for these issues would be to bump build-xtask's timeout or cache the rustup toolchain, and guard print_test_results against empty artifacts.

PR #1927, opened 2026-08-06
spdm: HEARTBEAT / HEARTBEAT_ACK with watchdog

TimMarkey:
## Summary

Adds SPDM **HEARTBEAT / HEARTBEAT_ACK** support to the Caliptra MCU Rust SPDM responder, together with the complete liveness mechanism the command exists for — HeartbeatPeriod advertisement, a per-session watchdog, keep-alive on session traffic, and automatic teardown. Closes the OCP 2.7 SPDM-15 gap (HEARTBEAT_ACK required).

The liveness machinery is behind an **off-by-default** feature `spdm-set-heartbeat`. The shipped default build is unchanged: no `HBEAT_CAP` is advertised and `HeartbeatPeriod` is 0 — the spec-legal "heartbeat not supported" behavior.

## Changes

**Message layer (always compiled):**
- HEARTBEAT (`0xE8`) / HEARTBEAT_ACK (`0x68`) handler: validates version + the 4-byte shape (reserved Param1/Param2 ignored per DSP0274), returns HEARTBEAT_ACK. Phase-gated to an established secure session; an unsecured HEARTBEAT gets `SessionRequired`, not `UnsupportedRequest`.

**Full liveness (feature `spdm-set-heartbeat`, DSP0274 1.3.0 section 10.20):**
- **Conditional period** — KEY_EXCHANGE_RSP advertises a non-zero HeartbeatPeriod (default 3 s) iff both endpoints set `HBEAT_CAP`, else 0. Computed once and folded into both the partial and full KEY_EXCHANGE_RSP builds to keep the transcript hash consistent.
- **Watchdog** — a per-session watchdog is armed at 2× the advertised period on FINISH_RSP transmission, only when the peer set `HBEAT_CAP`.
- **Keep-alive** — any successfully handled secured command (not just HEARTBEAT) restarts the watchdog, so live traffic keeps the session open.
- **Teardown** — on watchdog expiry the run loop clears the session and erases its keys.
- `HBEAT_CAP` is advertised only when the feature is enabled.

## Design notes
- The clock lives on the PAL trait (`SpdmPalIoTransport`) with an inert default (`now_ms → 0`, `sleep_ms` pends forever), so the stack crate's host unit tests need no alarm capsule; the real PAL impl is alarm-capsule-backed (libtock Alarm via `AsyncAlarm`), compiled only under the feature.
- The run loop races `recv_request` against the nearest session deadline using a minimal local two-future `select` (embassy-futures is not vendored), polling `recv_request` first so an arriving request wins a simultaneously-expiring timer.
- The fake alarm capsule now tracks a virtual monotonic clock (command `TIME` reflects accumulated `set_relative` deltas) so tests can observe time move forward.

## Testing
Green with the feature both **on** and **off**:
- 5 HEARTBEAT-handler cases (well-formed ACK, version mismatch, nonzero-reserved accepted, truncated/trailing rejected)
- 7 liveness-watchdog cases (arm, touch/keep-alive, nearest-deadline, expiry, secured keep-alive through the dispatch path, cap advertise/hide)
- 2 codec HeartbeatPeriod-encoding cases

### DIFF
--- a/runtime/userspace/api/spdm-lib/codec/src/key_exchange.rs
+++ b/runtime/userspace/api/spdm-lib/codec/src/key_exchange.rs
@@ -14,6 +14,32 @@ pub const ECDH_P384_EXCHANGE_DATA_SIZE: usize = 96;
 /// Random data length in KEY_EXCHANGE req/rsp.
 pub const KEY_EXCHANGE_RANDOM_DATA_LEN: usize = 32;
 
+/// KEY_EXCHANGE_RSP HeartbeatPeriod (Param1 of the response, DSP0274): a count
+/// of **seconds**. Zero is a distinguished value — heartbeat is supported but
+/// not desired on this session, so no liveness watchdog runs. Wrapping the raw
+/// byte keeps the unit and that semantics in the type instead of passing bare
+/// `u8`s around.
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
+pub struct HeartBeatPeriod(pub u8);
+
+impl HeartBeatPeriod {
+    /// Heartbeat supported but not desired: no watchdog on this session.
+    pub const DISABLED: HeartBeatPeriod = HeartBeatPeriod(0);
+
+    /// Whole seconds carried by this period.
+    #[inline]
+    pub fn secs(self) -> u8 {
+        self.0
+    }
+
+    /// True when a non-zero period was negotiated (liveness watchdog active).
+    #[inline]
+    pub fn is_enabled(self) -> bool {
+        self.0 != 0
+    }
+}
+
 // ---- Request ---------------------------------------------------------------
 
 /// KEY_EXCHANGE request fixed body (after SPDM header).
@@ -61,7 +87,7 @@ impl KeyExchangeReqBody {
 ///   signature(96) | responder_verify_data(0|48) ]
 /// ```
 pub struct KeyExchangeRsp<'a> {
-    pub heartbeat_period: u8,
+    pub heartbeat_period: HeartBeatPeriod,
     pub rsp_session_id: u16,
     pub random_data: &'a [u8; KEY_EXCHANGE_RANDOM_DATA_LEN],
     pub exchange_data: &'a [u8; ECDH_P384_EXCHANGE_DATA_SIZE],
@@ -91,7 +117,7 @@ impl ResponseBody for KeyExchangeRsp<'_> {
 
     fn encode_body(&self, w: &mut WireWriter<'_>) -> Result<(), WireError> {
         // heartbeat_period (0 = no heartbeat on this session)
-        w.write_bytes(&[self.heartbeat_period])?;
+        w.write_bytes(&[self.heartbeat_period.0])?;
         // reserved
         w.write_bytes(&[0u8])?;
         // rsp_session_id (LE)
@@ -153,7 +179,7 @@ mod tests {
         let random = [0u8; KEY_EXCHANGE_RANDOM_DATA_LEN];
         let exchange = [0u8; ECDH_P384_EXCHANGE_DATA_SIZE];
         let body = KeyExchangeRsp {
-            heartbeat_period: period,
+            heartbeat_period: HeartBeatPeriod(period),
             rsp_session_id: 0xABCD,
             random_data: &random,
             exchange_data: &exchange,
--- a/runtime/userspace/api/spdm-lib/codec/src/key_exchange.rs
+++ b/runtime/userspace/api/spdm-lib/codec/src/key_exchange.rs
@@ -61,6 +61,7 @@ impl KeyExchangeReqBody {
 ///   signature(96) | responder_verify_data(0|48) ]
 /// ```
 pub struct KeyExchangeRsp<'a> {
+    pub heartbeat_period: u8,
     pub rsp_session_id: u16,
     pub random_data: &'a [u8; KEY_EXCHANGE_RANDOM_DATA_LEN],
     pub exchange_data: &'a [u8; ECDH_P384_EXCHANGE_DATA_SIZE],
@@ -89,8 +90,8 @@ impl ResponseBody for KeyExchangeRsp<'_> {
     }
 
     fn encode_body(&self, w: &mut WireWriter<'_>) -> Result<(), WireError> {
-        // heartbeat_period = 0 (no heartbeat)
-        w.write_bytes(&[0u8])?;
+        // heartbeat_period (0 = no heartbeat on this session)
+        w.write_bytes(&[self.heartbeat_period])?;
         // reserved
         w.write_bytes(&[0u8])?;
         // rsp_session_id (LE)
@@ -140,5 +141,48 @@ impl KeyExchangeRsp<'_> {
         } else {
             0
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{ResponseBody, SpdmMsgHdrPdu, SpdmVersion, WireWriter};
+
+    fn encode(period: u8) -> ([u8; 256], usize) {
+        let random = [0u8; KEY_EXCHANGE_RANDOM_DATA_LEN];
+        let exchange = [0u8; ECDH_P384_EXCHANGE_DATA_SIZE];
+        let body = KeyExchangeRsp {
+            heartbeat_period: period,
+            rsp_session_id: 0xABCD,
+            random_data: &random,
+            exchange_data: &exchange,
+            meas_summary_hash: None,
+            opaque_data: &[],
+            signature: &[],
+            responder_verify_data: None,
+        };
+        let mut buf = [0u8; 256];
+        let mut w = WireWriter::new(&mut buf);
+        body.encode_with_header(SpdmVersion::V13, &mut w).unwrap();
+        let len = w.position();
+        (buf, len)
+    }
+
+    // The HeartbeatPeriod is the first body byte (immediately after the 2-byte
+    // common header) and must carry exactly the negotiated value.
+    #[test]
+    fn heartbeat_period_is_first_body_byte() {
+        let (buf, _) = encode(3);
+        assert_eq!(buf[SpdmMsgHdrPdu::SIZE], 3);
+        // reserved byte stays zero.
+        assert_eq!(buf[SpdmMsgHdrPdu::SIZE + 1], 0);
+    }
+
+    // A zero period (heartbeat not desired / not supported) encodes as zero.
+    #[test]
+    fn heartbeat_period_zero_encodes_zero() {
+        let (buf, _) = encode(0);
+        assert_eq!(buf[SpdmMsgHdrPdu::SIZE], 0);
     }
 }
--- a/runtime/userspace/api/spdm-lib/pal/Cargo.toml
+++ b/runtime/userspace/api/spdm-lib/pal/Cargo.toml
@@ -18,7 +18,6 @@ caliptra-mcu-libsyscall-caliptra = { workspace = true, optional = true }
 caliptra-mcu-libtock_platform = { workspace = true, optional = true }
 caliptra-mcu-libtock_alarm = { workspace = true, optional = true }
 caliptra-mcu-libtockasync = { workspace = true, optional = true }
-embassy-sync = { workspace = true, optional = true }
 zerocopy = { workspace = true }
 
 [features]
@@ -36,5 +35,4 @@ spdm-set-heartbeat = [
   "caliptra-mcu-libtock_platform",
   "caliptra-mcu-libtock_alarm",
   "caliptra-mcu-libtockasync",
-  "embassy-sync",
 ]
--- a/runtime/userspace/api/spdm-lib/pal/Cargo.toml
+++ b/runtime/userspace/api/spdm-lib/pal/Cargo.toml
@@ -15,6 +15,9 @@ mcu-caliptra-api-lite = { workspace = true, features = ["mailbox-io"] }
 mcu-error = { workspace = true }
 caliptra-mcu-libsyscall-caliptra = { workspace = true, optional = true }
 caliptra-mcu-libtock_platform = { workspace = true, optional = true }
+caliptra-mcu-libtock_alarm = { workspace = true, optional = true }
+caliptra-mcu-libtockasync = { workspace = true, optional = true }
+embassy-sync = { workspace = true, optional = true }
 zerocopy = { workspace = true }
 
 [features]
@@ -24,4 +27,13 @@ set-certificate = [
   "caliptra-mcu-spdm-traits/set-certificate",
   "caliptra-mcu-libsyscall-caliptra",
   "caliptra-mcu-libtock_platform",
+]
+# HEARTBEAT liveness watchdog: pulls in the libtock alarm capsule so the PAL
+# can provide a real monotonic clock (now_ms/sleep_ms) to the stack.
+spdm-set-heartbeat = [
+  "caliptra-mcu-libsyscall-caliptra",
+  "caliptra-mcu-libtock_platform",
+  "caliptra-mcu-libtock_alarm",
+  "caliptra-mcu-libtockasync",
+  "embassy-sync",
 ]
--- a/runtime/userspace/api/spdm-lib/pal/Cargo.toml
+++ b/runtime/userspace/api/spdm-lib/pal/Cargo.toml
@@ -16,6 +16,9 @@ caliptra-mcu-scratch-alloc = { workspace = true }
 mcu-error = { workspace = true }
 caliptra-mcu-libsyscall-caliptra = { workspace = true, optional = true }
 caliptra-mcu-libtock_platform = { workspace = true, optional = true }
+caliptra-mcu-libtock_alarm = { workspace = true, optional = true }
+caliptra-mcu-libtockasync = { workspace = true, optional = true }
+embassy-sync = { workspace = true, optional = true }
 zerocopy = { workspace = true }
 
 [features]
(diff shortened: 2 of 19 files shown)


TimMarkey:
Please approve the run for this fork, thanks. cc @clundin25 @swenson @zhalvorsen @parvathib

TimMarkey:
@clundin25 @parvathib @RaunakGu — Please review this. Thanks!

clundin25:
@TimMarkey it looks like this was pushed during a github outage. Do you mind pushing the branch to re-trigger the CI?

TimMarkey:
@clundin25 — CI ran after the re-push. Two jobs failed, and I reviewed both logs; neither failures were caused by this change. Would you mind re-running the two failed jobs when you get a chance? I don't have Actions re-run permission on the upstream repo.

analyze (size analysis) — this isn't a size regression; the job never got to a size verdict. The size-history walker aborted while checking out a historical baseline commit (5796651) because a build had dirtied Cargo.lock, and its git checkout has no stash/force:

error: Your local changes to the following files would be overwritten by checkout:
	Cargo.lock
fatal: working trees containing submodules cannot be moved or removed
Error: Git error: ... ["checkout","--no-recurse-submodule","--quiet","5796651..."] exited with status code Some(1)
test_artifacts — a single unrelated timeout: test_pldm_fw_update::test::test_fw_update_e2e ran over 60s on the hw-model-fpga and timed out (test timed out). It's a PLDM firmware-update e2e test, unrelated to the SPDM heartbeat change here.

Both look transient/infra, so a re-run should clear them. If analyze fails again at that same Cargo.lock checkout, it looks like a robustness issue in the size-history step (it should stash or force the checkout) rather than anything in this PR -Thanks!

RaunakGu:
@TimMarkey  Tim, I had pushed fix for size_history issue, please rebase

TimMarkey:
@clundin25 @parvathib — Please review this.  All tests have passed. 
Thanks!

clundin25:
@TimMarkey I will take a look but I am not familiar with the SPDM code so I'll defer to @parvathib for the final sign off

runtime/userspace/api/spdm-lib/codec/src/key_exchange.rs
```diff
@@ -61,6 +61,7 @@ impl KeyExchangeReqBody {
 ///   signature(96) | responder_verify_data(0|48) ]
 /// ```
 pub struct KeyExchangeRsp<'a> {
+    pub heartbeat_period: u8,
```
clundin25: What is the unit of time for this period? Should the `u8` be wrapped by a new type to make that clear?

runtime/userspace/api/spdm-lib/pal/src/clock.rs
```diff
@@ -0,0 +1,67 @@
...
+fn get_ticks() -> Result<u32, ErrorCode> {
+    DefaultSyscalls::command(DRIVER_NUM, command::TIME, 0, 0).to_result()
+}
+
+/// Current value of the free-running monotonic millisecond clock. Returns 0 if
+/// the alarm capsule is unavailable, which keeps the watchdog inert rather than
+/// tearing sessions down spuriously.
+pub(crate) fn now_ms() -> u64 {
+    let (Ok(ticks), Ok(freq)) = (get_ticks(), get_frequency()) else {
+        return 0;
```
clundin25, line 40: Would an `Option` or `Result` be more idiomatic if the alarm capsule is having issues?

runtime/userspace/api/spdm-lib/pal/src/clock.rs
```diff
@@ -0,0 +1,67 @@
...
+    };
+    if freq == 0 {
+        return 0;
+    }
+    (ticks as u64).saturating_div((freq as u64).saturating_div(1000).max(1))
+}
+
+/// Sleep for at least `ms` milliseconds using the alarm capsule.
+pub(crate) async fn sleep_ms(ms: u64) {
```
clundin25: Why not just take a u32 if it's being truncated?

clundin25: Also, why does this API discard errors?

runtime/userspace/api/spdm-lib/pal/src/io.rs
```diff
@@ -204,4 +204,20 @@ impl<M: MeasurementProvider> SpdmPalIoTransport for McuSpdmPal<M> {
         let transport = unsafe { self.transport_mut() };
         transport.send_response(kind, msg).await
     }
+
+    /// Monotonic millisecond clock backed by the libtock alarm capsule,
+    /// used by the HEARTBEAT liveness watchdog. Returns 0 when the alarm
+    /// capsule is unavailable, which keeps the watchdog inert.
+    #[cfg(feature = "spdm-set-heartbeat")]
+    fn now_ms(&self) -> u64 {
```
clundin25: nit: You could encode the the unit with a type?

```rust
fn now(&self) -> Milliseconds
```

runtime/userspace/api/spdm-lib/stack/src/key_exchange.rs
```diff
@@ -361,6 +367,28 @@ async fn key_exchange_inner<'a, Pal: SpdmPal, const N: usize>(
...
+        {
+            return crate::heartbeat::DEFAULT_HEARTBEAT_PERIOD_SECS;
+        }
+        0
+    }
+    #[cfg(not(feature = "spdm-set-heartbeat"))]
+    {
+        let _ = state;
+        0
+    }
```
clundin25: Does 

```Rust
if cfg!(feature = "spdm-set-heartbeat") {
 . . .
} else {
        let _ = state;
        0
}
```
Work for this? IIRC there was some issues with `cfg!` in this repo?

runtime/userspace/api/spdm-lib/stack/src/stack.rs
```diff
@@ -301,6 +302,20 @@ fn set_certificate_other_params() -> OtherParamSupport {
...
 }
 
+// Advertise HBEAT_CAP only when the heartbeat liveness watchdog is compiled in.
+// With the feature off, no HBEAT_CAP is advertised, so per DSP0274 the
+// HeartbeatPeriod is always zero and no liveness is expected (spec-legal
+// "heartbeat not supported").
+#[cfg(feature = "spdm-set-heartbeat")]
+fn heartbeat_cap_flags() -> CapFlags {
+    CapFlags::HBEAT
+}
```
clundin25: Similarly this could be folded into one function if `cfg!()` works

runtime/userspace/api/spdm-lib/stack/src/stack.rs
```diff
@@ -378,6 +393,33 @@ impl<Pal: SpdmPal, const MAX_SESSIONS: usize, Vdm: SpdmVdmBackend>
...
+                                    continue;
+                                }
+                            }
+                        }
+                        None => break self.pal.recv_request().await?,
+                    }
+                }
+            };
+            #[cfg(not(feature = "spdm-set-heartbeat"))]
             let io = self.pal.recv_request().await?;
```
clundin25: Does it make sense to wrap this in a function now? This will quickly become complex as more feature flags modify the loops behavior

runtime/userspace/api/spdm-lib/pal/Cargo.toml
```diff
@@ -26,3 +29,12 @@ set-certificate = [
...
 ]
+# HEARTBEAT liveness watchdog: pulls in the libtock alarm capsule so the PAL
+# can provide a real monotonic clock (now_ms/sleep_ms) to the stack.
+spdm-set-heartbeat = [
+  "caliptra-mcu-libsyscall-caliptra",
+  "caliptra-mcu-libtock_platform",
+  "caliptra-mcu-libtock_alarm",
+  "caliptra-mcu-libtockasync",
+  "embassy-sync",
+]
```
clundin25, line 38: Why does this need to be feature gated?

runtime/userspace/api/spdm-lib/codec/src/key_exchange.rs
TimMarkey: This is the HeatbeatPeriod field which is Param1 of KEY_EXCHANGE_RSP from DSP0274, so it is a byte in length.  If you want, I can add a comment in the code.

runtime/userspace/api/spdm-lib/pal/src/clock.rs
TimMarkey: The 0 return is intentional — it's a "fail safe/inert" design rather than dropped error handling.

now_ms() is a monotonic clock that the heartbeat watchdog uses only to measure elapsed time (deadline = start + timeout). If the alarm capsule is unavailable and we return 0, the clock simply never advances, so no deadline is ever reached and the watchdog just stays dormant. A broken clock therefore degrades to "no liveness watchdog," which is the pre-heartbeat behavior — it never tears down a healthy session by mistake.

If I returned Option/Result instead, every caller would have to handle the error, and the only sensible thing each one could do is treat "no clock" as "don't advance / don't time out" — i.e. exactly what returning 0 already does. So the sentinel keeps that fallback in one place instead of repeating it at each call site.

The one issue is that 0 is also a legitimate clock reading at t=0, but since the value is only ever used for relative elapsed-time math, that's harmless. I can add a comment making the fail-inert intent explicit if that helps.

runtime/userspace/api/spdm-lib/pal/src/clock.rs
TimMarkey: ms is u64 because this implements the SpdmPalIoTransport::sleep_ms(&self, ms: u64) trait method, and the stack hands it a u64 deadline delta. The alarm capsule below only takes Milliseconds(u32), so the narrowing has to happen at this boundary — taking u32 here wouldn't remove it, just move it up into the trait impl.

Also, line 56 caps to the max instead of wrapping around, (u32::try_from(ms).unwrap_or(u32::MAX)), not a truncation — anything past u32::MAX ms (~49 days) pins to the max rather than wrapping. Heartbeat delays are on the order of seconds, so it never actually clamps in practice.

I can rename this function to ms_u32 / add a short comment so it's clear that it clamps oversized delays instead of letting them wrap.   I'd keep the u64 signature to match the trait.

runtime/userspace/api/spdm-lib/pal/src/io.rs
TimMarkey: I understand why we should encode the unit in the type. The snag is that now_ms here is implementing the SpdmPalIoTransport trait method, and that trait lives in the spdm-traits crate, which doesn't depend on libtock — so it can't name Milliseconds without pulling the libtock alarm dependency down into the abstract PAL trait (and everything built on it, including the stack).

There's also a width mismatch: Milliseconds wraps a u32, but the watchdog deadlines in the stack are done in u64 ms, so we'd either lose range or end up converting anyway.

So I'd lean toward keeping u64 ms and just making the unit clear in the name and doc comment.

runtime/userspace/api/spdm-lib/stack/src/key_exchange.rs
TimMarkey: ok, I folded it into one function with cfg!(), same as the earlier heartbeat_cap_flags change.

runtime/userspace/api/spdm-lib/stack/src/stack.rs
TimMarkey: ok, this one works cleanly with cfg!(). Both arms only return plain CapFlags constants (HBEAT / EMPTY) that always exist regardless of the feature, so there's no gated item to trip over. I'll fold it into a single function:


fn heartbeat_cap_flags() -> CapFlags {
    if cfg!(feature = "spdm-set-heartbeat") {
        CapFlags::HBEAT
    } else {
        CapFlags::EMPTY
    }
}

(Note that this is why it works here but not on heartbeat_period_for above — that one's feature-on branch returns the gated DEFAULT_HEARTBEAT_PERIOD_SECS, so cfg!() won't compile there without ungating it.)

runtime/userspace/api/spdm-lib/pal/Cargo.toml
TimMarkey: Gating it keeps the libtock alarm capsule (and libtockasync) out of builds that don't run the liveness watchdog — those deps are only used by clock.rs. It's also the pal half of the cross-crate spdm-set-heartbeat opt-in: the stack only advertises HBEAT_CAP and runs the watchdog under the same feature, so heartbeat stays off-by-default (spec-legal "heartbeat not supported"). Correctness doesn't require the gate — now_ms() is fail-inert if the capsule is absent.  Again, it's purely to avoid forcing the alarm dependency on every platform.

runtime/userspace/api/spdm-lib/codec/src/key_exchange.rs
clundin25: I think it makes sense to have a new type to encode some of the semantics of this into the type system and to avoid mixing it with random u8s

```rust
#[repr(transparent)]
struct HeartBeatPeriod(u8);
```

I would call this optional feedback because I am not familiar with the rest of the SPDM code, but we should as a rule of thumb try to use the type system to encode these special semantics.

E.g. 0 means heartbeat is supported but not desired, the unit of time is seconds

runtime/userspace/api/spdm-lib/pal/src/clock.rs
clundin25: If we go from heartbeat -> never ending heartbeat due to a broken clock, does that result in a system that now has a leak? 

How will the other side of the SPDM connection respond? When will we know to close it?

Would it be better to close the connection?

runtime/userspace/api/spdm-lib/pal/src/clock.rs
clundin25: Right, moving it up the boundary requires the caller to perform the conversion

runtime/userspace/api/spdm-lib/pal/src/io.rs
clundin25: Seems like this crate could define these types then?

runtime/userspace/api/spdm-lib/pal/Cargo.toml
clundin25: I see, dumb question, does it make sense to invert the feature requirement? e.g. a `clock` feature that makes the heartbeat available? Or is the desire to always be able to turn it off at compile time?

Will there be other SPDM commands / features that need clock access?

runtime/userspace/api/spdm-lib/stack/src/stack.rs
TimMarkey: OK, I extracted into a recv_next free function. It takes pal and sessions separately rather than being a &mut self method, because io borrows from pal and a &mut self method would tie up the rest of self while io is in use, blocking dispatch from touching sessions/state later in the loop.

runtime/userspace/api/spdm-lib/pal/Cargo.toml
TimMarkey: Right now heartbeat is the only thing that touches the clock — the watchdog run loop, arming the deadline on FINISH_RSP, and refreshing it on session traffic. No other command reads time.

So yes, the intent is to keep it something you can compile out entirely: with the feature off there's no HBEAT_CAP, no clock, and no alarm-capsule dependency.  Heartbeat not supported is still the default.

Inverting it into a clock feature that heartbeat turns on is a nice layering, but today that feature would have exactly one user, so it'd be adding structure ahead of need. If a second clock consumer is added in the future, then splitting clock out and making heartbeat depend on it is a small mechanical change. I'd prefer to wait until it is needed, but if you feel strongly about updating it now to prepare for a future use I can.

runtime/userspace/api/spdm-lib/pal/src/io.rs
TimMarkey: ok, I added a Milliseconds(u64) type to the traits crate and switched the clock methods to now() -> Milliseconds / sleep(dur: Milliseconds). The deadline arithmetic stays u64 internally, so the newtype just pins the unit at the trait boundary.

runtime/userspace/api/spdm-lib/codec/src/key_exchange.rs
TimMarkey: I added HeartBeatPeriod(u8) in the codec crate with a DISABLED constant and secs()/is_enabled(). It's now used both for the byte we send in KEY_EXCHANGE_RSP and for the period we store on the session. Unit (seconds) and the zero-means-not-desired meaning are documented on the type.

runtime/userspace/api/spdm-lib/pal/src/clock.rs
TimMarkey: Good point, HBEAT_CAP only obligates us to answer a HEARTBEAT with a HEARTBEAT_ACK, which we always do regardless of the clock. The part a bad timer actually defeats is the separate 10.21 rule that the responder shall tear the session down after two missed periods.

The spec also lets HeartbeatPeriod be 0 even when Heartbeat is supported ("a heartbeat is not desired on this session"), so there's a clean way to keep the capability but not promise a watchdog we can't back.

The catch here is that our clock failure is transient — the alarm syscall can fail on any call, and we have no signal at GET_CAPABILITIES time that it'll later go bad. So today the fail-inert path (clock returns 0, we just don't tear down) is the fallback for that runtime case. If we want to close the gap properly, the cleaner move would be a one-time clock probe at startup and advertise HeartbeatPeriod=0 when it fails, rather than handling it at runtime.

I'd like more feedback on whether that's worth doing, or whether fail-inert is an acceptable choice for a fault the spec doesn't cover.

clundin25: Your proposal looks good to me.

clundin25:
Thanks @TimMarkey I will try to take another look tonight or tomorrow morning

runtime/userspace/api/spdm-lib/pal/src/clock.rs
TimMarkey: I updated the code so that sleep_ms now takes a u32. The value is always a sleep length (deadline - now), which never gets big enough to need a u64, so u32 is always enough

clundin25 (approved):

PR #1928, opened 2026-08-06
fix(fpga): mctp test after hitless update reset

mlvisaya:
The integration test to exercise MCTP communication after a hitless update reset is flaky. Currently, it uses the ROM to perform directly hitless update reset, however, this is causing sometimes i3c to not initialize properly after reset.

Use the actual hitless update reset path in Runtime using the activation path in firmware update. This will cause MCU Runtime to follow the proper reset sequence to boot to the new image before doing the MCTP loopback test.


helloxiling (approved):
Looks good to me.

PR #1930, opened 2026-08-06
[feat] DOT: implement ownership transfer commands over MCI and SPDM VDM

mhatrevi:
## Summary

- Add the complete Runtime Device Ownership Transfer command family over MCI mailbox and SPDM VDM.
- Route both transports through shared Runtime backends that seal and verify DOT blobs, commit and verify OTP transitions, and enforce command-specific authorization.
- Add typed host APIs, SPDM client validation, platform DOT blob storage, capability reporting, and protocol documentation.

## Protocol

All commands use DOT family ID `0x00000011` and a four-byte DOT subcommand:

- Generic-authorized: `MDLK` (lock), `MDDS` (disable), `MDRT` (rotate), `MDOT` (backup blob)
- Native: `MDUC` (unlock challenge), `MDUL` (unlock), `MDST` (status), `MDRC` (recovery), `DOTW` (override challenge), `DOTX` (override)

SPDM carries protected commands as `AuthorizedCommand (0x12) -> family 0x11`; MCI sends them through its authorization gateway. Both transports verify:

```text
0x00000011(BE) || DOT_FourCC(LE) || exact payload || MACC nonce
```

Direct SPDM `0x11` requests for protected commands return `AccessDenied`.

## Runtime behavior

- LOCK and DISABLE seal the target ODD-state blob before burning one fuse bit.
- UNLOCK authenticates the current LAK challenge response, preserves the authenticated LAK hash in an empty-CAK blob, then burns one fuse bit.
- ROTATE burns two bits and seals with the effective key for the resulting parity and epoch.
- Backup and recovery authenticate the exact 168-byte blob with the current device-derived key.
- Override verifies the fused DOT recovery-key hash and hybrid challenge signatures.
- Blob writes and OTP burns are read-back verified before success is returned.
- Zero keys and invalid state transitions remain rejected.

Firmware-manifest DOT processing and ROM recovery behavior are unchanged. Restricting Runtime recovery and override commands to a ROM-signaled recovery boot is deferred; their native cryptographic and state checks remain mandatory.

## Validation

- `cargo test -p caliptra-mcu-spdm-vdm-handler --features device-ownership-transfer` (64 passed)
- `cargo test -p caliptra-mcu-tests-integration test_runtime_dot_ -- --nocapture` (11 passed)
- Portable host command types and transport tests (3 + 23 passed)
- Command authorization signer tests (7 passed)
- Runtime app check with `dot-mci-mailbox,dot-spdm-vdm`
- Main and host workspace rustfmt checks
- `cargo xtask docs`
- `git diff --check`


docs/src/caliptra_spdm_vdm_cmds.md
```diff
@@ -101,6 +102,7 @@ The following subcommands are assigned to the SPDM VDM IANA authorization-gated
 | `0x4D52_564B` (`MRVK`) | FuseRevokeVendorPubKey     | Supported     | Revoke vendor public key.                           |
 | `0x5256_4B48` (`RVKH`) | FuseRevokeVendorPkHash     | Supported     | Revoke vendor public key hash.                      |
 | `0x4946_504B` (`IFPK`) | FuseLockPartition          | Planned (TBD) | Lock fuse partition.                                |
+| `0x0000_0011`          | DeviceOwnershipTransfer    | Supported     | Carry authorization-gated DOT subcommands.          |
```
mlvisaya, line 106: nit: not a blocker. probably nicer to assign this as 'MDOT' 0x4D44_4F54 to align with other commands

runtime/kernel/capsules/src/memory_flash.rs
```diff
@@ -0,0 +1,147 @@
...
+        self.deferred_call.set();
+    }
+}
+
+impl<'a> FlashStorage<'a> for MemoryFlash<'a> {
+    fn set_client(&self, client: &'a dyn FlashStorageClient) {
+        self.client.set(client);
+    }
+
+    fn read(
```
mlvisaya, line 63: i see that all the operations here are deferred, and the result returned is just whether the operation has been queued, but the caller actually doesn't know if the operation really succeeded because it's done in the deferred action. 

should we add a wait_for_result() method so that caller can call would know the result of the operation? The result can be set by the deferred action.

platforms/emulator/runtime/src/board.rs
```diff
@@ -520,6 +533,20 @@ pub unsafe fn main() {
         execute: false,
     });
 
+    // The emulator root bus owns this dedicated region across MCU resets, and
```
mlvisaya, line 536: nit: do we also need to enable in fpga board.rs? maybe we can add in fpga but not add a test for it since the memoryflash peripheral probably doesn't exist in FPGA? currently the size-history analyzer checks the size of the FPGA firmware with all-features in it, so we don't include it in FPGA, the size-history will not include this change.

docs/src/caliptra_spdm_vdm_cmds.md
mhatrevi: Yeah, MDOT would be nicer, but SPDM already uses 0x11 as the DOT family ID. I kept 0x11 consistent across MCI and SPDM, and renamed Get Backup Blob from MDOT to MDBB to avoid the ambiguity.

runtime/kernel/capsules/src/memory_flash.rs
mhatrevi: The actual RAM read/write/erase happens before the function returns [Ok](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html); only the completion callback is deferred so it follows the async flash HIL and avoids re-entrancy issues. Any busy, size, or range error is returned immediately, and there isn’t another fallible operation in the deferred handler. The caller already waits for the normal [read_done](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[write_done](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[erase_done](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) callback, so I don’t think we need a separate wait_for_result(). I can add a comment to make that clearer.

platforms/emulator/runtime/src/board.rs
mhatrevi: Yeah, the same physical secondary flash can be used, but FPGA Runtime doesn’t have an async driver for that controller yet. Its current flash path is mailbox-backed imaginary flash, while ROM reads the physical secondary flash directly. Enabling DOT in FPGA [board.rs](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) now would expose commands without a working shared blob store. Also, I checked size-history: it builds FPGA release with [--no-default-features](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), so adding DOT to FPGA [all-features](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) wouldn’t include it there anyway. I documented the current emulator support and what FPGA needs before DOT can be enabled.

caliptra-util-host/apps/spdm/test-config.toml
```diff
@@ -72,3 +72,9 @@ key_index = 0
 enabled = false
 reserved = 0
 vendor_pk_hash_slot = 1
+
+[dot]
+enabled = false
```
mlvisaya: will this be tested by util host validator?

caliptra-util-host/apps/spdm/client/src/config.rs
```diff
@@ -262,3 +264,15 @@ pub struct RevokeVendorPkHashConfig {
 fn default_true() -> bool {
     true
 }
+
+#[derive(Debug, Clone, Default, Deserialize)]
```
mlvisaya: nit: not a blocker, removing Debug actually helps reduce size

mlvisaya (approved):
I think we're missing the MCU mailbox implementation in the caliptra-util-host side. For MCU mailbox commands, there should be additions in:

caliptra-util-host/apps/mailbox
caliptra-util-host/transport/src/transports/mailbox

It can be added in a future follow-up.

mhatrevi:
> I think we're missing the MCU mailbox implementation in the caliptra-util-host side. For MCU mailbox commands, there should be additions in:
> 
> caliptra-util-host/apps/mailbox caliptra-util-host/transport/src/transports/mailbox
> 
> It can be added in a future follow-up.

This is now added.

PR #1933, opened 2026-08-07
spdm: install the ML-DSA-87 IDevID cert into Caliptra

RaunakGu:
## What's fixed

**The ML-DSA-87 IDevID certificate was never installed into Caliptra.** The MCU read the ECC-384 cert from OTP partition 1 and installed it, but ignored the ML-DSA-87 cert already provisioned in partition 2. Caliptra never issues the IDevID cert, so only the MCU can install it — without this the PQC chain has no path to the vendor root.

## How

- Staged, not streamed — the cert is read into a 9 KiB heap scratch pool and sent as one payload. Streaming would hold the Caliptra mailbox with EXECUTE asserted across ~1,900 sequential 4-byte OTP syscalls and starve other mailbox users. The pool is heap-allocated and freed when the installer task ends, so it is not held for the life of the app.
- Length taken from the cert's own DER header, bounded by both the OTP partition size and Caliptra's 8192-byte cap, so a short cert can't splice `0xFF` fill into the chain.
- Best-effort: cannot fail cert-store init, since no SPDM requester can negotiate ML-DSA yet and a PQC defect must not break the ECC chain. Both failure paths log a warning.

## Scope

Write path only — nothing on the MCU can read the ML-DSA chain back today (its sole reader is behind `INVOKE_DPE_MLDSA87`; the MCU hardcodes `INVOKE_DPE_ECC384`). Landing it now because OTP is burned at manufacturing.

## Verification

3 new unit tests in `mcu-caliptra-api-lite` (32 total pass) covering DER length accept/reject, length guards, and checksum folding. Clean across all 8 feature configs the workflows build. **SRAM delta: 0** — the 8192-byte FPGA overflow is identical on base and this branch.

> **CI note:** the `analyze` failure is unrelated to this change — the size-history walker aborts on a dirty `Cargo.lock` left by main's commit `4ea26b146`, reproducible on plain `main` with none of this code. Fix upstream in chipsalliance/caliptra-infra#76.

### DIFF
--- a/platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
@@ -2,11 +2,17 @@
 
 //! Cert store initialization for the spdm-lib emulator platform.
 //!
-//! Reads the IDevID ECC-384 certificate from OTP, installs it into
-//! Caliptra, and configures the cert slots:
+//! Reads the IDevID certificates from OTP, installs them into Caliptra, and
+//! configures the cert slots:
 //!   - Slot 0 (Vendor): ReadOnly endorsement from static Root CA
 //!   - Slot 1 (Owner):  Managed endorsement (flash-backed, initially empty)
 //!   - Slot 2 (Tenant): Managed endorsement (flash-backed, initially empty)
+//!
+//! Caliptra generates the IDevID *keypair* and a self-signed CSR, but never the
+//! IDevID *certificate* — that is issued externally and provisioned into OTP.
+//! Both the ECC-384 (partition 1) and ML-DSA-87 (partition 2) IDevID certs are
+//! read here and prepended to their respective Caliptra cert chains via the
+//! `POPULATE_IDEV_*_CERT` mailbox commands.
 
 mod slot0_endorsements;
 
@@ -14,8 +20,16 @@ mod slot0_endorsements;
 use caliptra_mcu_config_emulator::flash::CERT_STORE_PARTITION;
 use caliptra_mcu_libsyscall_caliptra::external_otp::ExternalOtp;
 use caliptra_mcu_libsyscall_caliptra::DefaultSyscalls;
+use caliptra_mcu_libtock_console::Console;
 use caliptra_mcu_spdm_pal::cert::store::SharedCertStore;
-use mcu_caliptra_api_lite::{populate_idev_ecc384_cert, ApiAlloc};
+// `log_warn!` writes through the console writer, so the trait must be in scope;
+// but the macro compiles to nothing in configurations without a log transport,
+// which makes the import look unused. Same treatment as spdm/mod.rs.
+#[allow(unused_imports)]
+use core::fmt::Write as _;
+use mcu_caliptra_api_lite::{
+    mldsa87_cert_der_len, populate_idev_ecc384_cert, populate_idev_mldsa87_cert, ApiAlloc,
+};
 use mcu_error::McuResult;
 
 /// SPDM slot IDs for OCP PKI entities.
@@ -31,23 +45,55 @@ const ECC_DEVID_CERT_SIZE: usize = 547;
 /// OTP partition ID for the IDevID ECC certificate.
 const OTP_IDEVID_ECC_PARTITION: u32 = 0x01;
 
+/// OTP partition ID for the IDevID ML-DSA-87 certificate.
+const OTP_IDEVID_MLDSA_PARTITION: u32 = 0x02;
+
 #[cfg(feature = "test-mctp-spdm-set-certificate")]
 const MANAGED_SLOT_COUNT: usize = 2;
 #[cfg(feature = "test-mctp-spdm-set-certificate")]
 const MANAGED_SLOT_REGION_SIZE: usize = CERT_STORE_PARTITION.size / MANAGED_SLOT_COUNT;
 
-/// Initialize the cert store for all 3 slots.
+/// One-time Caliptra setup: read the IDevID certs from OTP and install them.
 ///
-/// Must be called once before SPDM responder tasks start.
-/// Slot 0 failure is fatal. Slots 1-2 stay unprovisioned if
-/// flash is empty (they'll be provisioned via SET_CERTIFICATE).
-/// One-time Caliptra setup: read IDevID from OTP and install.
+/// The ECC-384 install is required — its failure aborts cert-store init because
+/// every SPDM flow in use today signs with ECC. The ML-DSA-87 install is
+/// best-effort and *cannot* fail this function: no SPDM requester can negotiate
+/// ML-DSA today (`AsymAlgos` has no PQC codepoint), so a PQC-provisioning defect
+/// must not take down the ECC chain that attestation actually depends on.
 pub async fn populate_idev<A: ApiAlloc>(alloc: &A) -> McuResult<()> {
     populate_idev_from_otp(alloc).await
 }
 
-/// Configure endorsement chains on the shared cert store.
-/// Called once from spdm_task before spawning responders.
+/// Install the IDevID ML-DSA-87 cert, off the responder-startup path.
+///
+/// Deliberately *not* part of [`populate_idev`]. The OTP driver reads 4 bytes
+/// per syscall, so a 7741-byte certificate costs 1,936 sequential reads against
+/// the ECC certificate's 137. Doing that before the responders start delays
+/// them past the point where a requester that is already connected gives up
+/// ("Error sending SPDM request"). Nothing can read the ML-DSA chain yet, so the
+/// install has no reason to gate SPDM coming up.
+///
+/// Failures are logged and swallowed: no SPDM requester can negotiate ML-DSA
+/// today (`AsymAlgos` has no PQC codepoint), so a PQC-provisioning defect must
+/// not disturb the ECC chain that attestation actually depends on. An
+/// unprovisioned or unreadable PQC IDevID on a device that should have one is a
+/// provisioning escape, so it is reported rather than dropped silently.
+pub async fn populate_idev_mldsa<A: ApiAlloc>(alloc: &A) {
+    if let Err(e) = populate_idev_mldsa_from_otp(alloc).await {
+        let mut cw = Console::<DefaultSyscalls>::writer();
+        crate::log_warn!(
+            cw,
+            "SPDM: ML-DSA-87 IDevID cert not installed: 0x{}",
+            crate::Hex32(u32::from(e))
+        );
+    }
+}
+
+/// Configure endorsement chains on the shared cert store, for all 3 slots.
+///
+/// Called once from spdm_task before spawning responders. Slot 0 failure is
+/// fatal. Slots 1-2 stay unprovisioned if flash is empty (they'll be
+/// provisioned via SET_CERTIFICATE).
 pub async fn setup_endorsements<A: ApiAlloc>(store: &SharedCertStore, alloc: &A) -> McuResult<()> {
     // Slot 0 (Vendor): ReadOnly endorsement with static Root CA.
     // Retry on mailbox busy (SHA calls during root cert hashing).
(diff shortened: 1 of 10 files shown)


platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
```diff
@@ -145,3 +191,115 @@ async fn populate_idev_from_otp<A: ApiAlloc>(alloc: &A) -> McuResult<()> {
...
+async fn populate_idev_mldsa_from_otp<A: ApiAlloc>(alloc: &A) -> McuResult<()> {
+    let otp = ExternalOtp::<DefaultSyscalls>::new();
+
+    // Submit the cert's own DER length, not the whole partition: a production
+    // cert shorter than its partition would otherwise splice the 0xFF fill into
+    // the chain.
+    let Some(cert_size) = mldsa_cert_der_len(&otp).await? else {
+        // No cert provisioned — leave the MLDSA chain as Caliptra built it. Say
+        // so: this is the branch that silently decides a device ships without a
+        // PQC IDevID, so it must not be indistinguishable from a successful
```
zhalvorsen: nit: can you remove this double negative in the comment?

runtime/userspace/api/caliptra-api-lite/src/cert.rs
```diff
@@ -56,6 +56,92 @@ pub async fn populate_idev_ecc384_cert<A: ApiAlloc>(alloc: &A, cert: &[u8]) -> M
...
+
+/// Build the `chksum(4) | cert_size(4)` request header for
+/// `POPULATE_IDEV_MLDSA87_CERT`.
+///
+/// The mailbox checksum covers the certificate bytes as well as the header.
+/// Because [`calc_checksum`] already returns the *negated* running sum, folding
+/// in the payload's byte-sum is a single subtraction:
+/// `0 - (S_hdr + S_cert) == calc_checksum(cmd, hdr) - S_cert`.
+fn populate_idev_mldsa87_header(cert_size: u32, cert_bytesum: u32) -> [u8; PREFIX_LEN] {
+    let mut header = [0u8; PREFIX_LEN];
```
zhalvorsen: Using raw bytes here feels a bit brittle. Can we use a structure here with `zerocopy` that is tied to the original message to make sure it stays consistent with the Caliptra command:

```Rust
#[repr(C)]
#[derive(Debug, IntoBytes, FromBytes, Immutable, KnownLayout, PartialEq, Eq)]
pub struct PopulateIdevMldsa87CertReqHeader {
    pub header: MailboxReqHeader,
    pub cert_size: u32,
}

const _: () = assert!(offset_of!(PopulateIdevMldsa87CertReqHeader, header) == offset_of!(PopulateIdevMldsa87CertReq, header));
const _: () = assert!(offset_of!(PopulateIdevMldsa87CertReqHeader, cert_size) == offset_of!(PopulateIdevMldsa87CertReq, cert_size));
const _: () = assert!(size_of::<PopulateIdevMldsa87CertReqHeader>() + PopulateIdevMldsa87CertReq::MAX_CERT_SIZE == size_of::<PopulateIdevMldsa87CertReq>());
```

platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
```diff
@@ -145,3 +191,115 @@ async fn populate_idev_from_otp<A: ApiAlloc>(alloc: &A) -> McuResult<()> {
...
+            .read(partition_id, offset as u32)
+            .await
+            .map_err(|_| mcu_error::codes::INTERNAL_BUG)?;
+        let word_bytes = word.to_le_bytes();
+        for (d, s) in out.iter_mut().skip(offset).zip(word_bytes.iter()) {
+            *d = *s;
+        }
+    }
+
+    Ok(())
```
zhalvorsen, line 325: nit: I think the logic can be simplified here a little


```suggestion
    let mut offset = 0;

    while offset < out.len() {
        let word = otp
            .read(partition_id, offset as u32)
            .await
            .map_err(|_| mcu_error::codes::INTERNAL_BUG)?;
        
        let word_bytes = word.to_le_bytes();
        
        // Calculate how many bytes we actually need to copy (4 or whatever is remaining)
        let remaining = out.len() - offset;
        let copy_len = remaining.min(4);

        out[offset..offset + copy_len].copy_from_slice(&word_bytes[..copy_len]);
        
        offset += 4;
    }

    Ok(())
```

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
```diff
@@ -124,6 +131,44 @@ pub(crate) fn spawn_spdm_tasks(spawner: &Spawner) {
...
+/// cert store to be ready so the two installs do not contend for the mailbox,
+/// and nothing can read the ML-DSA chain yet, so ordering after SPDM comes up
+/// costs nothing.
+#[embassy_executor::task]
+async fn idev_mldsa_installer() {
+    // Wait for cert-store init (driven by whichever responder wins the race)
+    // before adding mailbox traffic of our own. A dedicated signal, because
+    // `CERT_STORE_DONE.wait()` consumes the value the responders need.
+    if !CERT_STORE_READY.wait().await {
+        return;
```
zhalvorsen, line 276: Should this return some type of status so the caller knows to try again later?

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
```diff
@@ -63,6 +63,11 @@ static CERT_STORE_DONE: Signal<CriticalSectionRawMutex, bool> = Signal::new();
 /// Cert store init state: 0 = uninit, 1 = in progress, 2 = done.
 static CERT_STORE_STATE: AtomicU8 = AtomicU8::new(0);
 
+/// Fired once with the cert-store init outcome, for the ML-DSA installer task.
+/// Separate from [`CERT_STORE_DONE`] because `Signal::wait` consumes the value
+/// and the responders are already contending for that one.
+static CERT_STORE_READY: Signal<CriticalSectionRawMutex, bool> = Signal::new();
```
amit-kumarh: Can we use a type like `embassy_sync::watch` that supports multiple consumers instead of creating multiple `Signal`s here?

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
```diff
@@ -124,6 +131,44 @@ pub(crate) fn spawn_spdm_tasks(spawner: &Spawner) {
...
+    // Sized for one ML-DSA-87 certificate (Caliptra caps it at 8192) plus the
+    // allocator bitmap and the 8-byte mailbox response — not the responders'
+    // 12 KiB, since this task only ever stages the certificate.
+    const MLDSA_SCRATCH_SIZE: usize = 9 * 1024;
+
+    #[repr(C, align(64))]
+    struct ScratchBuf([u8; MLDSA_SCRATCH_SIZE]);
+    static mut MLDSA_SCRATCH: ScratchBuf = ScratchBuf([0u8; MLDSA_SCRATCH_SIZE]);
+    // SAFETY: this task is the sole owner of `MLDSA_SCRATCH`.
+    let scratch_ptr: NonNull<u8> = unsafe { NonNull::new_unchecked(MLDSA_SCRATCH.0.as_mut_ptr()) };
```
zhalvorsen: nit: I'm not a big fan of raw `mut static` can we use `StaticCell` here instead?


```suggestion
    static MLDSA_SCRATCH: StaticCell<ScratchBuf> = StaticCell::new();
    let scratch = MLDSA_SCRATCH.init(ScratchBuf([0u8; MLDSA_SCRATCH_SIZE]));
    let scratch_ptr = core::ptr::NonNull::from(&mut scratch.0).cast::<u8>();
```

platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
```diff
@@ -145,3 +191,115 @@ async fn populate_idev_from_otp<A: ApiAlloc>(alloc: &A) -> McuResult<()> {
...
+    Ok(mldsa87_cert_der_len(word, partition_size))
+}
+
+/// Read `out.len()` bytes from the start of an OTP partition.
+///
+/// The driver is word-addressed, so whole 32-bit words are copied and any
+/// trailing 1..3 bytes are taken from the final word (the ML-DSA-87 cert is
+/// 7741 bytes, i.e. not a word multiple). Panic-free: every store goes through
+/// a checked slice, since a panic in this app is fatal.
+async fn read_otp_range(
```
amit-kumarh, line 304: nit: perhaps better in a follow-up, but we should have the ECC chain use this helper as well.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
```diff
@@ -124,6 +131,44 @@ pub(crate) fn spawn_spdm_tasks(spawner: &Spawner) {
...
+    }
+
+    // Sized for one ML-DSA-87 certificate (Caliptra caps it at 8192) plus the
+    // allocator bitmap and the 8-byte mailbox response — not the responders'
+    // 12 KiB, since this task only ever stages the certificate.
+    const MLDSA_SCRATCH_SIZE: usize = 9 * 1024;
+
+    #[repr(C, align(64))]
+    struct ScratchBuf([u8; MLDSA_SCRATCH_SIZE]);
+    static mut MLDSA_SCRATCH: ScratchBuf = ScratchBuf([0u8; MLDSA_SCRATCH_SIZE]);
```
amit-kumarh: The PR description says

> Streamed, not staged — 7.7 KiB won't fit the app's 12 KiB SPDM scratch pool; peak cost is one 256-byte chunk.

but it looks like we are allocating a separate 9k buffer here to stage the MLDSA cert. Can we update the description to match?

platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
```diff
@@ -31,23 +45,55 @@ const ECC_DEVID_CERT_SIZE: usize = 547;
...
-/// Initialize the cert store for all 3 slots.
+/// One-time Caliptra setup: read the IDevID certs from OTP and install them.
 ///
-/// Must be called once before SPDM responder tasks start.
-/// Slot 0 failure is fatal. Slots 1-2 stay unprovisioned if
-/// flash is empty (they'll be provisioned via SET_CERTIFICATE).
-/// One-time Caliptra setup: read IDevID from OTP and install.
+/// The ECC-384 install is required — its failure aborts cert-store init because
+/// every SPDM flow in use today signs with ECC. The ML-DSA-87 install is
+/// best-effort and *cannot* fail this function: no SPDM requester can negotiate
```
amit-kumarh: nit: While technically true righty now, we are in the process of adding SPDM 1.4 and MLDSA support, so we will be using the PQC codepaths. Might it be worth it to remove such references just so we don't have to come back for them later.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
```diff
@@ -124,6 +131,44 @@ pub(crate) fn spawn_spdm_tasks(spawner: &Spawner) {
...
+}
+
+/// Install the IDevID ML-DSA-87 certificate once the responders are serving.
+///
+/// Its own task because the install is ~1,900 sequential 4-byte OTP reads —
+/// doing it inside `ensure_cert_store_init` delayed responder startup past the
+/// point where an already-connected requester gives up. It waits for the ECC
+/// cert store to be ready so the two installs do not contend for the mailbox,
+/// and nothing can read the ML-DSA chain yet, so ordering after SPDM comes up
+/// costs nothing.
```
clundin25, line 259: Wonder if we should have a task dedicated to offloading OTP operations? I imagine this will be a bottleneck elsewhere, and a generic solution may save resources

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
amit-kumarh: Instead of making this memory `static`, should  we instead use a heap-allocated Vec or other buffer? That way, we don't have to permanently lock up this 9k of memory that's only used on boot and can at least release it back to the heap.

For example, as is done in `//platforms/emulator/runtime/userspace/apps/user/src/image_loader/mod.rs:345`

It looks like the heap size in `//platforms/emulator/runtime/userspace/apps/user/src/riscv.rs` is currently 8k, so we'd have to bump the size of it, but I think using some memory that can be returned to the heap is better than tying up the 9k forever

platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
amit-kumarh: also for a follow-up (not blocking): we should consider adding a chunked OTP read to the syscall interface

amit-kumarh (approved):
Could we actually merge this into `feature/spdm-1_4` branch as well? This is where we are performing all of the SPDM 1.4 / PQC work, and we will merge them all together into `main` from there

edit: created #2007 for this

RaunakGu:
@amit-kumarh , OTP read is not going to be there for all the integrator, do we need to add chunk based read?, I will create a discussion thread.

Did you already merge this to spdm-1.4 branch, or you want me to merge it there?

amit-kumarh:
@RaunakGu If we're reading a large PQC certificate, I think it makes sense to add a chunked read option to the `ExternalOtp` interface (regardless of what it's backed by, whether that's flash, OTP, etc.) so we don't have to make the round trip syscall for each word in the 8k cert. Glad to discuss further, that's an optimization that doesn't need to happen in this PR.

re : the spdm branch, I created #2007 to bring it up to date with main, which will include this PR on top, so should be all good from your end there.

parvathib (approved):

PR #1936, opened 2026-08-07
Add filename to flash image header

### LINKED ISSUE #1841: [Flash Layout] Filename field is missing from ImageHeader
Hi Caliptra team,

The latest flash layout document defines a 64-byte Filename field in each image information entry:

https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/docs/src/flash_layout.md

However, the ImageHeader structure in both main and main-2.1 does not include this field:

https://github.com/chipsalliance/caliptra-mcu-sw/blob/main/common/flash-image/src/lib.rs#L46
https://github.com/chipsalliance/caliptra-mcu-sw/blob/main-2.1/common/flash-image/src/lib.rs#L46

Current structure:

```rust
pub struct ImageHeader {
    pub identifier: u32,
    pub offset: u32,
    pub size: u32,
    pub image_checksum: u32,
    pub image_header_checksum: u32,
}
```

Could you please help synchronize the source code in both branches with the latest flash layout specification, including the Filename field and the related flash image generation and parsing logic?

Thanks!

mlvisaya:
Add the 64-byte filename field required by the flash layout and allow callers to set one optional network image filename when generating a combined flash image.

Fixes #1841


common/flash-image/src/lib.rs
```diff
@@ -47,11 +48,21 @@ pub struct ImageHeader {
     pub identifier: u32,
     pub offset: u32,
     pub size: u32,
+    pub filename: [u8; IMAGE_FILENAME_SIZE],
```
helloxiling: We expand `ImageHeader` from 20 to 84 bytes, but retains `HEADER_VERSION` 0x0002. Existing v2 images and ROMs use a 20-byte table stride, with no compatibility branch.  Shall we introduce a new format version or support both layouts explicitly?

builder/src/flash_image.rs
```diff
@@ -242,18 +242,29 @@ pub fn flash_image_create(args: &CaliptraBuildArgs) -> Result<()> {
...
     let flash_image = FlashImage::new(&images, &image_info);
     flash_image.write_to_file(offset, output_path)?;
 
     Ok(())
 }
 
-pub fn generate_image_info(images: Vec<FirmwareImage>) -> Vec<ImageHeader> {
+pub fn generate_image_info(
+    images: Vec<FirmwareImage>,
+    network_image_filename: Option<&str>,
```
helloxiling: Do we copy one filename into every image header?  Filename is per image and represents its TFTP path, so distinct image IDs generally require distinct paths. 
1. Should we consider storing filename on `ImageCfg` and test two entries with different names?  
2. It seems the production `all-build` caller never sets the new argument, and generated filenames may remain empty. Can we check that?

common/flash-image/src/lib.rs
mlvisaya: Good catch. I bumped `HEADER_VERSION` from `0x0002` to `0x0003` because adding the filename changes the `ImageHeader` wire layout from 20 to 84 bytes. Image generation and verification now both require v3; this change does not attempt to parse the v2 layout as v3.

builder/src/flash_image.rs
mlvisaya: Good points. I updated the implementation so the filename is per image: MCU RT and each SoC image use `ImageCfg::network_filename`, while Caliptra FMC/RT and the SoC manifest have dedicated optional `all-build` parameters. The flash-image test now covers distinct filenames for all five entries, including two different SoC image paths. `all-build` plumbs these values into base, feature-specific, and update images. They intentionally remain optional and zero-filled for current production invocations; those callers will supply their TFTP paths when network boot is implemented.

helloxiling: @mlvisaya , thanks for addressing the feedback. 
Should we also make changes to its consumer `FlashHeader::verify()` to check the version format so that ROM and runtime loaders can parse different headers properly?

builder/src/flash_image.rs
```diff
@@ -29,13 +29,20 @@ pub struct FlashImagePayload<'a> {
...
+    pub filename: [u8; MAX_FILENAME_LEN],
 }
 
 impl<'a> FirmwareImage<'a> {
-    pub fn new(identifier: u32, content: &'a [u8]) -> io::Result<Self> {
+    pub fn new(identifier: u32, content: &'a [u8], filename: Option<&[u8]>) -> io::Result<Self> {
+        let mut fname = [0u8; MAX_FILENAME_LEN];
+        if let Some(name) = filename {
+            let len = name.len().min(MAX_FILENAME_LEN);
+            fname[..len].copy_from_slice(&name[..len]);
```
helloxiling: We silently truncate filenames beyond 64 bytes, which could refer to the wrong TFTP object. Can we reject overlength paths instead?

mlvisaya: Addressed in `7547a3ec0`. `FirmwareImage::new()` now returns `InvalidInput` when the network filename exceeds 64 bytes instead of silently truncating it. The unit test covers the 64-byte boundary and rejection at 65 bytes.

builder/src/flash_image.rs
mlvisaya: Yes. `FlashHeader::verify()` now rejects any version other than `HEADER_VERSION` (`0x0003`) before accepting the header. Both firmware-update consumers call `verify()` before walking the image table, so they cannot interpret a v2 table using the v3 `ImageHeader` stride. I added a checksum-valid regression test confirming v3 is accepted and v2 is rejected in `7547a3ec0`.

helloxiling (approved):
Looks good to me.

mlvisaya:
thank you!

PR #1939, opened 2026-08-08
fix(caliptra-util-host): restore import ordering to unblock nightly precheckin

parvathib:
Fix nightly precheckin failure in caliptra-util-host


mlvisaya (approved):
LGTM

PR #1940, opened 2026-08-08
Add production support for FirmwareVersion and DeviceCapabilities commands

helloxiling:
Backport of #1906 to `main-2.1`.

The only intentional behavior difference is that MCU Runtime reports firmware version `2.1.0`.


mlvisaya (approved):
LGTM

PR #1941, opened 2026-08-09
Fix stale runtime SoC image descriptors

mlvisaya:
cargo xtask runtime built the MCU user app before CaliptraBuilder generated the SoC image descriptor configuration. If target/generated contained files from an earlier build, stale firmware IDs were baked into the runtime while the new Caliptra authorization manifest did not contain matching image metadata. GET_IMAGE_INFO (IME0) then failed with RUNTIME_IMAGE_METADATA_NOT_FOUND (0x000e0061).

Generate the attestation and SoC image descriptor configuration from the current --soc-images arguments before building the runtime. This also clears stale IDs for a plain runtime invocation. Treat an empty SoC image activation list as a successful no-op so it does not send an invalid zero-count ACTIVATE_FIRMWARE request.


helloxiling (commented):
The fix looks good. Just wondering why we haven't hit this issue in our CI testing. Shall we add test cases?

mlvisaya:
Good point. CI did not exercise `cargo xtask runtime` through the userspace image-loading path with stale generated SoC image descriptors, so it missed this stateful build-order issue.

I added commit `e373c3e3e` with a precheckin smoke test that:
- seeds `target/generated` with a stale SoC image descriptor,
- invokes the real `xtask runtime` flow using the normal default features plus `test-xtask-runtime`,
- exits successfully only after the existing image-loading flow completes, and
- propagates emulator failures back to precheckin.

Without the fix, the stale image ID is compiled into userspace and the test fails with the `IME0` metadata error. With the fix, the generated configuration is refreshed before compilation and the test passes. `cargo xtask precheckin` passes with this coverage.

helloxiling (approved):
Thanks for addressing the comments. The latest patch looks good.

PR #1947, opened 2026-08-10
Add hitless firmware update attestation tests

parvathib:
Adds three hitless-update attestation variants (hitless, hitless-tcb, hitless-mixed) and fixes the byte-order mismatch that hung them at ACTF.

caliptra-sw's AxiRootBus decoded MemoryReadResponse payloads big-endian while AxiRootBus::write emits to_le_bytes(), so every event-backed DMA copy byte-swapped each 32-bit word. That corrupted the staging -> MCU SRAM copy: authorize_image saw a hash mismatch and activate_fw returned early without restoring FW_EXEC_CTRL[2], leaving the MCU halted. Paired with chipsalliance/caliptra-sw#4076; both halves must land together.

Producer side, establishing the invariant that every event-backed MemoryRead producer serializes little-endian:
  - root_bus.rs: drop the ExternalTestSram 4-byte chunk reversal, which only compensated for the big-endian decode
  - i3c.rs: serialize recovery-interface reads little-endian
  - root_bus.rs: pass a word index to read_mcu_mbox0_csr_mbox_sram, which multiplies by 4 internally

Also: - mci.rs: keep polling while the MCU is halted so FW_EXEC_CTRL[2] is
    observed and the MCU resumes
  - model_emulated.rs: pass soc_ifc to Mci so hitless reset can drive ss_generic_fw_exec_ctrl
  - i3c_socket.rs, mctp_transport.rs: shut the socket down on disconnect so the server can accept the next client
  - update_sm.rs: don't unwrap a send on a closed queue
  - test_mctp_spdm_attestation.rs: pick validator ports randomly; the default first-free scan handed back the port just released by the cold-boot validator, still in TIME_WAIT


mlvisaya:
looks good to me

mlvisaya (approved):

PR #1952, opened 2026-08-11
Migrate userspace to caliptra API lite

mlvisaya:
Replace remaining userspace consumers of caliptra-mcu-libapi-caliptra with mcu-caliptra-api-lite, add allocator-backed firmware update and SHA-512 support, migrate example tests, and remove the legacy API crate.


platforms/emulator/runtime/userspace/apps/example/src/main.rs
```diff
@@ -168,33 +168,31 @@ pub(crate) async fn async_main<S: Syscalls>() {
     {
         let pcr_quote_alloc = test_get_device_state::init_pcr_quote_allocator();
```
parvathib: nit: we could call this "test allocator" instead of pcr quote allocator. or just alloc like below.

platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/device_ops.rs
```diff
@@ -355,6 +357,23 @@ pub fn revoke_vendor_pk_hash(vendor_pk_hash_slot: u32) -> CaliptraCmdResult<()>
         .map_err(|_| CaliptraCompletionCode::OperationFailed)
 }
 
+async fn hash_all<A: ApiAlloc>(
```
parvathib: Could we make this common and move it into caliptra-api-lite's sha.rs instead?

Also, we should validate the length against the algo before the copy.

platforms/emulator/runtime/userspace/apps/user/src/firmware_update/mod.rs
```diff
@@ -37,6 +46,19 @@ pub async fn firmware_update<D: DMAMapping>(
         // Device rebooted due to firmware update, skip firmware update
         return Ok(());
     }
+
+    let mut scratch = Vec::new();
+    scratch
+        .try_reserve_exact(FW_UPDATE_SCRATCH_SLOTS)
+        .map_err(|_| ErrorCode::Fail)?;
```
parvathib: ErrorCode::NoMem  may be better to return here than just Fail.

platforms/emulator/runtime/userspace/apps/user/src/firmware_update/mod.rs
```diff
@@ -37,6 +46,19 @@ pub async fn firmware_update<D: DMAMapping>(
         // Device rebooted due to firmware update, skip firmware update
         return Ok(());
     }
+
+    let mut scratch = Vec::new();
```
parvathib: Why heap Vec  for the scratch rather than the  `static  +  StaticBitmapAllocatorCell`  pattern already used by  `mcu_mbox  (16 KB)` and the `SPDM tasks (12 KB each)`? Static moves an under-sizing failure to link time, instead of a runtime  `try_reserve_exact`   ->  `ErrorCode::Fail  →  System::exit(1)`  mid-update.

Is it a RAM call, and if so that's fair.  The tests seem to pass with the current HEAP_MEM. 
Not asking to change it. But since the tradeoff is `cheaper RAM, runtime failure mode`, could we add a note on `FW_UPDATE_SCRATCH_SIZE` so integrators keep it in mind when tuning the memory?

platforms/emulator/runtime/userspace/apps/example/src/main.rs
```diff
@@ -168,33 +168,31 @@ pub(crate) async fn async_main<S: Syscalls>() {
...
+        test_caliptra_crypto::test_caliptra_ecdh(alloc).await;
+        test_caliptra_crypto::test_caliptra_hmac(alloc).await;
+        test_caliptra_crypto::test_caliptra_aes_gcm_cipher(alloc).await;
+        test_caliptra_crypto::test_caliptra_ecdsa(alloc).await;
         System::exit(0);
     }
 
     #[cfg(feature = "test-caliptra-certs")]
     {
-        // test_caliptra_certs::test_get_idev_csr().await;
```
parvathib, line 190: Any reason why this is dropped? can we keep this in case if we have to ever reprovision our test environment?

runtime/userspace/api/caliptra-api-lite/src/dpe.rs
```diff
@@ -578,49 +578,6 @@ pub async fn dpe_get_cert_chain_chunk<A: ApiAlloc>(
...
 
-/// Invoke DPE `CertifyKey` (P-384 / SHA-384, X.509 format) for the
-/// default context handle and the given 48-byte `label`. Writes the
-/// emitted leaf certificate DER into `dst` and returns the rotated context
-/// handle plus the number of bytes actually written.
-///
-/// Prefer [`dpe_certify_key_cert_slice`] when the caller only needs a
-/// slice of the certificate.
-#[inline(never)]
-pub async fn dpe_certify_key<A: ApiAlloc>(
```
parvathib, line 589: Could we keep this one? It should not cost any binary size. Some integrators may want to use it for smaller certs.

platforms/emulator/runtime/userspace/apps/example/src/main.rs
mlvisaya: Done in 8ad3b4489. Renamed the test allocator to `alloc`.

platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/device_ops.rs
mlvisaya: Done in 8ad3b4489. The shared one-shot helper now lives in `caliptra-api-lite` as `sha::hash_all`, and it validates the output length against `algo.hash_size()` before copying.

runtime/userspace/api/caliptra-api-lite/src/image_loader/mod.rs
```diff
@@ -215,6 +216,10 @@ fn convert_dma_cptra_addr_to_mcu_addr(
         .map_err(|_| ErrorCode::Fail)
 }
 
+pub async fn core_image_info(image_id: u32) -> Result<GetImageInfoResp, ErrorCode> {
+    get_image_info(&Mailbox::new(), image_id).await
```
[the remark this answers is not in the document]
mlvisaya: Fixed in 8ad3b4489. `get_image_info` now retains the mailbox response length and parses only the returned bytes, so a truncated response is rejected. I also added a short-response regression test.

platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
```diff
@@ -1,506 +1,175 @@
...
-pub async fn test_caliptra_ecdsa() {
-    println!("Starting Caliptra mailbox ECDSA test");
-    test_ecdsa().await;
-    println!("ECDSA test completed successfully");
+        .unwrap_or_else(|_| test_exit(1));
+    let mut mac = [0u8; 48];
+    cm_hmac(alloc, &okm, &data, &mut mac)
+        .await
+        .unwrap_or_else(|_| test_exit(1));
+    println!("HMAC/HKDF test completed successfully: {}", HexBytes(&mac));
```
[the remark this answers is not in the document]
mlvisaya: Fixed in 8ad3b4489. Restored the deterministic expected-MAC comparison, so the test again validates the HMAC/HKDF output rather than only command success.

platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
```diff
@@ -1,506 +1,175 @@
...
-
-    match Ecdsa::ecdsa_verify(pubkey_x, pubkey_y, &signature, msg_hash).await {
-        Ok(_) => {
-            println!("ECDSA signature verified successfully");
-        }
-        Err(e) => {
-            println!("Failed to verify ECDSA signature: {:?}", e);
-            test_exit(1);
-        }
+        .unwrap_or_else(|_| test_exit(1));
```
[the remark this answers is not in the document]
mlvisaya: Fixed in 8ad3b4489. The test now hashes a message, signs it through DPE, and verifies the signature against the returned P-384 public key using the `ECDSA384_SIGNATURE_VERIFY` (`ECV2`) command.

platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_certs.rs
```diff
@@ -57,430 +69,81 @@ const SIGNED_IDEV_CERT_DER: [u8; 541] = [
...
-        }
+pub async fn test_certify_key(alloc: &BitmapAllocator) {
+    let (_, size) = dpe_certify_key_cert_size(alloc, None, &TEST_KEY_LABEL)
+        .await
+        .unwrap_or_else(|_| test_exit(1));
+    if size == 0 {
+        test_exit(1);
     }
-    println!("Sign with attestation key test completed successfully");
+    println!("Attestation key certificate size: {}", size);
```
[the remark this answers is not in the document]
mlvisaya: Fixed in 8ad3b4489. The test now reads the complete certificate body in chunks with `dpe_certify_key_cert_slice` after obtaining its size.

platforms/emulator/runtime/userspace/apps/user/src/firmware_update/mod.rs
mlvisaya: This path was removed in 8ad3b4489 when the firmware-update scratch storage was changed from a heap `Vec` to a static allocator. There is no longer a `try_reserve_exact` error to map to `ErrorCode::Fail` or `NoMem`.

platforms/emulator/runtime/userspace/apps/user/src/firmware_update/mod.rs
mlvisaya: Changed in 8ad3b4489. Firmware update now uses a dedicated 2 KiB `StaticBitmapAllocatorCell`, matching the existing static-allocation pattern. I also reduced the user-app stack by 2 KiB to offset the static pool.

platforms/emulator/runtime/userspace/apps/example/src/main.rs
mlvisaya: Restored in 8ad3b4489. The `test_get_idev_csr` helper and its commented invocation are preserved for reprovisioning. It remains opt-in because the command is provisioning-state dependent and fails in the normal provisioned test environment.

runtime/userspace/api/caliptra-api-lite/src/dpe.rs
mlvisaya: Restored in 8ad3b4489. `dpe_certify_key` remains available and is implemented safely by composing the bounded certificate-size and certificate-slice operations, avoiding the oversized one-shot mailbox behavior.

parvathib (approved):
LGTM

PR #1957, opened 2026-08-11
[dpe]: Lazily initialize the exported cdi context

clundin25:
Export the CDI on first call to the get dpe signer cert command.

This avoids placing large ML-DSA and ECDSA certificates in RAM.


runtime/userspace/api/caliptra-api/src/signer.rs
```diff
@@ -1,36 +1,45 @@
...
+        Self { mailbox }
     }
 }
 
 #[async_trait]
 impl<'a> OcpLockSigner for CaliptraDpeSigner<'a> {
-    async fn sign(&self, label: &[u8], data: &[u8], signature: &mut [u8]) -> CaliptraApiResult<()> {
+    async fn sign(
+        &self,
+        _label: &[u8],
```
zhalvorsen, line 39: nit: Now that the label is no longer used, should this parameter be dropped?

zhalvorsen (approved):

runtime/userspace/api/caliptra-api/src/signer.rs
clundin25: Going to leave this for now, @arul-sri is about to touch this code too, so I'll remove it once he has finished his work

PR #1964, opened 2026-08-12
Reduce all-features FPGA SRAM usage

mlvisaya:
Keep mandatory DOE support while fitting the release all-features build in 512 KiB SRAM.

- use allocator-backed DPE certificate operations to shrink retained mailbox futures
- stream ML-DSA verification requests instead of reserving an 11 KiB static request
- share the production SPDM VDM backend type across MCTP and DOE
- reduce each SPDM scratch pool from 12 KiB to 8 KiB

This reduces user-app .bss from 96,892 bytes to 53,652 bytes and leaves 8 KiB of rounded SRAM headroom.


helloxiling (approved):
Looks good to me.

PR #1971, opened 2026-08-13
[auth-cmds] Further unify command authorization and execution

zhalvorsen:
There are two flows that callers can use to execute authorized commands.

1. MCU mailbox
2. SPDM-VDM commands

Previously the execution was written separately. This means two ways that would need to be maintained and fixed. This also leaves two separate avenues where bugs or vulnerabilities could be introduced.

This change unifies the authorization and handlers so there is a single source of truth. A new `AuthorizedCmdExecutor` struct has been added that both mechanism now use.

This also means all authorized commands are now supported in both with the exception of the raw fuse operations which are currently denylisted in the SPDM-VDM path.

In order to make authorization uniform, the `MailboxReqHeader` has been dropped from the authorized portion of the command for mailbox commands. This field only contains a checksum of the command so it wasn't critical.

The original handlers from runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs have been moved without modification to the new authorized command file. They will be altered to work with both handlers in a subsequent commit to make it easier to review.


runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
```diff
@@ -0,0 +1,317 @@
...
+    if e == codes::MAILBOX_BUSY {
+        CaliptraCompletionCode::CaliptraMailboxBusy
+    } else if e == codes::INVARIANT || e == codes::INTERNAL_BUG {
+        CaliptraCompletionCode::OperationFailed
+    } else if e.domain() == mcu_error::domain::MEMORY {
+        CaliptraCompletionCode::InsufficientResources
+    } else {
+        CaliptraCompletionCode::GeneralError
+    }
+}
```
clundin25: nit: Looks like a `From` implementation?

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
```diff
@@ -0,0 +1,317 @@
...
+    let otp: Otp<DefaultSyscalls> = Otp::new();
+    otp.write_raw(req.word_addr, req.data, req.mask)
+        .map_err(|e| match e {
+            caliptra_mcu_libtock_platform::ErrorCode::Invalid => {
+                CaliptraCompletionCode::InvalidParameter
+            }
+            _ => CaliptraCompletionCode::OperationFailed,
+        })?;
+
+    Ok(0)
```
clundin25, line 119: Is this always 0? Maybe worth removing `usize` or including a comment

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
```diff
@@ -0,0 +1,317 @@
...
+            .read(otp::reg::CALIPTRA_FW_SVN, i as u32)
+            .map_err(|_| CaliptraCompletionCode::OperationFailed)?;
+    }
+
+    // Convert the fuses to the SVN value
+    let fused_min_svn = {
+        // Value is take as the most significant bit set in fuses
+        let fuse: u128 = u128::from_le_bytes(current_fuses.as_bytes().try_into().unwrap());
+        128 - fuse.leading_zeros()
+    };
```
clundin25: This makes me think maybe we should have some `Svn` type

clundin25 (approved):

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
zhalvorsen: That's a good call. I'll do that in a follow up. We will probably need multiple definitions depending on which SVN we are referring to `Svn32`, `Svn64`, and `Svn128`

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
zhalvorsen: The only one that is non-zero is the FuseRead command

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
zhalvorsen: Also, a good idea. This was copy-pasted from the other file, but I'll make the change in this PR

romtime/src/mci_mbox_cmd.rs
parvathib, line 188: I think this will panic for some partitions. `MAX_FUSE_DATA_WORDS` is 128 words (512 bytes), but the buffer it fills (`resp.payload.data`) is now only 128 bytes (32 words).

May be pass  `MAX_FUSE_DATA_SIZE / 4`  here, matching what the runtime handler does in  `authorized_command.rs:83`?. 
Or redefine `MAX_FUSE_DATA_WORDS`  as  `MAX_FUSE_DATA_SIZE / 4` in `messages.rs`

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
```diff
@@ -0,0 +1,332 @@
...
+            }
+            CommandId::MC_FE_PROG => handle_fe_prog(alloc, payload).await,
+            CommandId::MC_FUSE_REVOKE_VENDOR_PUB_KEY => {
+                handle_revoke_vendor_pub_key(alloc, payload).await
+            }
+            CommandId::MC_FUSE_REVOKE_VENDOR_PK_HASH => handle_revoke_vendor_pk_hash(payload).await,
+            CommandId::MC_FUSE_READ => handle_fuse_read(payload, resp_buf).await,
+            CommandId::MC_FUSE_WRITE => handle_fuse_write(payload).await,
+            CommandId::MC_FUSE_LOCK_PARTITION => handle_fuse_lock_partition(payload).await,
+            _ => Err(CaliptraCompletionCode::UnsupportedOperation),
```
parvathib, line 55: Rebase heads-up: main has added  authorized  DOT commands and sub-commands since the merge-base - `MC_DEVICE_OWNERSHIP_TRANSFER`  family (#1930, four inner sub-commands). 
After a rebase, unless updated, they'll all fall through to this arm and silently start returning `UnsupportedOperation`.

DOT is a two-level dispatch (inner cmd ID is the first 4 bytes of  payload ), so it'd need a nested match here rather than another flat arm. Maybe worth discussing and agreeing on that shape before the rebase?

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
```diff
@@ -0,0 +1,332 @@
...
+        .map_err(|_| CaliptraCompletionCode::OperationFailed)?;
+    if booted_pk_hash == pk_hash_from_slot {
+        return Err(CaliptraCompletionCode::InvalidParameter);
+    }
+
+    otp.revoke_vendor_pk_hash(vendor_pk_hash_slot)
+        .map_err(|_| CaliptraCompletionCode::OperationFailed)?;
+
+    Ok(0)
+}
```
parvathib: NIT: the  `..`  here drops `flags`  (and `reserved` at lines 243 and 300), so the non-zero checks that were on the VDM path are gone.

It'd nice to have those checks.

Maybe add  if `req.flags != 0 { return Err(CaliptraCompletionCode::InvalidParameter); }`  here, and the same for  `req.reserved`  in the two revoke handlers?

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
```diff
@@ -0,0 +1,332 @@
...
+
+async fn handle_increase_caliptra_min_svn<A: ApiAlloc>(
+    alloc: &A,
+    payload: &[u8],
+) -> CaliptraCmdResult<usize> {
+    // Decode the request
+    let req = FuseIncreaseCaliptraMinSvnReqPayload::ref_from_bytes(payload)
+        .map_err(|_| CaliptraCompletionCode::InvalidPayloadSize)?;
+    let FuseIncreaseCaliptraMinSvnReqPayload { svn, .. } = *req;
```
parvathib: NIT: the  ..  here drops flags  (and reserved at lines 243 and 300), so the non-zero checks that were on the VDM path are gone.

It'd nice to have those checks.

Maybe add  if req.flags != 0 { return Err(CaliptraCompletionCode::InvalidParameter); }  here, and the same for  req.reserved  in the two revoke handlers?

romtime/src/mci_mbox_cmd.rs
zhalvorsen: Good catch! I redefine `MAX_FUSE_DATA_WORDS` as `MAX_FUSE_DATA_SIZE / 4`

runtime/userspace/api/caliptra-common-commands/src/authorized_command.rs
zhalvorsen: Yeah, lets keep those same checks. I added them back

PR #1972, opened 2026-08-13
docs: remove KV store service from MCU SDK spec 

helloxiling:
Removes the Key-Value Store service from the MCU documentation, since it is not a service the MCU SDK provides for now.


mlvisaya (approved):
LGTM

PR #1973, opened 2026-08-14
Fix caliptra-util-host validator failures

mlvisaya:
Remove the GetDeviceCapabilities precheck from the SPDM VDM validator because that command is only supported over MCTP VDM.

Parse FE_PROG authorized requests from byte slices instead of an aligned zerocopy reference. SPDM VDM payloads begin at an unaligned offset, which previously returned InvalidParameter before signature verification and caused authorization-negative tests to receive 0x02 instead of AccessDenied 0x0c. Add regression coverage for the unaligned wire payload.

Make isolated SPDM fuse suites signal completion and failure through per-test state instead of terminating the entire Cargo process. This allows all prefix-matched suites to run serially without hanging or stopping after the first suite.

Enable userspace logging for the MCTP VDM validator feature so ClearDebugLog is available. Update CI to run the current isolated SPDM suites, build one filtered firmware bundle for all validator tests, reuse it through CPTRA_FIRMWARE_BUNDLE, and run precheckin concurrently with validator tests.


parvathib (approved):
LGTM

PR #2005, opened 2026-08-19
ocp-lock: support ML-DSA keys from DPE for endorsement certs

### LINKED ISSUE #1732: [OCP-LOCK]: Add ML-DSA support

arul-sri:
Implement EndorsementAlgorithm enum to support both ECDSA P-384 and ML-DSA-87 (FIPS 204) signatures from DPE for OCP LOCK endorsement certificates.

- Add CaliptraDpeSigner::with_algorithm for selecting the signing algorithm.
- Dispatch SIGN_WITH_EXPORTED_MLDSA commands with MldsaSignType::Mu.
- Update X.509 certificate builder with standard ML-DSA-87 and direct BIT STRING signature encoding.
- Expand max endorsement certificate buffer to 8 KB.

Fixes #1732


runtime/userspace/api/caliptra-libapi/src/ocp_lock.rs
```diff
@@ -183,6 +199,9 @@ pub struct Certificate<'a> {
 pub trait OcpLockSigner: Send + Sync {
     async fn sign(&self, label: &[u8], data: &[u8], signature: &mut [u8]) -> CaliptraApiResult<()>;
     fn signature_size(&self) -> usize;
+    fn algorithm(&self) -> EndorsementAlgorithm {
+        EndorsementAlgorithm::EcdsaP384Sha384
+    }
```
clundin25: Let's omit a default implementation

runtime/userspace/api/caliptra-libapi/src/ocp_lock.rs
```diff
@@ -460,39 +484,54 @@ impl<'a> OcpLock<'a> {
...
+                    },
+                    signature: BitStringRef::new(0, sig_der_slice)?,
+                }
+            }
+            EndorsementAlgorithm::MlDsa87 => {
+                let actual_sig = if sig_len == 4628 && signature_bytes[4627] == 0 {
+                    &signature_bytes[..4627]
+                } else {
+                    &signature_bytes[..sig_len]
+                };
```
clundin25: I think this is some weirdness that happens when encoding to ASN.1?

Do you mind adding a comment to document why we did this

common/mcu-mbox/src/messages.rs
```diff
@@ -1815,6 +1815,8 @@ impl McuResponseVarSize for ExportAttestedCsrResp {}
 #[derive(Debug, IntoBytes, FromBytes, Immutable, KnownLayout, PartialEq, Eq, Default)]
 pub struct DpeSignerContextCertReq {
     pub hdr: MailboxReqHeader,
+    /// Asymmetric algorithm (0x0001=ECC384, 0x0002=MLDSA87)
+    pub algorithm: u32,
```
clundin25: Can you treat this as a type, and define each algorithm as an associated constant? That prevents any logical mistakes.

E.g.

```rust
pub struct EndorsementAlgorithm(u32); 

impl EndorsementAlgorithm {
  const ECDSA_384: Self = Self(0x1);
  const MLDSA_87: Self = Self(0x2);
}

pub struct DpeSignerContextCertReq {
    pub hdr: MailboxReqHeader,
    pub algorithm: EndorsementAlgorithm,
}
```

runtime/userspace/api/caliptra-libapi/src/ocp_lock.rs
```diff
@@ -207,14 +224,17 @@ impl OcpLock<'_> {
     pub const P384_SIGNATURE_SIZE: usize = 96;
     pub const P384_PUB_KEY_SIZE: usize = 1 + 2 * Self::P384_SCALAR_SIZE;
 
+    // ML-DSA-87 signature size (FIPS 204)
+    pub const MLDSA87_SIGNATURE_SIZE: usize = 4628;
```
clundin25: I think this is padded by 1 byte. Can we use 4627? I think that means we have to do some padding elsewhere, but I think it makes the real size clear

clundin25: Since we need word alignment in the mailbox, we usually add a `_padding: [u8; 1]` to make it explicit and only appear in the mailbox API

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -821,9 +821,11 @@ impl<'a, H: CaliptraCmdHandler, A: CommandAuthorizer, Alloc: McuMboxScratch>
 
     async fn handle_dpe_signer_context_cert<'r>(
         &mut self,
-        _req: &[u8],
+        req: &[u8],
         resp_buf: &'r mut [u8],
     ) -> McuResult<(&'r mut [u8], MbxCmdStatus)> {
+        let _req = DpeSignerContextCertReq::ref_from_bytes(req)
+            .map_err(|_| errors::INVALID_PARAMS)?;
```
clundin25: I think we are missing the code to determine if we need to get the ECDSA or ML-DSA certificate chain from Caliptra

common/mcu-mbox/src/messages.rs
arul-sri: Done. Defined `EndorsementAlgorithm(pub u32)` with associated constants `ECDSA_384 = Self(0x1)` and `MLDSA_87 = Self(0x2)`, and updated `DpeSignerContextCertReq` and `GetOcpLockEndorsementCertReq` to use this type across the codebase.

runtime/userspace/api/caliptra-libapi/src/ocp_lock.rs
arul-sri: Done. Updated `MLDSA87_SIGNATURE_SIZE` and `MAX_SIGNATURE_BYTES` to the exact FIPS 204 size of `4627` bytes, and removed the manual padding strip workaround when constructing the X.509 `BitStringRef`. (The mailbox response struct `SignWithExportedMldsaResp` in `caliptra-api` handles the 1-byte mailbox alignment explicitly with `_padding: [u8; 1]`.)

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
arul-sri: Good catch. `handle_dpe_signer_context_cert` now checks `req.algorithm`:
- `EndorsementAlgorithm::ECDSA_384` invokes DPE with `DPE_PROFILE_P384_SHA384` (`CMD_INVOKE_DPE`).
- `EndorsementAlgorithm::MLDSA_87` invokes DPE with `DPE_PROFILE_MLDSA87` (`CMD_INVOKE_DPE_MLDSA87`).
- Any other value returns `INVALID_PARAMS`.

Also added `profile: u32` as a parameter to `export_cdi_and_stash` and `dpe_derive_context_exported_cdi` to pass the selected profile down to the DPE mailbox, and increased `DPE_MAX_LEAF_CERT_SIZE` to 8 KiB to accommodate ML-DSA leaf certificates.

common/mcu-mbox/src/messages.rs
```diff
@@ -587,13 +588,11 @@ pub enum McuMailboxResp {
     FuseRevokeVendorPkHash(FuseRevokeVendorPkHashResp),
     // Certificate commands
     ExportAttestedCsr(ExportAttestedCsrResp),
-    DpeSignerContextCert(DpeSignerContextCertResp),
```
clundin25, line 590: It looks like these commands were inadvertently deleted?

platforms/emulator/runtime/userspace/apps/user/src/caliptra_cmd_handler/mod.rs
```diff
@@ -469,11 +471,17 @@ impl CaliptraCmdHandler for CaliptraCmdBackend {
...
         hpke_handle: &HpkeHandle,
+        algorithm: MboxEndorsementAlgorithm,
         cert_buf: &mut [u8],
     ) -> CaliptraCmdResult<usize> {
-        let mailbox = Mailbox::new();
+        let algo = match algorithm {
+            MboxEndorsementAlgorithm::ECDSA_384 => EndorsementAlgorithm::EcdsaP384Sha384,
+            MboxEndorsementAlgorithm::MLDSA_87 => EndorsementAlgorithm::MlDsa87,
+            _ => return Err(CaliptraCompletionCode::InvalidParameter),
+        };
```
clundin25: I think we should implement a `TryFrom<MboxEndorsementAlgorithm> for EndorsementAlgorithm` so we can re-use this conversion in multiple places

tests/integration/src/runtime/test_ocp_lock.rs
```diff
@@ -426,6 +430,177 @@ fn test_get_ocp_lock_endorsement_cert_cmd() -> Result<()> {
             expected_key_len,
             "Public key length mismatch"
         );
+
+        // 4. Get OCP LOCK Endorsement Certificate (ML-DSA-87) from MCU Mailbox
+        let mldsa_cmd = caliptra_mcu_mbox_common::messages::GetOcpLockEndorsementCertReq {
```
clundin25: I think we are missing checking that the endorsement certificate is signed by the DPE signer context from `DpeSignerContextCertReq`

runtime/userspace/api/caliptra-api/src/dpe.rs
```diff
@@ -500,13 +526,41 @@ pub async fn dpe_derive_context_exported_cdi<A: ApiAlloc>(
...
+        24 * 1024
+    } else {
+        size_of::<InvokeDpeRespPrefix>()
+            + size_of::<DeriveContextExportedCdiRespPrefix>()
+            + cert_dst.len().min(DPE_MAX_LEAF_CERT_SIZE)
+    };
     let mut rsp = alloc.alloc(max_resp_len)?;
+    let axi_response = if profile == DPE_PROFILE_MLDSA87 {
+        const MCU_SRAM_LOCAL_BASE: u32 = 0x4000_0000;
+        const MCU_SRAM_AXI_BASE: u32 = 0xA8C0_0000;
```
clundin25: Are there any pre-existing constants we can use? Also we may accidentally overwrite a slice of the SRAM that has reserved some space for data.

I am not sure what the best solution is, hopefully there is an existing pattern we can re-use?

tests/integration/src/lib.rs
```diff
@@ -1239,6 +1239,7 @@ mod test {
     }
 
     pub fn start_runtime_hw_model(params: TestParams) -> DefaultHwModel {
+        std::env::set_var("CPTRA_EMULATOR_SS_MCI_OFFSET", "0x00000000a8000000");
```
clundin25: What happened here? Is this correct?

runtime/userspace/api/caliptra-api/src/dpe.rs
arul-sri: 1. **Constants**:
   These addresses come directly from the Caliptra hardware specification:
   - MCU local SRAM starts at `0x4000_0000`.
   - On the subsystem AXI bus, MCU SRAM is located at `0xA800_0000 + 0x00C0_0000` (`0xA8C0_0000`).
   
   I updated the code to define `MCI_BASE_AXI_ADDRESS` (`0xA800_0000`) and `MCU_SRAM_AXI_OFFSET` (`0x00C0_0000`) explicitly with doc comments so the derivation is clear.

2. **SRAM Safety**:
   There is no risk of overwriting reserved memory because `rsp` is not a hardcoded address, instead it is dynamically allocated from the scratch pool (`alloc.alloc(max_resp_len)`). Caliptra RT is also told `axi_max_size = max_resp_len`, so the DMA transfer is strictly bounded to stay within this allocated buffer.

tests/integration/src/lib.rs
arul-sri: `CPTRA_EMULATOR_SS_MCI_OFFSET` sets the emulator's subsystem MCI base address (`0xA800_0000`) so the Caliptra HW model can route AXI DMA writes back to the MCU SRAM.

It shouldn't have been set globally in `start_runtime_hw_model` in `lib.rs`. I have removed it from `lib.rs` and moved it into the specific test (`test_get_ocp_lock_endorsement_cert_mldsa_cmd`) under `TEST_LOCK`, matching the pattern used by `test_soc_boot.rs`, `test_mctp_spdm_attestation.rs`, and `test_firmware_update.rs`.

runtime/userspace/api/caliptra-api/src/dpe.rs
clundin25: Ah yes, you are right this is copying directly into the reserved heap memory.

However, I believe the FPGA and the emulator have a different SRAM offset.

Can we define these constants elsewhere in a platform specific way, and plumb that to the DPE layer?

runtime/userspace/api/spdm-lib/stack/src/digests.rs
clundin25, line 61: @arul-sri I am not sure this change should be here, maybe it's why the SPDM test fails?

tests/integration/src/runtime/test_ocp_lock.rs
```diff
@@ -430,3 +436,278 @@ fn test_get_ocp_lock_endorsement_cert_cmd() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_get_ocp_lock_endorsement_cert_mldsa_cmd() -> Result<()> {
+    let _lock = crate::test::TEST_LOCK.lock().unwrap();
+    std::env::set_var("CPTRA_EMULATOR_SS_MCI_OFFSET", "0x00000000a8000000");
```
clundin25: Is it possible to write the test without this? I think this is a pattern we should avoid, because it makes the tests brittle.

What problem does it address?

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
```diff
@@ -1724,7 +1734,12 @@ fn response_buffer_size<H: CaliptraCmdHandler>(cmd: u32) -> usize {
             size_of::<OcpLockEnumerateHpkeHandlesResp>()
         }
         c if c == CommandId::MC_DPE_SIGNER_CONTEXT_CERT => {
-            size_of::<MailboxRespHeaderVarSize>() + 2048
+            size_of::<MailboxRespHeaderVarSize>()
+                + if cfg!(feature = "ocp-lock") {
+                    caliptra_mcu_mbox_common::messages::MAX_ENDORSEMENT_CERT_SIZE
+                } else {
+                    caliptra_mcu_mbox_common::messages::MAX_RESP_DATA_SIZE
+                }
```
clundin25: I think we should instead increase the `MAX_RESP_DATA_SIZE`, so there is no drift between OCP LOCK enabled firmware

runtime/userspace/api/caliptra-libapi/src/ocp_lock.rs
```diff
@@ -191,7 +233,7 @@ pub struct OcpLock<'a> {
 }
 
 impl OcpLock<'_> {
-    pub const MAX_ENDORSEMENT_CERT_SIZE: usize = 2048;
+    pub const MAX_ENDORSEMENT_CERT_SIZE: usize = 12 * 1024;
```
clundin25, line 247: Can you add TODO + tracking bug to calculate this at build time

common/mcu-mbox/src/messages.rs
clundin25, line 590: It looks like these commands are still removed

runtime/userspace/api/spdm-lib/stack/src/digests.rs
arul-sri: Reverted this change back to the main-2.1 baseline.

tests/integration/src/runtime/test_ocp_lock.rs
arul-sri: When Caliptra RT executes DPE ML-DSA signing, Caliptra writes the signature to address 0xA800_0000 via DMA. In the emulator, this memory write fails unless CPTRA_EMULATOR_SS_MCI_OFFSET is set.

I removed set_var from the test and moved the default configuration into hw/model during emulator boot instead, so tests don't touch environment variables.

runtime/userspace/api/mcu-mbox-lib/src/cmd_interface.rs
arul-sri: Done! Removed the cfg(feature = "ocp-lock") check in cmd_interface.rs and raised the pool to 44 KiB across all builds.

runtime/userspace/api/caliptra-libapi/src/ocp_lock.rs
arul-sri: Added the TODO comment and tracking issue #2049 to calculate MAX_ENDORSEMENT_CERT_SIZE dynamically at build time.

common/mcu-mbox/src/messages.rs
arul-sri: Restored DpeSignerContextCert and GetOcpLockEndorsementCert back into enum McuMailboxResp and restored all match arms.

runtime/userspace/api/caliptra-api/src/dpe.rs
```diff
@@ -489,15 +523,49 @@ fn parse_derive_context_response(rsp: &[u8], rsp_len: usize) -> McuResult<DpeDer
...
+        const MCI_BASE_AXI_ADDRESS: u32 = 0xA800_0000;
+        const MCU_SRAM_AXI_OFFSET: u32 = 0x00C0_0000;
+        const MCU_SRAM_AXI_BASE: u32 = MCI_BASE_AXI_ADDRESS + MCU_SRAM_AXI_OFFSET;
+
+        let sram_offset = (rsp.as_ptr() as u32)
+            .checked_sub(MCU_SRAM_LOCAL_BASE)
+            .ok_or(INVARIANT)?;
+        let axi_addr = MCU_SRAM_AXI_BASE
+            .checked_add(sram_offset)
+            .ok_or(INVARIANT)?;
```
clundin25, line 548: In a follow up PR, we should consider wrapping this with a helper function or abstraction

clundin25 (approved):

PR #2006, opened 2026-08-20
Update caliptra-sw to latest caliptra-2.0 revision

parvathib:
## Summary
- update all synchronized `caliptra-sw` pins from `85981e1bc28662a9a9fa5040cbb38ac46e548217` to the latest `caliptra-2.0` commit, `e5ceee831f1b6b49a8d139a705133e47deab8711`
- regenerate `Cargo.lock` only for the Caliptra revision and its required transitive dependencies
- adapt the MCU builder call sites to the revised `FwId` API

## Validation
- `cargo metadata --locked --no-deps --format-version 1`
- `cargo check --locked -p caliptra-mcu-builder -p caliptra-mcu-hw-model`
- `cargo fmt --all -- --check`

### DIFF
--- a/builder/src/firmware.rs
+++ b/builder/src/firmware.rs
@@ -1,6 +1,6 @@
 // Licensed under the Apache-2.0 license
 
-use caliptra_builder::FwId;
+use caliptra_builder::{FirmwareType, FwId};
 
 pub mod hw_model_tests {
     use super::*;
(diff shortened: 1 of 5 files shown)


builder/src/firmware.rs
```diff
@@ -1,50 +1,50 @@
...
+use caliptra_builder::{FirmwareType, FwId};
 
 pub mod hw_model_tests {
     use super::*;
 
     pub const MAILBOX_RESPONDER: FwId = FwId {
         crate_name: "caliptra-mcu-test-fw-mailbox-responder",
         bin_name: "caliptra-mcu-test-fw-mailbox-responder",
-        features: &["emu"],
+        fw_type: FirmwareType::Source { features: &["emu"] },
```
mhatrevi, line 11: Why was this change needed?

parvathib: This is due to https://github.com/chipsalliance/caliptra-sw/pull/3999, which replaced  `FwId` 's  `features`  field with  `fw_type: FirmwareType` , enabling firmware to be fetched prebuilt.  `features` now applies only to the variant that builds from source. Without this, we get build errors.

mhatrevi (approved):

PR #1962, opened 2026-08-12
ci(spdm): gate feature/spdm-1_4 pull requests on the SPDM suites

parvathib:
## Summary

- Gate `feature/spdm-1_4` PRs with MCTP conformance, MCTP PCR quote, DOE conformance, and DOE TDISP/IDE; keep nightly coverage unchanged and cancel superseded runs.
- Use the `caliptra-main-14` SPDM 1.4 stack, including its `CHALLENGE_AUTH` fix.
- Make PCR-quote CI use compatible requester capabilities and report requester failures directly.
- New Validator cases still require a future spdm-emu submodule bump.

PR #2009, opened 2026-08-21
backport feature/spdm-1_4 commits to main 2.1

parvathib:
Backports SPDM validator gating from #1962 and ML-DSA-87 IDevID certificate installation by @RaunakGu from #1933 to  main-2.1.

### DIFF
--- a/.github/workflows/spdm-validator.yml
+++ b/.github/workflows/spdm-validator.yml
@@ -6,6 +6,16 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
   workflow_call:
+  # Gate pull requests targeting main or main-2.1 with the SPDM suites instead
+  # of only running them in the nightly job. Other branches are deliberately
+  # excluded: these tests build spdm-emu from source and run four emulator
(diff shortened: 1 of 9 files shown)


.github/workflows/spdm-validator.yml
```diff
@@ -6,6 +6,16 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
   workflow_call:
+  # Gate pull requests targeting main or main-2.1 with the SPDM suites instead
+  # of only running them in the nightly job. Other branches are deliberately
+  # excluded: these tests build spdm-emu from source and run four emulator
```
amit-kumarh, line 11: Is this still a concern here that we slow down CI for main / main-2.1 ?

amit-kumarh (approved):

.github/workflows/spdm-validator.yml
parvathib: Given that the job runs in parallel and isn't among the longer-running tests, the additional cost should be fairly minimal. The benefit is catching issues sooner and reducing the risk of regressions. We can revert it once we complete 1.4 support.

Issue #2010, opened 2026-08-24
feat: add set_dpe_only_slot() API for DPE-only SPDM certificate chains


yujieh-mrvl:
## Summary

The SPDM responder assembles `GET_CERTIFICATE` responses as `[endorsement] + [DPE chain] + [leaf cert]`. Today the only way to configure the endorsement segment is `set_endorsement_chain()`, which requires a non-empty static cert slice (Root CA + IDevID) baked into the firmware image.

For deployments where the verifier obtains the trust anchor out-of-band (e.g., platform registry), those static certs are unnecessary. The chain should start directly with the first DPE cert (LDevID), yielding `RootHash = SHA-384(LDevID)` and zero endorsement bytes.

## Proposed change

**`pal/src/cert/store.rs`** — add one public method:

```rust
pub async fn set_dpe_only_slot<A: ApiAlloc>(
    &self, alloc: &A, idx: usize, first_dpe_cert: &[u8], key_pair_id: u8,
) -> McuResult<()>
```

- Computes `RootHash = SHA-384(first_dpe_cert)`
- Writes `ReadOnlyEndorsement::new(&[], hash)` (empty chain, valid hash) into the slot
- Reuses existing types — no new enum variants, no struct layout changes

## Usage

Platform integrators choose one of two initialization paths per slot:

```rust
// Option A: legacy — static endorsement certs in firmware image
store.set_endorsement_chain(alloc, idx, &[ROOT_CA, IDEVID], root_hash, key_pair_id).await?;

// Option B: DPE-only — no static certs, trust anchor = first DPE cert
store.set_dpe_only_slot(alloc, idx, &first_dpe_cert_bytes, key_pair_id).await?;
```

Both APIs coexist; no feature flags or conditional compilation required.

## Resulting chain comparison

| Path | Chain content | RootHash |
|------|--------------|----------|
| `set_endorsement_chain` | `RootCA ‖ IDevID ‖ LDevID ‖ FMC ‖ RT ‖ Leaf` | `SHA-384(RootCA)` |
| `set_dpe_only_slot` | `LDevID ‖ FMC ‖ RT ‖ Leaf` | `SHA-384(LDevID)` |

## Trust model

[...]

PR #2011, opened 2026-08-24
feat(spdm-pal): add set_dpe_only_slot() for DPE-only certificate chains

yujieh-mrvl:
Add a new `set_dpe_only_slot()` method to `SharedCertStore` that configures a slot with an empty endorsement chain and `RootHash = SHA-384(first DPE cert)`. This enables SPDM certificate chains starting directly from LDevID without static Root CA or IDevID certs, for deployments where the verifier obtains trust anchors out-of-band. Unit tests included.

Related: #2010

### DIFF
--- a/runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
+++ b/runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
@@ -102,6 +102,47 @@ impl SharedCertStore {
         Ok(())
     }
 
+    /// Configure a DPE-only slot with no static endorsement certs.
+    ///
+    /// The SPDM certificate chain contains only the DPE device chain
+    /// (LDevID → FMC → RT) plus a CertifyKey leaf — no Root CA or IDevID
+    /// in the endorsement segment.
+    ///
+    /// The `*_first_dpe_cert` bytes (typically the LDevID cert) are used
+    /// **only** to compute the RootHash for each algorithm's SPDM
+    /// CertificateChain header; they are not stored in the endorsement
+    /// segment. Pass `mldsa_first_dpe_cert = None` on platforms that do not
+    /// serve an ML-DSA-87 chain for this slot.
+    pub async fn set_dpe_only_slot<A: ApiAlloc>(
+        &self,
+        alloc: &A,
+        idx: usize,
+        ecc_first_dpe_cert: &[u8],
+        mldsa_first_dpe_cert: Option<&[u8]>,
+        key_pair_id: u8,
+    ) -> McuResult<()> {
+        if idx >= NUM_CERT_SLOTS || ecc_first_dpe_cert.is_empty() {
+            return Err(mcu_error::codes::INVARIANT);
+        }
+        static EMPTY_CHAIN: &[&[u8]] = &[];
+
+        let ecc_hash = compute_root_hash(alloc, ecc_first_dpe_cert).await?;
+        let mut ro = ReadOnlyEndorsement::new(EMPTY_CHAIN, ecc_hash);
+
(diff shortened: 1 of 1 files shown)


yujieh-mrvl:
@amit-kumarh Could you please review this PR when you have a chance? Thanks!

amit-kumarh (commented):
This should be rebased on top of #2061 which adds postquantum certificate chain support. Please run `cargo fmt` as well to satisfy the precheckin. Otherwise, LGTM. Thanks!

runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
```diff
@@ -102,6 +102,47 @@ impl SharedCertStore {
...
+        if idx >= NUM_CERT_SLOTS || ecc_first_dpe_cert.is_empty() {
+            return Err(mcu_error::codes::INVARIANT);
+        }
+        static EMPTY_CHAIN: &[&[u8]] = &[];
+
+        let ecc_hash = compute_root_hash(alloc, ecc_first_dpe_cert).await?;
+        let mut ro = ReadOnlyEndorsement::new(EMPTY_CHAIN, ecc_hash);
+
+        if let Some(cert) = mldsa_first_dpe_cert {
+            if !cert.is_empty() {
```
amit-kumarh, line 133: nit: We could consider making this noisier, to catch a platform accidentally passing an empty buffer as an MLDSA cert instead of silently falling back to ECC only

amit-kumarh (approved):
LGTM. Can you please authorize the CLA as well to allow merging this? @yujieh-mrvl

runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
yujieh-mrvl: Great catch, thanks! You're right — silently falling back to ECC-only means a platform that fumbles its ML-DSA cert (e.g. passes an uninitialized/zero-length buffer) would ship a slot that quietly only serves ECC, which is exactly the kind of bug that's painful to notice in the field.

**I'll make the contract explicit and fail loudly:**

- `None` → the platform intentionally opts out of an ML-DSA chain (the only valid way to "skip")
- `Some(non-empty)` → serve ML-DSA
- `Some(empty)` → treated as a programming error, returns `INVARIANT` instead of degrading to ECC-only

Concretely, I'm pulling the input checks into a small **synchronous** helper called at the very top of `set_dpe_only_slot()`, before any allocation/hashing. This fails fast and makes it unit-testable without an async executor (this crate has no `block_on`/executor dev-dependency, and the `Some(empty)` path would otherwise sit behind the ECC `compute_root_hash().await`):

```rust
fn validate_dpe_only_inputs(
    idx: usize,
    ecc_first_dpe_cert: &[u8],
    mldsa_first_dpe_cert: Option<&[u8]>,
) -> McuResult<()> {
    if idx >= NUM_CERT_SLOTS || ecc_first_dpe_cert.is_empty() {
        return Err(mcu_error::codes::INVARIANT);
    }
    if matches!(mldsa_first_dpe_cert, Some(c) if c.is_empty()) {
        return Err(mcu_error::codes::INVARIANT);
    }
    Ok(())
}
```

`set_dpe_only_slot()` then calls `validate_dpe_only_inputs(...)?` first, and the ML-DSA branch drops its inner emptiness check (non-empty is now guaranteed).

I'll also:
- update the doc comment to state the `Some(empty)` → `INVARIANT` contract
- add a test pinning it: `Some(empty)` → `Err`; `None` / `Some(non-empty)` → `Ok` (plus the existing empty-ECC / out-of-range cases)
Does this approach look good to you? I'll push the change once you confirm.

PR #2014, opened 2026-08-24
fix: align SPDM VDM debug unlock with common format

### LINKED ISSUE #1748: Debug unlock out of band SPDM VDM testing issues
Hi there,
I am testing debug unlock SPDM VDM by using out-of-band solution with MCU emulator. With the latest main-2.1 source code at commit 36aee811c5a75dc8f5f598141e07948e575e389c,  I observed below 3 issues. Please help to check. Thanks.

1. SPDM/MCTP Standard ID mismatch between source code and documentation.
in the source code caliptra-util-host/transport/src/transports/spdm_vdm/protocol.rs is 0x4,
/// SPDM/MCTP StandardsBodyId for IANA-registered vendors.
pub const REGISTRY_ID_IANA: u16 = 0x04;
but in docs https://github.com/chipsalliance/caliptra-mcu-sw/blob/main-2.1/docs/src/caliptra_spdm_vdm_cmds.md in "OCP VDM Header" table "Standard ID" field, the value of Standard ID	2 bytes is 0x0005(IANA Enterprise ID format)

2. For "Request Debug Unlock" command "Request Payload:" according to documentation https://github.com/chipsalliance/caliptra-mcu-sw/blob/main-2.1/docs/src/caliptra_common_commands.md, the "unlock_level" field is at payload offset 4, after the 4-byte length_dwords=2. 
But the target source code function handle_request_debug_unlock() it reads "unlock_level" from the first byte of input buffer — exactly the length_dwords low byte. 
So the target isn't expecting that 4-byte length prefix. Please confirm which one is the correct? For "Authorize Debug Unlock Token", the length is properly handled, seems host tool should follow documentation.

3. For "Authorize Debug Unlock Token" command, I observed Caliptra core reports mailbox checksum error, MCU seems doesn't add correct checksum when calling Caliptra Mailbox command.
I will need to apply below patch to make it working.

    Fix debug-unlock token mailbox checksum in emulator VDM hook

    authorize_debug_unlock_token previously passed None as the mailbox
    header, assuming the host had already prepended a MailboxReqHeader
    checksum to the token payload. The host actually streams the raw VDM
    payload (length_dwords at [0:3]) with no checksum, so Caliptra RT
    rejected the request on a checksum mismatch.

    Compute the Caliptra mailbox checksum over cmd_id + token_data and
    prepend it as the 4-byte header before streaming, so Caliptra RT
    receives [checksum(4)][token_data].

    Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

diff --git a/platforms/emulator/runtime/userspace/apps/user/src/spdm/caliptra_vdm.rs b/platforms/emulator/runtime/userspace/apps/user/src/spdm/caliptra_vdm.rs
index 112f7a72..98edd61d 100644
--- a/platforms/emulator/runtime/userspace/apps/user/src/spdm/caliptra_vdm.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/spdm/caliptra_vdm.rs
@@ -110,17 +110,30 @@ impl CaliptraVdmCommands for CaliptraVdmHook {
         scratch: &A,
     ) -> CaliptraVdmResult<()> {
         let cmd = PRODUCTION_AUTH_DEBUG_UNLOCK_TOKEN_CMD;
-        // The host sends the production debug-unlock token payload in Caliptra RT
-        // mailbox format, including the MailboxReqHeader/checksum, and large
-        // requests are streamed directly to the mailbox. Do not synthesize
-        // another checksum header here or the mailbox would receive
-        // `[new_checksum || host_checksum || token]`.
+
+        // Compute Caliptra mailbox checksum: 0 - (sum_of_cmd_id_LE + sum_of_token_data).
+        // The host sends the raw VDM payload (length_dwords at [0:3]); we prepend
+        // the checksum header so Caliptra RT sees [checksum(4)][token_data].
+        let byte_sum: u32 = cmd
+            .to_le_bytes()
+            .iter()
+            .chain(token_data.iter())
+            .map(|&b| b as u32)
+            .sum();
+        let checksum = 0u32.wrapping_sub(byte_sum);
+        let checksum_header = checksum.to_le_bytes();
+
         let mut token_stream = SlicePayloadStream::new(token_data);
         let mut resp_buf = ApiAlloc::alloc(scratch, PRODUCTION_AUTH_DEBUG_UNLOCK_TOKEN_RSP_LEN)
             .map_err(map_mcu_err)?;

         Mailbox::<DefaultSyscalls>::new()
-            .execute_with_payload_stream(cmd, None, &mut token_stream, &mut resp_buf)
+            .execute_with_payload_stream(
+                cmd,
+                Some(&checksum_header),
+                &mut token_stream,
+                &mut resp_buf,
+            )
             .await
             .map_err(map_mailbox_error)?;
         Ok(())

mlvisaya:
Implemented SPDM VDM command format for Request Debug unlock does not conform with the format defined in the Caliptra Common Commands Spec.

The SPDM implementation intentionally accepts exactly one byte, unlock_level, and rejects trailing bytes. However, the transport-agnostic documentation still specifies [length(4), unlock_level, reserved(3)].

The MCU Mailbox follows the format defined in spec.

This change is to make the format used by the SPDM VDM to be the same as the MCU Mailbox one as defined in the Caliptra Common Commands Spec.

Fixes https://github.com/chipsalliance/caliptra-mcu-sw/issues/1748

### DIFF
--- a/caliptra-util-host/Cargo.lock
+++ b/caliptra-util-host/Cargo.lock
@@ -441,7 +441,7 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 [[package]]
 name = "caliptra-api-types"
 version = "0.1.0"
-source = "git+https://github.com/chipsalliance/caliptra-sw?rev=85981e1bc28662a9a9fa5040cbb38ac46e548217#85981e1bc28662a9a9fa5040cbb38ac46e548217"
+source = "git+https://github.com/chipsalliance/caliptra-sw?rev=e5ceee831f1b6b49a8d139a705133e47deab8711#e5ceee831f1b6b49a8d139a705133e47deab8711"
 dependencies = [
  "caliptra-image-types",
 ]
@@ -449,7 +449,7 @@ dependencies = [
 [[package]]
 name = "caliptra-cfi-derive"
 version = "0.1.0"
-source = "git+https://github.com/chipsalliance/caliptra-sw?rev=85981e1bc28662a9a9fa5040cbb38ac46e548217#85981e1bc28662a9a9fa5040cbb38ac46e548217"
+source = "git+https://github.com/chipsalliance/caliptra-sw?rev=e5ceee831f1b6b49a8d139a705133e47deab8711#e5ceee831f1b6b49a8d139a705133e47deab8711"
 dependencies = [
  "paste",
  "proc-macro2",
@@ -460,7 +460,7 @@ dependencies = [
 [[package]]
 name = "caliptra-cfi-lib"
 version = "0.1.0"
-source = "git+https://github.com/chipsalliance/caliptra-sw?rev=85981e1bc28662a9a9fa5040cbb38ac46e548217#85981e1bc28662a9a9fa5040cbb38ac46e548217"
+source = "git+https://github.com/chipsalliance/caliptra-sw?rev=e5ceee831f1b6b49a8d139a705133e47deab8711#e5ceee831f1b6b49a8d139a705133e47deab8711"
 dependencies = [
  "caliptra-error",
  "caliptra-registers",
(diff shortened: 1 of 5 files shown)


helloxiling (approved):
Looks good to me.

mlvisaya:
thanks

PR #2019, opened 2026-08-25
[spdm] update spdm-lib CERTIFICATE to support large offset/length 

### LINKED ISSUE #1794: SPDM_1.4: GET_CERTIFICATE / CERTIFICATE - Runtime support

amit-kumarh:
New fields were added in SPDM 1.4 to support 32 bit length and offset fields in the GET_CERTIFICATE commands. Update the codec and stack modules to support these fields

### DIFF
--- a/runtime/userspace/api/spdm-lib/codec/src/certificate.rs
+++ b/runtime/userspace/api/spdm-lib/codec/src/certificate.rs
@@ -27,13 +27,43 @@
 //! `CertChain` carries `PortionLength` bytes of the SPDM cert-chain
 //! wire format (length(2) | reserved(2) | root_hash(48) | DER).
 
-use zerocopy::{little_endian::U16, FromBytes, Immutable, IntoBytes, KnownLayout, Unaligned};
+use zerocopy::{
+    little_endian::{U16, U32},
+    FromBytes, Immutable, IntoBytes, KnownLayout, Unaligned,
+};
 
 use crate::{ReqRespCode, ResponseBody, WireError, WireWriter};
+use bitfield_struct::bitfield;
+
+/// `Param1` field of GET_CERTIFICATE / CERTIFICATE
+/// Bits 0..=3 carry the `SlotID`, and bit 7 is the `LargeCertChain` flag (SPDM 1.4).
+#[bitfield(u8)]
+#[derive(FromBytes, IntoBytes, KnownLayout, Immutable, Unaligned, PartialEq, Eq)]
+pub struct GetCertificateParam1 {
+    #[bits(4)]
+    pub slot_id: u8,
+    #[bits(3)]
+    _reserved: u8,
+    pub large_cert_chain: bool,
+}
+
+const PARAM1_SLOT_ID_MASK: u8 = 0xF;
 
 /// 6-byte GET_CERTIFICATE request body (after the 2-byte SPDM
 /// common header).
-#[derive(FromBytes, IntoBytes, KnownLayout, Immutable, Unaligned, Copy, Clone, Debug, Default)]
+#[derive(
+    FromBytes,
+    IntoBytes,
+    KnownLayout,
+    Immutable,
+    Unaligned,
+    Copy,
+    Clone,
+    Debug,
+    Default,
+    PartialEq,
+    Eq,
+)]
 #[repr(C)]
 pub struct GetCertificateReqBody {
     /// `Param1` — `SlotID` in bits 0..=3.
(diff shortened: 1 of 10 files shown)


runtime/userspace/api/spdm-lib/codec/src/certificate.rs
```diff
@@ -52,13 +76,153 @@ impl GetCertificateReqBody {
...
+    Clone,
+    Debug,
+    Default,
+    PartialEq,
+    Eq,
+)]
+#[repr(C)]
+pub struct GetCertificateLargeReqBody {
+    /// `Param1` — `SlotID` in bits 0..=3, `LargeCertChain` in bit 7.
+    pub slot_id: u8,
```
clundin25: Can we use a flag type defined by the `bit_flags` crate? That lead to higher level / idiomatic code

clundin25 (approved):

runtime/userspace/api/spdm-lib/codec/src/certificate.rs
```diff
@@ -27,13 +27,37 @@
...
-use zerocopy::{little_endian::U16, FromBytes, Immutable, IntoBytes, KnownLayout, Unaligned};
+use zerocopy::{
+    little_endian::{U16, U32},
+    FromBytes, Immutable, IntoBytes, KnownLayout, Unaligned,
+};
 
 use crate::{ReqRespCode, ResponseBody, WireError, WireWriter};
 
+/// `Param1` bit 7 — LargeCertChain (SPDM 1.4, DSP0274 §10.8).
+pub const REQ_PARAM1_LARGE_CERT_CHAIN: u8 = 0x80;
```
zhalvorsen: It looks like other SPDM files are using `def_flag_set_le` for definitions of constants like this. Would that be applicable here as well?

runtime/userspace/api/spdm-lib/codec/src/certificate.rs
```diff
@@ -52,13 +76,153 @@ impl GetCertificateReqBody {
...
+        } else {
+            CertificateRspBody::SIZE
+        }
+    }
+
+    /// Maximum cert chain length that can be represented in this request mode.
+    #[inline]
+    pub fn max_length_cap(&self) -> usize {
+        if self.is_large() {
+            usize::MAX
```
zhalvorsen: Should this be `u32::MAX` to make it more explicit because `usize` is architecture specific?

runtime/userspace/api/spdm-lib/codec/src/certificate.rs
```diff
@@ -52,13 +76,153 @@ impl GetCertificateReqBody {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum GetCertificateReq<'a> {
+    Standard(&'a GetCertificateReqBody),
+    Large(&'a GetCertificateLargeReqBody),
+}
+
+impl<'a> GetCertificateReq<'a> {
+    /// Parse a GET_CERTIFICATE request body (excluding the 2-byte common SPDM header).
+    pub fn parse(body: &'a [u8]) -> Result<Self, WireError> {
+        let param1 = body.first().copied().ok_or(WireError)?;
```
zhalvorsen: The spec includes a few checks that we should do here.

1. If not large cert chain, check that `large_offset` and `large_length` are both absent so it should probably check the body size
2. If large cert chain, check that `offset` and `length` are both 0

runtime/userspace/api/spdm-lib/stack/src/certificate.rs
```diff
@@ -84,22 +99,38 @@ impl CertificateLargeResponse {
...
+                            self.remainder_len,
+                        ),
+                    })
+                    .map_err(|_| mcu_error::codes::INVARIANT)?;
+            } else {
+                writer
+                    .write(&CertificateRspBody {
+                        slot_id: self.slot_id,
+                        param2: self.param2,
+                        portion_length: U16::new(self.portion_len as u16),
```
zhalvorsen, line 130: Should we take this comment in the Spec into account here and for `remainder_length` instead of downcasting to u16?

> If a Requester uses Offset and Length fields although the certificate chain size cannot be represented by these fields, the Responder shall send an ERROR message with ErrorCode=DataTooLarge .

runtime/userspace/api/spdm-lib/stack/src/certificate.rs
```diff
@@ -182,35 +211,37 @@ pub(crate) async fn handle_get_certificate_req<'a, Pal: SpdmPal>(
     let total_len_usize = SPDM_CERT_CHAIN_HDR_LEN
         .checked_add(der_len)
         .ok_or(SPDM_UNSPECIFIED)?;
-    let total_len = u16::try_from(total_len_usize).map_err(|_| SPDM_UNSPECIFIED)?;
+
+    if total_len_usize > req.max_length_cap() {
+        return Err(SPDM_UNSPECIFIED);
```
zhalvorsen: Should this error be DataTooLarge? Maybe this is the correct location to handle the error mention in my previous DataTooLarge comment?

amit-kumarh: Yep, that error should be handled here. changed to be a DataTooLarge.

runtime/userspace/api/spdm-lib/stack/src/certificate.rs
amit-kumarh: The error check from the below comment takes place first, so by the time we get here, we know this is a safe downcast.

runtime/userspace/api/spdm-lib/codec/src/certificate.rs
amit-kumarh: good catch. done.

runtime/userspace/api/spdm-lib/codec/src/certificate.rs
amit-kumarh: addressed by reply to Carl's comment.

runtime/userspace/api/spdm-lib/codec/src/certificate.rs
amit-kumarh: migrated to the bitfields crate for this, which we can use because this is a u8, and therefore unaligned.

zhalvorsen (approved):

zhalvorsen (approved):

parvathib:
@amit-kumarh  Sorry I wasn't able to review this sooner. Thanks for adding the CERTS support. 

Some follow-ups: it looks like the PQC certs aren't actually getting exercised. The certificate cases that do run (5.1, 5.2, 5.4) all go against the classical ECC P-384 chain - there's no ML-DSA anywhere in the run. Separately, 5.6–5.9 come back skipped  with  `pass: 0, fail: 0` , so the large-field path isn't covered either.

1. **Algorithm selection**  MCU responder apparently hardcodes `pqc_asym_sel: PqcAsymAlgos::EMPTY`. We need to actually intersect the requester's  `PQCAsymAlgo` offer and select either ECC/MLDSA
2. **Cert store to serve MLDSA-87 cert chain** Although the  `SpdmPalAsymAlgo`  plumbing is available,  `asym_algo()`  still hardcodes ECC and the PAL still ignores the algo  parameter. Were you able to successfully test the ML-DSA cert chain end to end?
3. **LARGE_RESP_CAP advertise**. The cases you added aren't registered in the certificate test group yet, so they're currently skipped. Once they're enabled, what cert chain size will they actually exercise? If it's well under **64 KiB**, the standard  `Offset / Length`  fields are sufficient, there's no reason to advertise the cap, and the cases would just stay skipped anyway since they gate on it. If the chain does cross 64 KiB, then advertising is worth doing and the cases become meaningful.

amit-kumarh:
@parvathib 

Yes - the rest of the logic to store and serve the MLDSA cert chain is in https://github.com/chipsalliance/caliptra-mcu-sw/pull/2061 - this change was just ready first so I sent it out for review.

For the capability advertisement - it turns out LARGE_RESP_CAP also enables some large response fields in the vendor-defined commands. This was causing TDISP test failures in CI. I wanted to unblock Arthur's work on the next commands ASAP, so I'm not advertising the capability for now, and will circle back to it once the remaining certificate code is in.

The slot 0 cert chain is about 48-49k right now, so it will fit in the regular Offset/Length fields, but I think it's a good idea to support the large responses as the owner/tenant PQC chain could be longer?

parvathib:
> @parvathib
> 
> Yes - the rest of the logic to store and serve the MLDSA cert chain is in #2061 - this change was just ready first so I sent it out for review.
> 
> For the capability advertisement - it turns out LARGE_RESP_CAP also enables some large response fields in the vendor-defined commands. This was causing TDISP test failures in CI. I wanted to unblock Arthur's work on the next commands ASAP, so I'm not advertising the capability for now, and will circle back to it once the remaining certificate code is in.
> 
> The slot 0 cert chain is about 48-49k right now, so it will fit in the regular Offset/Length fields, but I think it's a good idea to support the large responses as the owner/tenant PQC chain could be longer?

Thanks for clarifying. I'll review #2061. I agree that the endorsement certificate chain can be longer, and the DPE leaf certificate may also grow in size. Ideally, we should support LARGE_RESP_CAP. However, since it isn't blocking anything right now, I'm okay enabling it later once there are no dependencies on it.

PR #2027, opened 2026-08-26
emulator/caliptra: gate the REQ_IDEVID_CSR clear on subsystem mode

chrisherczeg:
### Problem

`download_idev_id_csr` is registered as the IDEVID CSR download callback. It
stands in for the SoC: it drains the mailbox, writes the CSR out, completes the
command, and then unconditionally clears the request bit:

```rust
soc_mbox.status().write(|w| w.status(|w| w.cmd_complete()));

// Clear the Idevid CSR requested bit.
cptra_dbg_manuf_service_reg.modify(DebugManufService::REQ_IDEVID_CSR::CLEAR);
```

In **subsystem** mode that last step is wrong, because the MCU owns the
handshake there: it polls `CPTRA_FLOW_STATUS.idevid_csr_ready` and clears
`REQ_IDEVID_CSR` itself once it has consumed the CSR.

Clearing the bit from the callback races the ROM's flow-status update. If the
callback wins, the ready indication is torn down before the MCU ever observes
it, and the MCU polling loop never sees `idevid_csr_ready` assert.

In **standalone** mode the callback genuinely is the SoC, so clearing there is
correct and must be preserved.

### Fix

Thread the existing `subsystem_mode` value into the callback and gate the clear
on it. Standalone behaviour is byte-for-byte unchanged; subsystem mode now
leaves the request bit for the MCU.

The literal `true` that was previously passed inline to
`CaliptraRootBusArgs::subsystem_mode` is hoisted into a named local so both
uses stay in sync.

### Validation

- `cargo check -p caliptra-mcu-emulator-caliptra`
- `cargo fmt --check`

### DIFF
--- a/emulator/caliptra/src/caliptra.rs
+++ b/emulator/caliptra/src/caliptra.rs
@@ -164,6 +164,9 @@ pub fn start_caliptra(
     // in active mode, we don't update firmware here, as MCU will trigger it
     let upload_update_fw = UploadUpdateFwCb::new(|_| {});
 
+    // In subsystem mode the MCU owns clearing REQ_IDEVID_CSR.
+    let subsystem_mode = true;
+
     let bus_args = CaliptraRootBusArgs {
         clock: clock.clone(),
         pic: pic.clone(),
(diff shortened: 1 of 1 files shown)


chrisherczeg:

PR #2032, opened 2026-08-26
emulator/caliptra: gate the REQ_IDEVID_CSR clear on subsystem mode

chrisherczeg:
### Problem

`download_idev_id_csr` is registered as the IDEVID CSR download callback. It
stands in for the SoC: it drains the mailbox, writes the CSR out, completes the
command, and then unconditionally clears the request bit:

```rust
soc_mbox.status().write(|w| w.status(|w| w.cmd_complete()));

// Clear the Idevid CSR requested bit.
cptra_dbg_manuf_service_reg.modify(DebugManufService::REQ_IDEVID_CSR::CLEAR);
```

In **subsystem** mode that last step is wrong, because the MCU owns the
handshake there: it polls `CPTRA_FLOW_STATUS.idevid_csr_ready` and clears
`REQ_IDEVID_CSR` itself once it has consumed the CSR.

Clearing the bit from the callback races the ROM's flow-status update. If the
callback wins, the ready indication is torn down before the MCU ever observes
it, and the MCU polling loop never sees `idevid_csr_ready` assert.

In **standalone** mode the callback genuinely is the SoC, so clearing there is
correct and must be preserved.

### Fix

Thread the existing `subsystem_mode` value into the callback and gate the clear
on it. Standalone behaviour is byte-for-byte unchanged; subsystem mode now
leaves the request bit for the MCU.

The literal `true` that was previously passed inline to
`CaliptraRootBusArgs::subsystem_mode` is hoisted into a named local so both
uses stay in sync.

### Validation

- `cargo check -p caliptra-mcu-emulator-caliptra`
- `cargo fmt --check`

### DIFF
--- a/emulator/caliptra/src/caliptra.rs
+++ b/emulator/caliptra/src/caliptra.rs
@@ -164,6 +164,9 @@ pub fn start_caliptra(
     // in active mode, we don't update firmware here, as MCU will trigger it
     let upload_update_fw = UploadUpdateFwCb::new(|_| {});
 
+    // In subsystem mode the MCU owns clearing REQ_IDEVID_CSR.
+    let subsystem_mode = true;
+
     let bus_args = CaliptraRootBusArgs {
         clock: clock.clone(),
         pic: pic.clone(),
@@ -183,10 +186,15 @@ pub fn start_caliptra(
                 u32,
                 DebugManufService::Register,
             >| {
-                download_idev_id_csr(mailbox, log_dir.clone(), cptra_dbg_manuf_service_reg);
+                download_idev_id_csr(
+                    mailbox,
+                    log_dir.clone(),
+                    cptra_dbg_manuf_service_reg,
+                    subsystem_mode,
+                );
             },
         ),
-        subsystem_mode: true,
+        subsystem_mode,
         use_mcu_recovery_interface: args_use_mcu_recovery_interface,
         debug_intent: args.debug_intent,
         prod_dbg_unlock_keypairs: args.prod_dbg_unlock_keypairs.clone(),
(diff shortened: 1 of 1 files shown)


chrisherczeg:
Converting back to draft — CI found a real regression, and the premise of this change is wrong for upstream.

## What failed

`test_emulator` shards 1, 2 and 4: `test_streaming_soc_boot_manifest_good_svn` and
`test_streaming_soc_boot_manifest_svn_lt_fuse` hang and are killed at ~750s. The same
tests pass in ~17s and ~6s on an unrelated branch cut from the same base, at the same
nextest shard indices, so this is not flake.

## Root cause

I gated the `REQ_IDEVID_CSR` clear on `subsystem_mode` on the assumption that some other
agent clears the bit in subsystem mode. That is not true here:

- `download_idev_id_csr` has exactly one call site.
- `CaliptraRootBusArgs::subsystem_mode` is hardcoded `true` at that site.
- No other code in the tree references `REQ_IDEVID_CSR`, and nothing waits on
  `idevid_csr_ready`.

So gating on `subsystem_mode` does not preserve a standalone path — there is no standalone
path — it removes the clear unconditionally in every configuration. My claim that
"standalone behaviour is byte-for-byte unchanged" is vacuous.

With nothing clearing the request bit, Caliptra ROM never advances past the IDEVID CSR
handshake. The logs show it reaching `[idev] MLDSA Sign CSR w/ SUBJECT.KEYID = 8` and the
MCU then stalling forever at `Waiting for Caliptra to be ready for mbox`.

The race this patch describes is real, but it only manifests where the MCU-side clear
exists. Making the emulator depend on a clear that no code here performs is a
straightforward deadlock.

## Path forward

The clear has to stay on by default. I will re-propose as an opt-in `EmulatorArgs` flag that
defaults to today's behaviour, so the race can be avoided by configurations that do clear the
bit on the MCU side without changing the default. Happy to hear if maintainers would rather
see the MCU-side handshake modelled first instead.

PR #2033, opened 2026-08-26
emulator/lc_ctrl: reload lifecycle state from OTP on update reset

chrisherczeg:
### Problem

`LcCtrl` implements `warm_reset` by reloading its lifecycle state from OTP:

```rust
fn warm_reset(&mut self) {
    self.reload_from_otp();
}
```

but it has no `update_reset` implementation, so it falls through to the default
no-op. `UpdateReset` is the action the MCI schedules for warm resets, which
means a lifecycle state that was committed to OTP just before the reset is not
picked up: after the reset the controller keeps serving the *pre-transition*
state, and firmware reads a stale `LC_STATE`.

This is observable for any transition that is committed and then followed by a
reset — e.g. `MANUF` -> `SCRAP`, where firmware expects to boot into `SCRAP`
but still sees `MANUF`.

### Fix

Implement `update_reset` the same way as `warm_reset`. Both are reset entry
points into the same controller and both need the OTP-backed state re-read.

### Validation

- `cargo fmt --check`

### DIFF
--- a/emulator/periph/src/lc_ctrl.rs
+++ b/emulator/periph/src/lc_ctrl.rs
@@ -334,6 +334,14 @@ impl caliptra_mcu_emulator_registers_generated::lc::LcPeripheral for LcCtrl {
         self.reload_from_otp();
     }
 
+    fn update_reset(&mut self) {
+        // UpdateReset is the action the MCI schedules for warm resets.
+        // Reload lifecycle state from OTP so the new committed state
+        // (e.g. SCRAP after a MANUF→SCRAP transition) is visible to
+        // firmware on the next boot.
+        self.reload_from_otp();
+    }
+
     fn read_status(&mut self) -> ReadWriteRegister<u32, lc_ctrl::bits::Status::Register> {
         ReadWriteRegister::new(self.status.reg.get())
     }


chrisherczeg:
Converting back to draft — CI found a real regression, and it comes from the same
reset-domain assumption I thought I had removed when splitting this change.

## What failed

- `test_emulator` shard 2: `test_firmware_update_flash_missing_caliptra_image`,
  `test_streaming_soc_boot_one_component_id_for_all` and
  `test_streaming_soc_boot_successful_non_identical` all hit the 900s nextest timeout.
- `caliptra-sw-mcu-rom-smoke-test / test_artifacts_subsystem`:
  `test_activate_firmware::test_activate_mcu_fw_success` times out at 180s.

All four pass in 17–20s on an unrelated branch cut from the same base, at the same shard
indices, so this is not flake.

## Root cause

`TimerAction::UpdateReset` is scheduled from `mci.rs` under `MCI: MCU proceeding with
reboot` — the MCU firmware boot reset. It is an MCU-domain reset, not a subsystem reset.
In hardware the lifecycle controller is not in that reset domain, so it should not be
re-initialised there.

Beyond re-reading the OTP-backed state, `reload_from_otp()` also clears transient
controller state:

```rust
self.status = (STATUS_INITIALIZED | STATUS_READY).into();
self.mutex_claimed = false;
self.transition_target = 0;
self.token = [0; 4];
```

Every failing test goes through `RESET_REASON: 0x00000002` / `Firmware boot reset
detected` and then hangs at `[FW Upd] Start`, i.e. the LCC mutex and status are being
reset underneath an in-flight flow by a reset that should not touch the LCC at all.

I split this from a larger change and dropped a paired DCCM-clearing hunk on the grounds
that it encoded SoC-specific reset semantics. That reasoning was right, but I should have
applied it to the `update_reset` hook itself — both halves rest on the same assumption
about what an update reset resets.

## Open question for maintainers

Which reset domain is `LcPeripheral::update_reset` intended to represent? If it is only
ever driven by an MCU-domain reboot, then the stale-`LC_STATE`-after-transition problem
needs to be solved somewhere else, and I would rather ask than guess. If a narrower reload
(refresh `lc_state_index`/`lc_transition_cnt` only, leaving mutex/status/token alone) is
the shape you would accept, I am happy to send that instead.

PR #2035, opened 2026-08-26
spdm: keep serving when the IDevID cert is not provisioned

### LINKED ISSUE #2034: [SPDM] An unprovisioned IDevID cert should not disable SPDM: init failure kills every responder, and 0xFF OTP is spliced in silently
## Summary

A device with no IDevID certificate provisioned in external OTP should still run SPDM — degraded, not dead. Today it does neither cleanly: depending on how the OTP read behaves, the responders either **all die** or **silently serve a corrupt chain**.

Two problems, both on `main`.

## Problem 1 — one cert-store init failure kills *every* SPDM transport

`spdm_mctp_responder` (`spdm/mod.rs:344-351`) and `spdm_doe_responder` (`:418-425`):

```rust
if let Err(e) = ensure_cert_store_init(allocator).await {
    crate::log_error!(cw, "SPDM_MCTP: cert store init failed: 0x{}", ...);
    return;                     // task ends here
}
...
stack.run().await               // never reached - the only place SPDM requests are served
```

Because the state machine in `ensure_cert_store_init` is shared, a single failure takes down all transports at once:

- first caller sees `CERT_STORE_STATE == 0`, sets `1`, runs init, fails, resets to `0`, sends `false` (`spdm/mod.rs:207-215`)
- the other responders already observed state `1`, so they wait on `CERT_STORE_INIT` and receive `false` → `Err` → they `return` too

Net effect: **no response to `GET_VERSION` on any transport** — MCTP, DOE, and (with the same pattern) a mailbox responder. The device is indistinguishable from hung, when it could have answered everything that does not depend on slot 0.

## Problem 2 — unprovisioned OTP is spliced in as `0xFF`, and init reports success

`populate_idev_from_otp` (`spdm/cert_store/mod.rs:218-227`):

```rust
let mut cert_buf = [0u8; ECC_DEVID_CERT_SIZE];        // fixed 547
read_otp_range(&otp, OTP_IDEVID_ECC_PARTITION, &mut cert_buf).await?;
retry_on_mailbox_busy!(populate_idev_ecc384_cert(alloc, &cert_buf))
```

There is no DER validation. On an unprovisioned partition the read **succeeds** and yields 547 bytes of `0xFF`, which are sent to Caliptra. `POPULATE_IDEV_ECC384_CERT` does no parse, no key-match and no signature check, so the bytes are prepended to `drivers.ecc_cert_chain` verbatim. Cert-store init returns `Ok`, SPDM comes up, and the corruption only surfaces as a chain-validation failure at the remote verifier — with no local diagnostic.

Note the asymmetry: the ML-DSA path added later **does** guard this. `mldsa87_cert_der_len` (`cert_store/mod.rs:289`) derives the length from the cert's own DER header and skips the install when the partition does not hold a valid SEQUENCE. The ECC path has no equivalent.

## Why "unprovisioned should be allowed to run"

- Bring-up, pre-provisioning and lab parts legitimately have empty OTP. Losing SPDM entirely on those makes the failure much harder to diagnose than losing only slot-0 attestation.
- Integrators may not store the IDevID cert in external OTP at all (e.g. NAND-backed storage). Such a part has empty OTP `0x01` by design, and today gets Problem 2 silently.
- `GET_VERSION` / `GET_CAPABILITIES` / `NEGOTIATE_ALGORITHMS` do not depend on slot 0. Refusing to serve them buys no security — a verifier cannot be misled by a chain it was never handed. Failing the cert-dependent commands is the honest behaviour.

## Suggested fix

1. **Do not end the responder task on cert-store init failure** (`spdm/mod.rs:344-351`, `:418-425`). Log and continue into `stack.run()`. VCA still works; `GET_DIGESTS` / `GET_CERTIFICATE` / `CHALLENGE` fail for the affected slot.
2. **Add a DER-header guard to the ECC path**, mirroring the ML-DSA one: if partition `0x01` does not begin with a plausible `SEQUENCE`, log and skip the install rather than submitting `0xFF`. An empty slot 0 is then an explicit, visible state instead of a corrupt chain.
3. **Do not let one transport's failure propagate to the others**, or at minimum record which transport observed it. Today the `CERT_STORE_INIT` broadcast of `false` fans a single failure out to every waiter.
4. Optional: mark slot 0 unprovisioned in the cert store so `GET_DIGESTS` reports it via `ProvisionedSlotMask` instead of hashing garbage. That is the spec-shaped way to say "this slot has no chain".

## Notes

- Problems 1 and 2 are mutually exclusive per boot: a *successful* read of an erased partition gives Problem 2; a *failing* OTP read or mailbox error gives Problem 1.
- Same reasoning as the review feedback on the ML-DSA installer: a best-effort provisioning step must be distinguishable in the log from success, and must not take down the paths that do not depend on it.

RaunakGu:
Fixes #2034

An unprovisioned IDevID slot 0 currently has two bad outcomes.

**Cert-store init failure took down every SPDM transport.** Both responders returned before reaching `stack.run()`, the only place SPDM requests are handled, so nothing answered `GET_VERSION` on any path — the device looked hung rather than degraded. The failure is also shared: the first caller resets `CERT_STORE_STATE` and broadcasts `false`, so one failure fans out to every waiter. Now it logs and keeps serving.

Slot 0 is left genuinely unprovisioned in that case. `populate_idev` reports whether it installed a cert, and slot 0's endorsement is skipped when it did not — because `SlotEndorsement::ReadOnly` makes `is_provisioned()` return `true` (`spdm-lib/pal/src/cert/endorsement.rs:165`), so endorsing it anyway would advertise the slot as provisioned while serving a chain whose Root CA makes no statement about this device. Left `Empty`, the SPDM slot masks report it absent, cert-dependent commands fail with a clean error, and VCA still works.

**An erased OTP partition was installed as a certificate.** The ECC path submitted a fixed 547 bytes and nothing downstream parses the DER — not this code, and not Caliptra's `POPULATE_IDEV_ECC384_CERT` handler — so an erased partition reading back `0xFF` was spliced into the attestation chain while init reported success. It only surfaced as a chain-validation failure at the remote verifier.

Both paths now share `cert_der_len(header_word, partition_size, max_cert_size)` and submit the cert's own DER length, bounded by the partition and by Caliptra's per-algorithm cap. That also covers a *short* or partly-programmed cert, not just a fully erased partition. The ECC cert is staged in scratch rather than a 1 KiB stack buffer, matching the ML-DSA path. Length policy lives in `mcu-caliptra-api` so it is covered by host unit tests; user-app is excluded from `cargo test`.

### Testing
- `cargo test -p mcu-caliptra-api` — 38 pass, including new `ecc384_cert_der_len_cases` (cap, partition bound, erased-OTP)
- `cargo clippy --all-targets -- -D warnings` clean; `cargo check` clean for the riscv `spdm` build
- `test_mctp_spdm_responder_conformance` on the emulator — **659 pass / 0 fail**, unchanged from before this series

### Not covered
Every emulator fixture programs both OTP partitions unconditionally (`tests/integration/src/lib.rs:1389`), so no integration test exercises the unprovisioned path. Adding one needs a `TestParams` field to pass `None`; happy to do it here or as a follow-up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3733,7 +3733,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ccc2776f0c61eca1ca32528f85548abd1a4be8fb53d1b21c013e4f18da1e7090"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.109",
 ]
 
 [[package]]
--- a/platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/spdm/cert_store/mod.rs
@@ -10,9 +10,9 @@
 //!
 //! Caliptra generates the IDevID *keypair* and a self-signed CSR, but never the
 //! IDevID *certificate* — that is issued externally and provisioned into OTP.
-//! Both the ECC-384 (partition 1) and ML-DSA-87 (partition 2) IDevID certs are
-//! read here and prepended to their respective Caliptra cert chains via the
-//! `POPULATE_IDEV_*_CERT` mailbox commands.
+//! When provisioned, the ECC-384 (partition 1) and ML-DSA-87 (partition 2) IDevID
+//! certs are read here and prepended to their respective Caliptra cert chains via
+//! the `POPULATE_IDEV_*_CERT` mailbox commands.
 
 mod slot0_endorsements;
 
@@ -28,7 +28,7 @@ use caliptra_mcu_spdm_pal::cert::store::SharedCertStore;
 #[allow(unused_imports)]
 use core::fmt::Write as _;
 use mcu_caliptra_api::{
-    is_der_cert_header, mldsa87_cert_der_len, populate_idev_ecc384_cert,
+    ecc384_cert_der_len, mldsa87_cert_der_len, populate_idev_ecc384_cert,
     populate_idev_mldsa87_cert, ApiAlloc,
 };
 use mcu_error::McuResult;
@@ -143,12 +143,12 @@ async fn yield_now() {
     YieldNow(false).await
 }
 
-/// One-time Caliptra setup: read the IDevID certs from OTP and install them.
+/// One-time Caliptra setup: read the IDevID ECC-384 cert from OTP and install it.
 ///
-/// Only the ECC-384 install runs here; its failure fails cert-store init, which
-/// leaves slot 0 unprovisioned but no longer stops the responders from serving.
-/// The ML-DSA-87 install is best-effort and lives in its own task.
-pub async fn populate_idev<A: ApiAlloc>(alloc: &A) -> McuResult<()> {
+/// Returns `false` when partition 1 holds no cert, so the caller can leave slot 0
+/// unprovisioned rather than endorsing a chain with no IDevID in it. The
+/// ML-DSA-87 install is best-effort and lives in its own task.
+pub async fn populate_idev<A: ApiAlloc>(alloc: &A) -> McuResult<bool> {
     populate_idev_from_otp(alloc).await
 }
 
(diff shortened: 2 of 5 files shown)


RaunakGu:
@zhalvorsen, @clundin25, @parvathib, Please help to review when you have some bandwidth, thank you

RaunakGu:
@mhatrevi, @parvathib -

Can you please take a look into this one,
Thanks

parvathib:
> @mhatrevi, @parvathib -
> 
> Can you please take a look into this one, Thanks

Thanks for the PR. I’ve documented the expected Vendor Slot 0 lifecycle, development flow, and validation responsibility in #2034. Please refer to that issue for the detailed rationale and expected behavior.

PR #2037, opened 2026-08-26
docs: correct 2.1 key revocation flows

mhatrevi:
## Summary

- identify the updated image as the Caliptra FMC + RT bundle rather than MCU Runtime
- show Caliptra Core, rather than MCU ROM, performing firmware authentication
- document one-use challenge-response authorization and active-key checks
- clarify that the rotation strap selects the second functional slot by scan order
- document target-slot metadata requirements and rotation-strap deassertion
- describe the reference implementation's maximum-key-index restriction accurately

## Validation

- `mdbook build docs`
- `git diff --check`

### DIFF
--- a/docs/src/integrator-guide.md
+++ b/docs/src/integrator-guide.md
@@ -491,7 +491,7 @@ sequenceDiagram
      Requester->>Requester: Stage Caliptra FMC + RT bundle signed with replacement key
      Requester->>MCU_ROM: Trigger cold reboot
      MCU_ROM->>Fuses: Read selected hash, PQC type, and revocations
-     MCU_ROM->>Core: Populate fuses; RI_DOWNLOAD_FIRMWARE
+    MCU_ROM->>Core: Populate fuses and send RI_DOWNLOAD_FIRMWARE
      Core->>Core: Verify Caliptra bundle with replacement key
      Core->>MCU_RT: Load and verify MCU Runtime
      Requester->>MCU_RT: Get Auth Challenge
@@ -549,7 +549,7 @@ sequenceDiagram
     MCU_ROM->>Strap: Read rotation enabled
     MCU_ROM->>Fuses: Scan valid and functional slots
     MCU_ROM->>MCU_ROM: Select second functional slot S
-    MCU_ROM->>Core: Populate slot S fuses; RI_DOWNLOAD_FIRMWARE
+    MCU_ROM->>Core: Populate slot S fuses and send RI_DOWNLOAD_FIRMWARE
     Core->>Core: Verify Caliptra bundle against slot S
     Core->>MCU_RT: Load and verify MCU Runtime
     Requester->>MCU_RT: Get Auth Challenge
--- a/docs/src/integrator-guide.md
+++ b/docs/src/integrator-guide.md
@@ -474,9 +474,9 @@ needs to be revoked while the remaining keys covered by that hash stay trusted.
     Runtime then loads and verifies MCU Runtime through the SoC manifest.
 3. After MCU Runtime starts, the requester obtains a one-use authorization
     challenge and submits `MC_FUSE_REVOKE_VENDOR_PUB_KEY` for the old key.
-4. MCU Runtime reads the active key indexes from Caliptra `FW_INFO`, rejects a
-    request targeting a key used for the current boot, and burns the old key's
-    revocation bit.
+4. MCU Runtime reads the active key indexes from Caliptra `FW_INFO`. It rejects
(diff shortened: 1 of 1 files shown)


docs/src/integrator-guide.md
```diff
@@ -462,68 +462,105 @@ within the same PK hash slot or in a different slot.
...
-3. MCU Runtime validates that the target key was not used for the current boot
-    and burns the corresponding revocation bit.
+1. Stage a Caliptra FMC + RT firmware bundle signed with a replacement vendor
+    key covered by the same vendor PK hash.
+2. Cold reboot. MCU ROM forwards the selected hash, PQC type, and revocation
+    masks to Caliptra Core. Caliptra Core ROM authenticates the new bundle; its
+    Runtime then loads and verifies MCU Runtime through the SoC manifest.
+3. After MCU Runtime starts, the requester obtains a one-use authorization
+    challenge and submits `MC_FUSE_REVOKE_VENDOR_PUB_KEY` for the old key.
+4. MCU Runtime reads the active key indexes from Caliptra `FW_INFO`, rejects a
```
parvathib: NIT: This statement is slightly ambiguous. Could this clarify that the revocation bit is burned only if the requested key was not used for the current boot?

parvathib (approved):

docs/src/integrator-guide.md
```diff
@@ -462,68 +462,106 @@ within the same PK hash slot or in a different slot.
...
-    sends `MC_FUSE_REVOKE_VENDOR_PUB_KEY` to MCU Runtime for the old key.
-3. MCU Runtime validates that the target key was not used for the current boot
-    and burns the corresponding revocation bit.
+1. Stage a Caliptra FMC + RT firmware bundle signed with a replacement vendor
+    key covered by the same vendor PK hash.
+2. Cold reboot. MCU ROM forwards the selected hash, PQC type, and revocation
+    masks to Caliptra Core. Caliptra Core ROM authenticates the new bundle; its
+    Runtime then loads and verifies MCU Runtime through the SoC manifest.
+3. After MCU Runtime starts, the requester obtains a one-use authorization
+    challenge and submits `MC_FUSE_REVOKE_VENDOR_PUB_KEY` for the old key.
```
RaunakGu, line 476: This needs to be Vendor Authenticated since Vendor key is revoked right?
How does owner perform vendor Authentication in field?

docs/src/integrator-guide.md
```diff
@@ -462,68 +462,106 @@ within the same PK hash slot or in a different slot.
...
+4. MCU Runtime reads the active key indexes from Caliptra `FW_INFO`. It rejects
+    the request if the requested key was used for the current boot; otherwise,
+    it burns that key's revocation bit.
+5. On subsequent cold boots, MCU ROM forwards the updated revocation mask and
+    Caliptra Core rejects firmware authenticated with the revoked key.
```
RaunakGu, line 481: When does the SVN update happens? Does higher SVN will be encoded together with new firmware?

RaunakGu: In this doc SVN section talks about it, SVN update happens when a firmware has higher SVN in its manifest compared to to OTP, if that is true..

And then if we use standalone command to update SVN then a new firmware download is required with higher firmware SVN before even a SVN otp update happens

docs/src/integrator-guide.md
mhatrevi: The current reference implementation uses a separate test-only command-authorization credential, not the firmware vendor or owner key. The final production authorization scheme, including key ownership and provisioning, is still under discussion.

docs/src/integrator-guide.md
mhatrevi: Yes, with one clarification: a higher firmware SVN does not automatically update OTP. The authenticated MCU Component SVN Manifest must explicitly request a higher minimum SVN, which MCU ROM validates and burns during boot. Alternatively, `MC_FUSE_INCREASE_CALIPTRA_MIN_SVN` requires higher-SVN Caliptra firmware to be running first. Key revocation itself does not update SVN.

docs/src/integrator-guide.md
mhatrevi: Updated

docs/src/integrator-guide.md
RaunakGu: Right, I was thinking even in that case a vendor authentication can't be done since it is required Vendor HSM interaction, No?,

So, do you think we could decide to keep REVOCATION fuse update command under Owner Authorization?

PR #2043, opened 2026-08-27
emulator/mci: model MCU_LSU_AXI_USER instead of stubbing it

chrisherczeg:
### Problem

`MCI_REG_MCU_LSU_AXI_USER` is a read-only strap carrying the AXI USER ID the MCU
presents on its LSU port. `Mci` never implemented the accessor, so reads fall
through to the generated stub:

```
[EMU] Non-functional register stub: read mci::mci_reg_mcu_lsu_axi_user
```

and return `0` unconditionally. Firmware that reads the strap back — for example
to program matching AXI USER filters and then verify them — cannot be exercised
in the emulator.

### Change

Store the strap value in `Mci`, return it from `read_mci_reg_mcu_lsu_axi_user`,
and allow it to be set via a new `--mcu-lsu-axi-user` option.

The register is read-only in the RDL (there is no generated
`write_mci_reg_mcu_lsu_axi_user`), so a read accessor backed by a settable field
is sufficient and cannot drift from a write path.

### Compatibility

The option defaults to `0`, which is exactly what the stub returned, so the
value observed by firmware is unchanged unless the option is passed. The only
other difference is that the stub warning no longer prints.

### Validation

- `cargo test -p caliptra-mcu-emulator-periph` — 86 passed
- `cargo check`, `cargo clippy --all-targets` (no new warnings), `cargo fmt --check`

### DIFF
--- a/emulator/app/src/emulator.rs
+++ b/emulator/app/src/emulator.rs
@@ -153,6 +153,10 @@ pub struct EmulatorArgs {
     #[arg(long)]
     pub soc_manifest: PathBuf,
 
+    /// Value reported by the read-only MCI_REG_MCU_LSU_AXI_USER strap.
+    #[arg(long, value_parser=maybe_hex::<u32>, default_value_t = 0)]
+    pub mcu_lsu_axi_user: u32,
+
     #[arg(long)]
     pub i3c_port: Option<u16>,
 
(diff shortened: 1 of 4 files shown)


chrisherczeg:

there is no new FFI surface.

PR #2044, opened 2026-08-27
emulator/caliptra: model the SS_SOC_DBG_UNLOCK_LEVEL[0] strap

chrisherczeg:
### Problem

At reset deassertion the RTL latches the Caliptra security state as

```
unlocked = (debug_intent AND SS_SOC_DBG_UNLOCK_LEVEL[0]) OR ss_dbg_manuf_enable
```

and forces the device lifecycle to Production when not unlocked.

`start_caliptra` hardcoded the result:

```rust
// Since we are at reset deassertion, SS_SOC_DBG_UNLOCK_LEVEL is 0.
// So it should ALWAYS be locked at reset deassertion.
let is_unlocked = false;
```

so the requested device lifecycle is *always* overridden to Production and a
debug-unlocked boot cannot be modelled at all. The comment block above it
already recorded the intended formula and described the hardcode as a stand-in
for the strap the emulator did not have.

### Change

Add the strap as an explicit argument on `StartCaliptraArgs` / `InitParams` and
apply the documented formula. `ss_dbg_manuf_enable` is still not modelled, so
the OR term is omitted and that is called out in the comment rather than left
implicit.

### Compatibility

The new field defaults to `false`. With `debug_intent && false` the result is
`is_unlocked == false` — identical to the hardcoded value — so behaviour is
unchanged for every existing caller unless the new `--ss-soc-dbg-unlock-level0`
flag is passed.

### Validation

- `cargo test -p caliptra-mcu-emulator-periph -p caliptra-mcu-emulator-caliptra`
- `cargo check` across `emulator-app`, `emulator-caliptra`, `hw-model`
- `cargo clippy --all-targets` (no new warnings), `cargo fmt --check`

### DIFF
--- a/emulator/app/src/emulator.rs
+++ b/emulator/app/src/emulator.rs
@@ -152,6 +152,12 @@ pub struct EmulatorArgs {
     #[arg(long)]
     pub soc_manifest: PathBuf,
 
+    /// Models the SoC strap `SS_SOC_DBG_UNLOCK_LEVEL[0]` at reset deassertion.
+    /// Combined with debug intent to decide whether the Caliptra security state
+    /// is latched as Production or honours the requested device lifecycle.
+    #[arg(long, default_value_t = false)]
+    pub ss_soc_dbg_unlock_level0: bool,
+
     #[arg(long)]
     pub i3c_port: Option<u16>,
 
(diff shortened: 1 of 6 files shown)


chrisherczeg:

there is no new FFI surface.

PR #2045, opened 2026-08-27
emulator: add --pc-coverage to log unique MCU program counters

chrisherczeg:
### What this adds

An opt-in `--pc-coverage` flag that records every distinct MCU program counter
to `pc.log` in the log directory, one PC per line in first-seen order.

This gives a cheap coverage signal for ROM and runtime firmware — which blocks
were reached on a given run — without standing up an external tracer or
instrumenting the firmware build.

### Implementation

`--trace-instr` already installs a per-instruction step callback, so the
coverage hook reuses it rather than adding a second one:

- the callback is installed when **either** option is set
- the existing tracing logic is nested inside it unchanged, so disassembly
  output is byte-for-byte identical
- dedup is a `HashSet<u32>`; only first-seen PCs are written

The two features are fully independent — either, both, or neither can be
enabled. With neither set the callback is not installed at all, so the hot
stepping path is exactly as it is today.

### Validation

- `cargo test -p caliptra-mcu-emulator-periph -p caliptra-mcu-emulator-caliptra`
- `cargo check`, `cargo clippy --all-targets` (no new warnings), `cargo fmt --check`

### DIFF
--- a/emulator/app/src/emulator.rs
+++ b/emulator/app/src/emulator.rs
@@ -50,6 +50,7 @@ use clap_num::maybe_hex;
 use crossterm::event::{Event, KeyCode, KeyEvent};
 use std::cell::Cell;
 use std::cell::RefCell;
+use std::collections::HashSet;
 use std::fs::File;
 use std::io::{self, IsTerminal, Read, Write};
 use std::ops::Range;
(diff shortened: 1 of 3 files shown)


chrisherczeg:

there is no new FFI surface.

Issue #1993, opened 2026-08-18
Add build-time validation for per-component SVN metadata


mhatrevi:
Follow-up to #1966 and #1992.

## Context

The release pipeline currently relies on the integrator to keep three inputs synchronized:

1. `component_id` values in the signed SoC Manifest IMC.
2. `(component_id, current_svn, min_svn)` entries in the MCU Component SVN Manifest.
3. The platform `component_id -> SOC_IMAGE_MIN_SVN[i]` mapping in `SVN_FUSE_MAP`.

MCU ROM authenticates the SVN manifest as part of MCU Runtime, but it does not cross-check the SVN records against the IMC or component bytes. An incorrectly constructed but validly signed release can therefore claim an SVN for the wrong component or irreversibly burn an unintended fuse floor. Unmapped IDs are currently skipped by design, so intentional non-enforcement must remain expressible rather than being inferred from a missing mapping.

## Proposed work

Add a pre-signing/build-time validator driven from the same release configuration used to generate the SoC Manifest and MCU Component SVN Manifest.

## Acceptance criteria

- Reject duplicate SVN entries with conflicting values.
- Reject SVN entries for unknown IMC `component_id` values unless explicitly marked as intentional future/absent components.
- Detect protected IMC components that are missing required SVN metadata.
- Validate `min_svn <= current_svn` and platform fuse-range constraints.
- Validate `component_id -> SVN_FUSE_MAP` coverage, with an explicit allowlist/policy for intentionally unenforced components.
- Validate components sharing a fuse slot use a compatible floor policy.
- Run validation before artifacts are signed or OTP floor requests can ship.
- Add positive coverage for multiple unique `fw_id` instances sharing one `component_id`, plus negative tests for unknown, missing, duplicate, and conflicting records.

mlvisaya:
I have some discussion with @parvathib , we think there should be a single source of truth for the soc configurations.

There should be single source of truth configuration file that is used not only for verification but also for generation of the artifacts (auth manifest, mcu manifest, etc).

Also, we should probably better use TOML rather than JSON format. In general, Caliptra configs are written in TOML.

There are currently 4 artifacts that are dependent on the soc metadata:

SoC Authorization Manifest
MCU Component SVN Manifest
Attestation Manifest
SoC firmware ID list
And we have the builders, that generate those artifacts:

cargo xtask runtime-build
cargo xtask all-build
cargo xtask auth-manifest create
[...]

PR #2060, opened 2026-09-01
Adding build-time validation for per-component SVN metadata

v-focutter:
As per task [#1993](https://github.com/chipsalliance/caliptra-mcu-sw/issues/1993), adding build-time validation for SVN metadat.


mlvisaya:
Although the addition of a common policy file is good (ci-component-svn-policy), I think it would be better to have a single source of truth configuration file that is used not only for verification but also for **generation** of the artifacts (auth manifest, mcu manifest, etc). 

Also, we should probably better use TOML rather than JSON format. In general, Caliptra configs are written in TOML.

There are currently 4 artifacts that are dependent on the soc metadata:

1. SoC Authorization Manifest
2. MCU Component SVN Manifest
3. Attestation Manifest
4. SoC firmware ID list

And we have the builders, that generate those artifacts:
1. cargo xtask runtime-build 
2. cargo xtask all-build 
3. cargo xtask auth-manifest create
4. cargo xtask --features fpga_realtime -- fpga build


We should have a common config file that contains all the soc metadata that will be used to generate the 4 artifacts mentioned above.

The TOML config file would contain:
```
vendor = "Vendor A" // e.g. "Microsoft"
model = "Model-XXXX"
soc_manifest_svn = 5

[component_svn_manifest]
current_svn = 3
min_svn = 2
caliptra_runtime_min_svn = 7
soc_manifest_min_svn = 5

[[soc_images]]
path = "firmware/soc-image-a.bin"
fw_id = 0x0000_1000
component_id = 0x0000_1000
load_address = 0xB000_0000
staging_address = 0x8000_0000
exec_bit = 5
current_svn = 4
min_svn = 2
is_tcb = true
is_ak_target = false
network_filename = "soc-image-a.bin"

[[soc_images]]
path = "firmware/soc-image-b.bin"
fw_id = 0x0000_1001
component_id = 0x0000_1001
load_address = 0xB010_0000
staging_address = 0x8010_0000
exec_bit = 6
current_svn = 3
min_svn = 1
is_tcb = true
is_ak_target = false

// Feature specific overrides
[features."test-firmware-update-flash"]
soc_manifest_svn = 6

[features."test-firmware-update-flash".mcu_image]
staging_address = 0x6008_0000
current_svn = 6
min_svn = 4

[features."test-mctp-spdm-attestation"]
soc_manifest_svn = 7

[[features."test-mctp-spdm-attestation".soc_images]]
path = "firmware/attestation-image.bin"
fw_id = 0x2000
component_id = 0x2000
load_address = 0xB020_0000
staging_address = 0x8020_0000
exec_bit = 7
current_svn = 2
min_svn = 1
is_tcb = true
is_ak_target = false
```

We can pass this then to the builders:
1. cargo xtask runtime-build --component_config <component-config-toml>
2. cargo xtask all-build --component_config <component-config-toml>
3. cargo xtask auth-manifest create --component_config <component-config-toml>
4. cargo xtask --features fpga_realtime -- fpga build --component_config <component-config-toml>


Currently, there's a mcu_cfg parameter that contains some of the information in the TOML file, I think if "component-config" is specified we can ignore mcu_cfg parameter.

We can also add a validator to make sure that the soc manifest and the MCU runtime manifest are valid based on the component_config:

cargo xtask auth-manifest verify --component_config <component-config-toml> <Auth Manifest File> -> outputs valid/invalid after checking the auth manifest file with the component_config

.github/workflows/attestation.yml
```diff
@@ -144,6 +144,7 @@ jobs:
       - name: Build firmware with test SPDM responder features
         run: |
           cargo xtask all-build --separate-runtimes --runtime-features ${{ matrix.runtime_feature }} \
+            --component-svn-config builder/ci-component-svn-policy \
```
mlvisaya, line 147: we don't need the component-svn-policy anymore because the component_config is now the single source of truth including the SVNs. In here, we can just pass the --component-config

.github/workflows/build-test.yml
```diff
@@ -101,13 +101,16 @@ jobs:
       - name: Build ROM and runtime binaries
         run: |
           # Build with parallel-build feature to enable Rayon and unique target directories
-          cargo run -p xtask --features parallel-build -- all-build --platform emulator --separate-runtimes
+          cargo run -p xtask --features parallel-build -- all-build \
+            --platform emulator --separate-runtimes \
+            --component-svn-config builder/ci-component-svn-policy
```
mlvisaya: same here, and all of the instances of --component-svn-config, we should pass --component-config instead

builder/src/all.rs
```diff
@@ -628,6 +628,8 @@ pub struct AllBuildArgs<'a> {
     pub runtime_features: Option<&'a str>,
     pub separate_runtimes: bool,
     pub soc_images: Option<Vec<ImageCfg>>,
+    pub component_svn_validation: Option<crate::ComponentSvnValidationConfig>,
```
mlvisaya, line 690: can remove component_svn_validation, it will get it from component_config

builder/src/caliptra.rs
```diff
@@ -103,6 +103,7 @@ pub struct CaliptraBuilder {
     soc_images: Option<Vec<ImageCfg>>,
     mcu_image_cfg: Option<ImageCfg>,
     soc_manifest_svn: Option<u32>,
+    component_svn_validation: Option<ComponentSvnValidationConfig>,
```
mlvisaya, line 107: builder will now take ComponentConfig

PR #2064, opened 2026-09-02
spdm: separate streamed and buffered message limits

parvathib:
### Commit 1: SPDM streamed and buffered message limits

- Size streamed certificate responses from the requester receive limit.
- Advertise inbound logical request capacity independently of scratch buffering.
- Stream SET_CERTIFICATE directly into storage with a tunable standard CertChain limit.
- Reassemble MCTP AuthorizedCommand requests without inflating DOE task state.
- Validate CHUNK_SEND frames against responder transport and logical-request limits.

### Commit 2: Caliptra API ML-DSA signing primitives

- Add SHAKE256 and ML-DSA tr/mu helpers.
- Add raw public-key retrieval and normal-mailbox DPE signing.
- Model Caliptra 2.0 raw-message and 2.1 external-mu signing explicitly.

### Commit 3: SPDM ML-DSA cert-store signing

- Route typed signing inputs through the Measurement API and SPDM certificate store.
- Preserve DPE handle rotation across signing operations.

### DIFF
--- a/platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
+++ b/platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
@@ -36,56 +36,93 @@ use core::fmt::Write as _;
 use core::ptr::NonNull;
 use embassy_executor::Spawner;
 
-/// Largest single in-flight SPDM message (request or response) this responder
-/// supports.
+/// Scratch-backed limit for large messages retained in full.
 ///
-/// This is a contract, not a measurement: it is advertised verbatim as
-/// `MaxSPDMmsgSize` in `CAPABILITIES`, so each responder scratch pool must be
-/// able to satisfy it. Raising this requires raising both scratch pools; the
-/// assertions below enforce that.
-///
-/// Sized to cover both directions, because `MaxSPDMmsgSize` caps every SPDM
-/// message and the `CHUNK_SEND` admission check compares against it before the
-/// streaming path is reached:
-///
-/// * Responses — the largest evidence this build can emit (an ML-DSA-87 PCR
-///   quote is 6388 bytes) plus SPDM and vendor-defined framing.
-/// * Requests — the production debug unlock token (7504 bytes of ML-DSA-87 key
-///   and signature material) plus framing. This one is streamed into the
-///   Caliptra mailbox and never occupies the scratch pool, but declaring a
-///   smaller `MaxSPDMmsgSize` would have the responder reject it outright.
-///
-/// The assertions below check both against what the build actually enables, so
-/// turning on a larger generator or a larger inbound command fails the build
-/// rather than failing at runtime.
-const MAX_SPDM_MSG_SIZE: usize = {
+/// It contributes to `MaxSPDMmsgSize` only when buffered large requests are
+/// enabled. Raising it requires larger scratch pools.
+const MAX_BUFFERED_SPDM_MSG_SIZE: usize = {
     let declared = 8 * 1024;
     assert!(
         declared
             >= caliptra_mcu_spdm_vdm_handler::iana::ocp::caliptra_vdm::large_response_capacity::<
                 crate::caliptra_cmd_handler::CaliptraCmdBackend,
             >(),
-        "MaxSPDMmsgSize is smaller than the largest Caliptra VDM response this build can \
-         produce; raise MAX_SPDM_MSG_SIZE (and both scratch pools with it) or disable an \
-         evidence generator"
-    );
-    assert!(
-        declared >= MAX_STREAMED_VDM_REQUEST_LEN,
-        "MaxSPDMmsgSize is smaller than the largest Caliptra VDM request this build must \
-         accept (the debug unlock token); raise MAX_SPDM_MSG_SIZE and both scratch pools with it"
+        "SPDM scratch capacity is smaller than the largest buffered VDM response; raise \
+         MAX_BUFFERED_SPDM_MSG_SIZE and both responder scratch pools"
     );
     declared
 };
 
-/// Largest streamed Caliptra VDM request this responder must accept.
+/// Endpoint-wide logical request limit advertised as `MaxSPDMmsgSize`.
 ///
-/// The production debug unlock token dominates every other inbound request: it
-/// carries an ML-DSA-87 public key and signature. It is streamed into the
-/// Caliptra mailbox and never occupies the scratch pool, so it costs no memory
-/// here, but [`MAX_SPDM_MSG_SIZE`] must still admit it.
+/// This is the maximum across buffered and streamed inbound request paths.
+const MAX_INBOUND_SPDM_REQUEST_SIZE: usize = {
+    let mut required = MAX_TRANSPORT_MTU;
+    if MAX_BUFFERED_SPDM_REQUEST_LEN > required {
+        required = MAX_BUFFERED_SPDM_REQUEST_LEN;
+    }
+    if MAX_STREAMED_SPDM_REQUEST_LEN > required {
+        required = MAX_STREAMED_SPDM_REQUEST_LEN;
+    }
+    required
+};
+
+/// Maximum across scratch-backed request handlers. Add new buffered request
+/// limits to this block.
+const MAX_BUFFERED_SPDM_REQUEST_LEN: usize = {
+    let mut required = 0;
+    if MAX_BUFFERED_VDM_REQUEST_LEN > required {
+        required = MAX_BUFFERED_VDM_REQUEST_LEN;
+    }
+    required
+};
+
+/// Logical request ceiling for generic VDM CHUNK_SEND reassembly.
+///
+/// Caliptra `AuthorizedCommand` requests are retained here in full before VDM
+/// dispatch. Streamed debug-unlock and SET_CERTIFICATE requests bypass this
+/// buffer. Raise this together with both SPDM scratch pools when a buffered
+/// request grows, or add a streaming handler instead.
+const MAX_BUFFERED_VDM_REQUEST_LEN: usize = MAX_BUFFERED_SPDM_MSG_SIZE;
+const _: () = assert!(
+    MAX_BUFFERED_VDM_REQUEST_LEN
+        >= caliptra_mcu_spdm_vdm_handler::iana::ocp::caliptra_vdm::
+            MAX_AUTHORIZED_COMMAND_SPDM_REQUEST_LEN,
+    "buffered SPDM request capacity is smaller than an AuthorizedCommand request"
+);
+
+/// Maximum across streaming request handlers. Add new streamed request limits
+/// to this block.
+const MAX_STREAMED_SPDM_REQUEST_LEN: usize = {
+    #[cfg(feature = "test-mctp-spdm-set-certificate")]
+    {
+        if MAX_STREAMED_SET_CERTIFICATE_REQUEST_LEN > MAX_STREAMED_VDM_REQUEST_LEN {
+            MAX_STREAMED_SET_CERTIFICATE_REQUEST_LEN
+        } else {
+            MAX_STREAMED_VDM_REQUEST_LEN
+        }
+    }
+
+    #[cfg(not(feature = "test-mctp-spdm-set-certificate"))]
+    {
+        MAX_STREAMED_VDM_REQUEST_LEN
+    }
+};
+
+/// Maximum logical size of the streamed debug-unlock VDM request.
 const MAX_STREAMED_VDM_REQUEST_LEN: usize =
-    caliptra_mcu_spdm_vdm_handler::iana::ocp::caliptra_vdm::LARGE_REQUEST_FRAMING_LEN
+    caliptra_mcu_spdm_vdm_handler::iana::ocp::caliptra_vdm::SPDM_REQUEST_FRAMING_LEN
         + core::mem::size_of::<mcu_caliptra_api::mailbox::ProductionAuthDebugUnlockToken>();
+
+/// Maximum streamed SET_CERTIFICATE request for a CertChain.
+///
+/// This limit is tunable based on the integrator's requirements.
+#[cfg(feature = "test-mctp-spdm-set-certificate")]
+const MAX_STANDARD_CERT_CHAIN_LEN: usize = u16::MAX as usize;
+#[cfg(feature = "test-mctp-spdm-set-certificate")]
+const MAX_STREAMED_SET_CERTIFICATE_REQUEST_LEN: usize = caliptra_mcu_spdm_codec::SpdmMsgHdrPdu::SIZE
+    + caliptra_mcu_spdm_codec::SetCertificateReqBody::SIZE
+    + MAX_STANDARD_CERT_CHAIN_LEN;
 
 /// Conservative upper bound on transport MTU. The real MTU is a runtime
 /// transport property, so the budget uses a declared ceiling instead.
@@ -110,14 +147,14 @@ const TRANSIENT_MAILBOX_PEAK: usize = 2560;
 /// The receive buffer is not counted: it is shrunk to the actual frame length
 /// immediately after receive, and the request that triggers a large response
 /// is always a single small frame.
-const LARGE_MSG_PATH_PEAK: usize = MAX_SPDM_MSG_SIZE + MAX_TRANSPORT_MTU;
+const LARGE_MSG_PATH_PEAK: usize = MAX_BUFFERED_SPDM_MSG_SIZE + MAX_TRANSPORT_MTU;
 
 /// Peak concurrent allocation on the certificate / secure-session path: the
 /// mailbox working set plus the secured-message plaintext and ciphertext
 /// staging buffers.
 const CRYPTO_PATH_PEAK: usize = TRANSIENT_MAILBOX_PEAK + 2 * MAX_TRANSPORT_MTU;
 
-/// Minimum scratch pool that can satisfy [`MAX_SPDM_MSG_SIZE`].
+/// Minimum scratch pool that can satisfy [`MAX_BUFFERED_SPDM_MSG_SIZE`].
 ///
 /// The two request paths are mutually exclusive: a single request either
 /// builds a large chunked response or runs the certificate/crypto path, never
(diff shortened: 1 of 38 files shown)


runtime/userspace/api/spdm-lib/stack/src/chunk/mod.rs
```diff
@@ -374,15 +376,7 @@ pub(crate) fn validate_buffered_large_response_with_capacity<Pal: SpdmPal>(
...
-    let peer_max_spdm_msg_size = if state.peer_max_spdm_msg_size == 0 {
-        local_max_spdm_msg_size
-    } else {
-        state.peer_max_spdm_msg_size as usize
-    };
-    let effective_max_spdm_msg_size = local_max_spdm_msg_size.min(peer_max_spdm_msg_size);
-
-    if large_resp_len > capacity || large_resp_len > effective_max_spdm_msg_size {
+    if large_resp_len > state.max_buffered_response_size(capacity)? {
         return Err(SPDM_UNSPECIFIED);
```
amit-kumarh, line 388: Perhaps we consider returning a ResponseTooLarge error instead of unspecified?

runtime/userspace/api/spdm-lib/stack/src/tests/certificate.rs
```diff
@@ -299,6 +301,47 @@ fn test_get_certificate_chunked_full_fetch() {
     assert_eq!(&payload[SPDM_CERT_CHAIN_HDR_LEN..], TEST_CERT_CHAIN);
 }
 
+#[test]
+fn test_get_certificate_chunked_full_fetch_exceeds_local_buffer() {
```
amit-kumarh: Is this test significantly different than the one above it? (`test_get_certificate_chunked_full_fetch_exceeds_local_buffer`) Maybe we can just use the larger STREAMED_CERT_CHAIN in that test?

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
```diff
@@ -889,17 +908,59 @@ async fn dispatch_large_request<Pal: SpdmPal, Vdm: SpdmVdmBackend>(
...
-            state,
-            pal,
-            io,
-            large_req,
-            secure_session,
-            out,
-        )
-        .await
-        .map_err(Into::into),
+        ReqRespCode::VENDOR_DEFINED_REQUEST => {
```
amit-kumarh: Why was this handling moved out of the `vendor_defined` module?

runtime/userspace/api/caliptra-api/src/dpe.rs
```diff
@@ -1434,6 +1668,16 @@ mod tests {
         assert_eq!(DERIVE_CONTEXT_DPE_PAYLOAD_LEN, (12 + 80) as u32);
     }
 
+    #[test]
+    fn sign_mldsa87_wire_layout() {
```
amit-kumarh: are these not taken care of by the const assertions at the top?

amit-kumarh (approved):
LGTM with the above comments

runtime/userspace/api/caliptra-api/src/dpe.rs
```diff
@@ -132,7 +168,17 @@ struct SignP384Cmd {
...
     label: [u8; DPE_LABEL_LEN],
     flags: U32,
-    digest: [u8; DPE_LABEL_LEN], // same size as hash (48)
+    digest: [u8; DPE_P384_DIGEST_SIZE],
+}
+
+/// `dpe::commands::SignMldsa87Cmd`.
+#[repr(C)]
+#[derive(FromBytes, IntoBytes, KnownLayout, Immutable, Unaligned)]
+struct SignMldsa87Cmd {
```
zhalvorsen, line 177: Can we use the definitions from DPE directly here to avoid bit rot?

runtime/userspace/api/caliptra-api/src/dpe.rs
```diff
@@ -970,6 +1041,43 @@ pub async fn dpe_certify_key_pubkey<A: ApiAlloc>(
...
+fn parse_certify_key_mldsa87_pubkey(
+    response: &[u8],
+    public_key: &mut [u8; CERTIFY_KEY_MLDSA87_PUBKEY_SIZE],
+) -> McuResult<()> {
+    validate_certify_key_prefix(response, DpeProfile::Mldsa87)?;
+    copy_bytes(
+        public_key,
+        internal_slice(
+            response,
+            CERTIFY_KEY_RESP_PUBKEY_X_OFF,
```
zhalvorsen, line 1075: Could use `offset_of!` here with the DPE defined response

```Rust
offset_of!(CertifyKeyMldsa87RespHdr, pubkey)

**Review by ArthurHeymans — COMMENTED** [review]
I have a few questions.

File: runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs [hunk 7]
```diff
@@ -152,8 +154,8 @@ async fn process_chunk_send<Pal: SpdmPal, Vdm: SpdmVdmBackend>(
         return Err(ChunkProcessError::Spdm(SPDM_UNEXPECTED_REQUEST));
     }
 
-    // Ensure the incoming request message fits our standard effective bounds, not raw MTU.
-    if req.len() > state.effective_data_transfer_size(pal) {
+    // Incoming chunks use our DataTransferSize; the peer's limit is outbound.
+    if req.len() > pal.mtu() {
```
ArthurHeymans, line 164: What prevents a second MCTP endpoint from continuing or aborting a plaintext chunked request started by another endpoint? The reassembly state appears to bind only to `session_id`, which is `None` for both peers.

runtime/userspace/api/spdm-lib/stack/src/tests/stack_set_certificate.rs
```diff
@@ -256,7 +346,7 @@ fn handshake_session_chunked_set_certificate_is_rejected() {
 
 #[test]
 fn plaintext_chunked_set_certificate_bad_root_hash_does_not_commit() {
-    let pal = TestPal::default();
+    let pal = chunked_set_certificate_pal();
```
ArthurHeymans, line 349: Could this test start with an already-provisioned certificate and verify that it survives the rejected stream? `begin_stream_update()` appears to erase the live region before the final DER and root hash are validated, while abort only clears `write_in_progress`.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
```diff
@@ -282,7 +320,8 @@ async fn spdm_doe_responder() {
             allocator,
             crate::cert_store::shared(),
             measurement_provider(),
-            MAX_SPDM_MSG_SIZE,
+            MAX_INBOUND_SPDM_REQUEST_SIZE,
```
ArthurHeymans, line 323: Is `SET_CERTIFICATE` guaranteed to be exposed through only one responder in supported configurations? If MCTP and DOE can run together, the unconditional `write_in_progress.store(true, ...)` appears to allow both tasks to begin writing the same shared slot.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
```diff
@@ -223,7 +260,8 @@ async fn spdm_mctp_responder() {
...
```
ArthurHeymans, line 263: How are the DOE certificate caches invalidated when MCTP updates this shared certificate store, and vice versa? The caches are task-local, while invalidation seems to affect only the writer's `TaskCertStore`.

runtime/userspace/api/spdm-lib/stack/src/challenge.rs
```diff
@@ -156,7 +156,13 @@ pub(crate) async fn handle_challenge<'a, Pal: SpdmPal>(
         .get_mut(head + no_sig_len..head + no_sig_len + ECC_P384_SIGNATURE_SIZE)
         .ok_or(SPDM_UNSPECIFIED)?;
     let sig_len = pal
-        .sign_hash(io, slot_id, asym_algo, &m1_hash, sig_slot)
+        .sign(
+            io,
+            slot_id,
+            asym_algo,
+            SigningInput::EccP384Digest(&m1_hash),
```
ArthurHeymans, line 163: How does this work when ML-DSA-87 is negotiated? This still allocates a 96-byte ECDSA signature and passes `EccP384Digest`, which appears incompatible with the ML-DSA profile selected by `asym_algo`.

runtime/userspace/api/spdm-lib/pal/src/cert/endorsement.rs
```diff
@@ -271,7 +271,21 @@ const MANAGED_ERASED_BYTE: u8 = 0xFF;
 #[cfg(feature = "set-certificate")]
 const MANAGED_KEY_USAGE_MASK: u16 = 0x0003;
 #[cfg(feature = "set-certificate")]
-const MANAGED_MAX_DER_LEN: usize = (u16::MAX as usize) - 52;
+const SPDM_CERT_CHAIN_HEADER_SIZE: usize = 4 + 48;
+/// This limit is tunable based on the integrator's requirements.
+#[cfg(feature = "set-certificate")]
+const MANAGED_MAX_DER_LEN: usize = (u16::MAX as usize) - SPDM_CERT_CHAIN_HEADER_SIZE;
```
ArthurHeymans, line 277: Should this limit reserve space for the DPE chain and generated leaf certificate that are appended later? An endorsement at this maximum makes the composed chain exceed `u16::MAX`, while `fill_cert_chain_portion()` casts the total length to `u16`.

runtime/userspace/api/spdm-lib/pal/src/cert/mod.rs
```diff
@@ -304,41 +304,19 @@ impl<M: MeasurementProvider> SpdmPalCertStore for McuSpdmPal<M> {
...
             .load(Ordering::Relaxed)
         {
             return Err(INVARIANT);
         }
 
-        let cert_slot = &self.cert_store.cert_slots()[idx];
-        let capacity = cert_slot
+        // SlotSizeRequested describes endorsement storage. The generated DPE
+        // and leaf certificates appended to GET_CERTIFICATE are not stored here.
+        self.cert_store.cert_slots()[idx]
```
ArthurHeymans, line 316: Should `read_cert_chain()` and `root_cert_hash()` also reject access while `write_in_progress` is set? Otherwise another responder may read erased or partially written shared flash during a streamed update.

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
```diff
@@ -695,7 +706,14 @@ async fn build_final_chunk_send_ack<'a, Pal: SpdmPal, Vdm: SpdmVdmBackend>(
     chunk_seq_num: u16,
 ) -> SpdmResult<PalBytes<'a, Pal>> {
     let len = state.large_msg_ctx.state.large_msg_size as usize;
-    let mut response_to_large_request = [0u8; LARGE_REQUEST_RESPONSE_BUF_SIZE];
+    let mut response_to_large_request =
+        match SpdmPalAlloc::alloc(pal, io, [0u8; LARGE_REQUEST_RESPONSE_BUF_SIZE]) {
```
ArthurHeymans, line 716: If this allocation fails, should the response still be a `CHUNK_SEND_ACK` carrying an early error and should the reassembly state be reset? The current path appears to return a bare SPDM error while leaving the completed request active.

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
ArthurHeymans, line 164: What eventually releases an active stream if its owner stops sending chunks or the responder exits after a transport error? I could not find a timeout, and several `Spdm` error paths preserve the request and any certificate write lock.

runtime/userspace/api/spdm-lib/stack/src/chunk/mod.rs
parvathib: Good point. The previous check conflated two limits, so I split them cleanly. We have 3 categories now:
- `Response size > (Requester's MaxSPDMmsgSize)` -> return  `ResponseTooLarge`  with  `ActualSize` as per spec.
- `Response size > responder buffered capacity` -> return `Unspecified`; this is actually a responder configuration issue we should have sufficient buffer for all our large buffered responses.
- `Response size > DataTransferSize`  but fits both requester/responder buffer limits, and chunking is enabled -> return  `LargeResponse` error with `handle` and serve it through `CHUNK_GET`.

runtime/userspace/api/spdm-lib/stack/src/tests/certificate.rs
parvathib: Agreed. The response format differed, but both tests exercised the same chunked full-fetch path. I updated the SPDM 1.4 large-certificate test to use STREAMED_CERT_CHAIN , so it also verifies a chain larger than local buffered capacity and removed the duplicate test.

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
parvathib: It was moved by accident. Restored it.

runtime/userspace/api/caliptra-api/src/dpe.rs
parvathib: Yes, the compile-time layout assertions already take care of this. I'll remove the tests.

runtime/userspace/api/caliptra-api/src/dpe.rs
parvathib: We cannot use the DPE structs directly as buffer views because they require 4 byte aligned storage, which `ApiAlloc` doesn't guarantee. However, as you suggested, we can use the DPE types as the authoritative layout through `size_of!`  and `offset_of!`  and remove the local mirrored definitions. This lets us continue reading and writing the allocator-backed byte buffers directly without placing req/resp copies on the stack. I’ll include that cleanup in the separate PR for the `caliptra-sw` bump and DPE profile update to hybrid.

zhalvorsen: Sounds good, thanks!

runtime/userspace/api/caliptra-api/src/dpe.rs
parvathib: Good suggestion. I’ll track using the DPE definitions through  `size_of!`  and  `offset_of!`  in a follow-up issue, since it depends on the DPE profile update.
I created an issue for this clean up in general: https://github.com/chipsalliance/caliptra-mcu-sw/issues/2073,  and will add a follow up PR.

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
parvathib: Caliptra’s current supported MCTP topology has a single requester and supports one outstanding large-message transfer. While reassembly is active, a  CHUNK_SEND  is accepted only if its handle and sequence number match the expected continuation. A new transfer begins at sequence zero, so it is rejected and the active assembly is reset. In this topology, that would indicate requester malfunction or a protocol violation rather than cross-endpoint interference.

runtime/userspace/api/spdm-lib/stack/src/tests/stack_set_certificate.rs
parvathib: That is a valid concern.  `begin_stream_update()` erases the live region before validation and abort only clears `write_in_progress` , so a rejected stream destroys the provisioned cert today. I have an upcoming refactor on cert store which includes the A/B cert-store provisioning change (stage to backup store, validate before switching the backup to active) fixes it, and I'll add the survival test there.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
parvathib: Both responders do share the same store, so you're right that nothing today prevents it. I have an upcoming cert-store refactor PR that adds synchronization for two transports concurrently updating the store, and I'll address this there.

platforms/emulator/runtime/userspace/apps/user/src/spdm/mod.rs
parvathib: You are right again - the store is shared but each task keeps its own view, so a write on one transport doesn't invalidate the other's cached state. The upcoming cert-store refactor PR tracks provisioning state centrally so updates are visible across transports, and I'll address this there along with the concurrent-write synchronization.

runtime/userspace/api/spdm-lib/stack/src/challenge.rs
parvathib: I believe that work is scoped to the SPDM 1.4 CHALLENGE support PR. This PR only adds the typed signing primitives and cert-store routing.

runtime/userspace/api/spdm-lib/pal/src/cert/endorsement.rs
parvathib: Good point.  `MANAGED_MAX_DER_LEN` is an integrator configured limit for the endorsement DER chain, so I am leaving it unchanged rather and not reserving a fixed amount for the DPE chain and generated leaf. 
However, I added checks for final composed length and use u16 conversions in both certificate streaming and hashing. 
We should also enable large cert chain fields end to end. I created an issue for this. We'll address in a separate PR: https://github.com/chipsalliance/caliptra-mcu-sw/issues/2075

runtime/userspace/api/spdm-lib/pal/src/cert/mod.rs
parvathib: Agree! those two don't check it today (unlike  cert_chain_len() / cert_chain_slot_size()) . The upcoming cert-store refactor PR takes a different approach: an update is staged separately and only published once fully validated, so readers keep seeing the current certificate for the duration and never observe an erased or partial state. No need to reject reads mid-update, and I'll cover this there.

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
parvathib: Addressed. On allocation failure, the stack now aborts and resets the reassembly, then returns `CHUNK_SEND_ACK`  with  `ERROR(Unspecified)`  in  `ResponseToLargeRequest` .  `EarlyErrorDetected`  remains clear because the last chunk was already received.

runtime/userspace/api/spdm-lib/stack/src/chunk/send.rs
parvathib: correct! there's no timeout today.  The upcoming cert-store A/B refactor makes that harmless instead of adding a timer: an abandoned stream leaves the current certificate intact, and the next authorized update supersedes the stale session.

ArthurHeymans (approved):

PR #2065, opened 2026-09-02
caliptra-api: add AxiDmaTarget abstraction and mcu_sram_to_axi_dma helper

arul-sri:
- Introduce the `dma` module in `mcu-caliptra-api` defining Caliptra subsystem AXI DMA integration constants (MCU_SRAM_LOCAL_BASE, MCI_BASE_AXI_ADDRESS, MCU_SRAM_AXI_OFFSET, MCU_SRAM_AXI_BASE).
- Implement AxiDmaTarget and mcu_sram_to_axi_dma for safe address translation from MCU local SRAM buffers to subsystem AXI DMA target descriptors.
- Add unit tests for AXI DMA address translation and bounds checking.
- Refactor dpe_derive_context_exported_cdi in dpe.rs to use the new helper.

### DIFF
--- a/runtime/userspace/api/caliptra-api/src/dma.rs
+++ b/runtime/userspace/api/caliptra-api/src/dma.rs
@@ -10,16 +10,23 @@
 //! - MCI (Microcontroller Interface) base address on the AXI bus: `0xA800_0000`
 //! - MCU SRAM offset within MCI: `0x00C0_0000` (12 MiB)
 //! - MCU SRAM base address on the AXI bus: `0xA8C0_0000`
-//! - MCU SRAM base address in the MCU's local address space: `0x4000_0000`
+//!
+//! However, the MCU local address space and SRAM size differ depending on the platform:
+//! - **Emulator**: Local SRAM base is `0x4000_0000`, size is 1024 KiB (or 512 KiB in release layout).
+//! - **FPGA**: Local SRAM base is `0xA8C0_0000`, size is 512 KiB.
 //!
 //! To allow Caliptra Core to DMA directly into a buffer allocated in MCU SRAM,
-//! local SRAM addresses must be translated to their subsystem AXI equivalents.
+//! local SRAM addresses must be validated (alignment, bounds) and translated to their
+//! subsystem AXI equivalents.
+//!
+//! # Note on DMA Abstractions
+//! The userspace syscall crate provides [`caliptra_mcu_libsyscall_caliptra::dma::DMAMapping`]
+//! for the MCU's internal AXI CDMA controller (`mcu_sram_to_mcu_axi` and `cptra_axi_to_mcu_axi`).
+//! In contrast, [`AxiDmaConfig`] and [`AxiDmaTarget`] handle address translation for Caliptra
+//! Core's internal DMA engine, which acts as a master on the subsystem AXI interconnect.
 
 use mcu_error::codes::INVARIANT;
 use mcu_error::McuResult;
-
-/// Physical base address of MCU SRAM from the MCU's local memory map.
-pub const MCU_SRAM_LOCAL_BASE: u32 = 0x4000_0000;
 
 /// Base address of the MCI on the subsystem AXI bus.
 pub const MCI_BASE_AXI_ADDRESS: u32 = 0xA800_0000;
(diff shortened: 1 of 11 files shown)


runtime/userspace/api/caliptra-api/src/dma.rs
```diff
@@ -0,0 +1,120 @@
...
+use mcu_error::McuResult;
+
+/// Physical base address of MCU SRAM from the MCU's local memory map.
+pub const MCU_SRAM_LOCAL_BASE: u32 = 0x4000_0000;
+
+/// Base address of the MCI on the subsystem AXI bus.
+pub const MCI_BASE_AXI_ADDRESS: u32 = 0xA800_0000;
+
+/// Offset of MCU SRAM within MCI address space (12 MiB).
+pub const MCU_SRAM_AXI_OFFSET: u32 = 0x00C0_0000;
```
clundin25: These addresses are different on FPGA. I think we need to pass these values from the platform specific board definitions, so this library code is agnostic to the shape of the hardware.

For example, in these two files

```
platforms/emulator/runtime/src/board.rs
platforms/fpga/runtime/src/board.rs
```

Pass some data type that contains these value.

I think we should also double check if there is some build time solution for this. Right now we would have to keep the board.rs file in sync with the linker / firmware builder manifests

runtime/userspace/api/caliptra-api/src/dma.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// Translates a raw pointer and length in MCU local SRAM into an [`AxiDmaTarget`].
+    pub fn from_mcu_sram_ptr(ptr: *const u8, len: usize) -> McuResult<Self> {
+        let local_addr = ptr as u32;
+        let sram_offset = local_addr
+            .checked_sub(MCU_SRAM_LOCAL_BASE)
+            .ok_or(INVARIANT)?;
+        let addr = MCU_SRAM_AXI_BASE
+            .checked_add(sram_offset)
+            .ok_or(INVARIANT)?;
```
clundin25, line 114: I think we need some upperbound here too, so we don't write past the SRAM memory

runtime/userspace/api/caliptra-api/src/dma.rs
```diff
@@ -0,0 +1,120 @@
...
+    ///
+    /// Returns [`INVARIANT`] if the buffer is not located within MCU local SRAM
+    /// (`< MCU_SRAM_LOCAL_BASE`), if pointer arithmetic overflows, or if the buffer
+    /// length cannot fit in a `u32`.
+    pub fn from_mcu_sram(buf: &[u8]) -> McuResult<Self> {
+        Self::from_mcu_sram_ptr(buf.as_ptr(), buf.len())
+    }
+
+    /// Translates a raw pointer and length in MCU local SRAM into an [`AxiDmaTarget`].
+    pub fn from_mcu_sram_ptr(ptr: *const u8, len: usize) -> McuResult<Self> {
```
clundin25: The DMA engine requires word alignment, we should check or change the parameters to guarantee this

clundin25: I think we should also consider disallowing sending a length of 0, but it would be good to check how the DMA engine would behave

runtime/userspace/api/caliptra-api/src/dma.rs
clundin25, line 1: There seems to be an existing DMA abstraction, can we use that?

runtime/userspace/api/caliptra-api/src/dma.rs
arul-sri: Good catch! I added an AxiDmaConfig struct that holds the platform addresses (local_sram_base, axi_sram_base, and sram_size), with presets for EMULATOR (0x4000_0000, 1MB) and FPGA (0xA8C0_0000, 512KB).

As for the build-time solution, AxiDmaConfig::DEFAULT now automatically picks FPGA when building with '--features fpga' and EMULATOR otherwise. This keeps it in sync with the linker manifests and avoids having to thread extra parameters through measurement-api and its callers. We also added mcu_sram_to_axi_dma_with_config in case a caller wants to pass an explicit config.

runtime/userspace/api/caliptra-api/src/dma.rs
arul-sri: Added! We now check that sram_offset + max_size <= sram_size so the DMA transfer can never extend past the end of MCU SRAM.

runtime/userspace/api/caliptra-api/src/dma.rs
arul-sri: Makes sense. We now check that both the buffer start address and the buffer length are 4-byte aligned (local_addr % 4 == 0 and len % 4 == 0).

arul-sri: Done! We now reject transfers with len == 0 right away.

runtime/userspace/api/caliptra-api/src/dma.rs
arul-sri: I looked into this. The existing DMAMapping trait is for the MCU's internal DMA controller (for MCU-initiated copies like in image loading). This use case is different because it's Caliptra Core's DMA engine writing into MCU SRAM from across the subsystem bus.

Using AxiDmaConfig keeps this specific to Caliptra DMA without having to add generic trait parameters across measurement-api and spdm-lib. I also added a comment in dma.rs explaining the distinction between the two.

Feel free to correct me if I am wrong.

runtime/userspace/api/caliptra-api/src/dma.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Maximum buffer size in bytes available for DMA transfer.
+    pub max_size: u32,
+}
+
+impl AxiDmaTarget {
+    /// Translates an MCU local SRAM buffer into an [`AxiDmaTarget`] using the default platform configuration.
+    #[inline]
+    pub fn from_mcu_sram(buf: &[u8]) -> McuResult<Self> {
+        AxiDmaConfig::DEFAULT.translate(buf)
+    }
```
clundin25, line 134: Let's drop having a "default" configuration. Ideally we should have compiler errors if there is something platform specific we need to implement.

runtime/userspace/api/caliptra-api/src/dma.rs
```diff
@@ -0,0 +1,259 @@
...
+pub fn mcu_sram_to_axi_dma(buf: &[u8]) -> McuResult<(u32, u32)> {
+    AxiDmaTarget::from_mcu_sram(buf).map(|t| t.as_tuple())
+}
+
+/// Helper function to translate an MCU local SRAM buffer into an `(axi_addr, max_size)` tuple
+/// using an explicit configuration.
+#[inline]
+pub fn mcu_sram_to_axi_dma_with_config(buf: &[u8], config: &AxiDmaConfig) -> McuResult<(u32, u32)> {
+    AxiDmaTarget::from_mcu_sram_with_config(buf, config).map(|t| t.as_tuple())
+}
```
clundin25, line 161: I wonder if we should wrap these with a trait that is defined in the platform specific code, and passed to either the kernel or somewhere else.

runtime/userspace/api/caliptra-api/src/dma.rs
clundin25: I feel that maybe we should try to have a kernel module that can translate a userspace address to the AXI address so we can use these DMA APIs.

clundin25:
@mlvisaya can you take a look at this? We need a way to DMA from Caliptra to an allocated object

runtime/userspace/api/caliptra-api/src/dma.rs
mlvisaya: 1. I think it's the same as DMAMapping:

pub trait DMAMapping: Send + Sync {
    /// Convert a local address in MCU SRAM to an AXI address addressable by the MCU DMA controller.
    fn mcu_sram_to_mcu_axi(&self, addr: u32) -> Result<AXIAddr, ErrorCode>;
    /// Convert a Caliptra AXI address to the MCU DMA accessible address.
    fn cptra_axi_to_mcu_axi(&self, addr: AXIAddr) -> Result<AXIAddr, ErrorCode>;
}
Can you check if that can be used?

2. We also don't want to put platform-specific code under runtime/* (code under here are part of the SDK and should be platform agnostic), i.e. it should not be aware if it's FPGA, emulator, etc. We can probably follow the pattern for image_loader where the DMA address map is defined in the platforms/* and we pass it to the API layer (https://github.com/chipsalliance/caliptra-mcu-sw/blob/31de9cf62dce0aceb4414b84c79bb0651c4bca78/platforms/emulator/runtime/userspace/apps/user/src/image_loader/mod.rs#L252). Ideally DMA address translation should be provided as an interface by kernel and initialized by platform's board.rs

arul-sri: I checked DMAMapping:

Currently, mcu_sram_to_mcu_axi is implemented for the MCU's internal CDMA controller, so on emulator it returns addr as AXIAddr (0x4000_xxxx). However, Caliptra Core's DMA engine accesses MCU SRAM over the Subsystem AXI bus at 0xA8C0_0000. If we pass 0x4000_xxxx to Caliptra DMA on emulator, it fails.

I agree that runtime/* should not contain platform-specific constants or cfg(feature = "fpga").

To follow the image_loader pattern, we can add fn mcu_sram_to_cptra_axi(&self, addr: u32) -> Result<AXIAddr, ErrorCode> to DMAMapping, implement it in EmulatedDMAMap and FpgaDMAMap under platforms/*, and pass the mapping down to the API.

Would you prefer adding mcu_sram_to_cptra_axi to DMAMapping like in image_loader, or should we introduce a kernel syscall for this translation?

runtime/userspace/api/caliptra-api/src/dma.rs
arul-sri: That makes sense to me. Having the kernel translate the userspace buffer using MCU_MEMORY_MAP (or passing a platform trait like DMAMapping does in image_loader) keeps the SDK clean and platform-agnostic.

As discussed in the comment below, we can either extend DMAMapping in platforms/* with mcu_sram_to_cptra_axi and pass it down, or add a kernel syscall. LMK which direction you'd prefer for this PR.

runtime/userspace/api/caliptra-api/src/dma.rs
mlvisaya: I think it would be cleaner to add it as a kernel syscall, so we can put the address mapping in board.rs, and just expose it through the dma syscall.

runtime/userspace/api/caliptra-api/src/dma.rs
arul-sri: Done. Completely removed AxiDmaConfig, the platform presets (EMULATOR, FPGA, DEFAULT), and the fpga feature flag. The address translation is now handled dynamically by the kernel driver via a syscall, so userspace no longer needs any compile-time defaults.

runtime/userspace/api/caliptra-api/src/dma.rs
arul-sri: Implemented!

- Added TRANSLATE_LOCAL_TO_CPTRA_AXI (command 5) to the kernel DMA capsule to validate word-alignment, bounds, non-zero length, and perform the AXI translation.
- Configured DmaComponent in both platforms/emulator/runtime/src/board.rs and platforms/fpga/runtime/src/board.rs using MCU_MEMORY_MAP and CALIPTRA_SRAM_AXI_BASE.
- Exposed mcu_sram_to_cptra_axi on userspace DMA<S>.
- Removed all platform-specific addresses and configs from runtime/*.

Verified with unit tests and the emulator integration test (test_get_ocp_lock_endorsement_cert_mldsa_cmd).

clundin25 (approved):
Great job! Thanks @arul-sri

PR #2067, opened 2026-09-03
caliptra-api: Add ML-KEM

embediver:
Adds commands to perform an ML-KEM key exchange through Caliptra:

1. `mlkem_key_gen` — generates an ML-KEM-1024 encapsulation key from a seed CMK.
2. `mlkem_encapsulate` — performs encapsulation against the encapsulation key,
    producing ciphertext and a shared secret CMK.
3. `mlkem_decapsulate` — performs decapsulation using the seed and ciphertext,
    recovering the shared secret CMK.

An emulator test was added for verification.

Signed-off-by: Marvin Gudel <marvin.gudel@9elements.com>
Assisted-by: Opencode:claude-sonnet-4-5

### DIFF
--- a/platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
+++ b/platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
@@ -6,10 +6,11 @@ use core::ptr::NonNull;
 use mcu_caliptra_api::{
     cm_hmac, cm_import, dpe_certify_key_pubkey, dpe_sign_ecc_p384, ecdh_finish, ecdh_generate,
     ecdsa_verify, hash_all, hkdf_expand, hkdf_extract, mldsa87_compute_mu, mldsa87_compute_tr,
-    rng_generate, sha_finish, sha_init, sha_update, spdm_aes_gcm_decrypt, spdm_aes_gcm_encrypt,
-    CmKeyUsage, HashAlgo, HkdfSalt, CERTIFY_KEY_MLDSA87_PUBKEY_SIZE,
-    CMB_ECDH_ENCRYPTED_CONTEXT_SIZE, CMB_ECDH_EXCHANGE_DATA_MAX_SIZE, DPE_LABEL_LEN,
-    DPE_MLDSA87_MU_SIZE, DPE_P384_SIGNATURE_SIZE, MLDSA87_TR_SIZE, SHA_CONTEXT_SIZE,
+    mlkem_decapsulate, mlkem_encapsulate, mlkem_key_gen, rng_generate, sha_finish, sha_init,
+    sha_update, spdm_aes_gcm_decrypt, spdm_aes_gcm_encrypt, CmKeyUsage, HashAlgo, HkdfSalt,
+    CERTIFY_KEY_MLDSA87_PUBKEY_SIZE, CMB_ECDH_ENCRYPTED_CONTEXT_SIZE,
+    CMB_ECDH_EXCHANGE_DATA_MAX_SIZE, DPE_LABEL_LEN, DPE_MLDSA87_MU_SIZE, DPE_P384_SIGNATURE_SIZE,
+    MLDSA87_TR_SIZE, MLKEM1024_CIPHERTEXT_SIZE, MLKEM1024_ENCAPS_KEY_SIZE, SHA_CONTEXT_SIZE,
 };
 
 const CRYPTO_SCRATCH_SIZE: usize = 16 * 1024;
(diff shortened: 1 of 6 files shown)


platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
```diff
@@ -102,6 +103,58 @@ pub async fn test_caliptra_ecdh(alloc: &BitmapAllocator) {
...
+        .unwrap_or_else(|_| test_exit(1));
+
+    println!("ML-KEM key exchange finished. Testing if keys match...");
+
+    // Verify both sides derived the same secret by using them with HMAC
+    let test_data = b"ML-KEM-1024 test data";
+    let mut mac1 = [0u8; 48];
+    let mut mac2 = [0u8; 48];
+
+    cm_hmac(alloc, &initiator_secret, test_data, &mut mac1)
```
embediver, line 137: CM_HMAC currently only allows 48 or 64 byte keys, ML-KEM-1024 keys are 32 byte.
That seems to be the cause of the test failure.
What is strange to me is, that the test is not reporting error but times out.

This should be fixed so that HMAC also supports 32 byte keys right?

embediver: https://github.com/chipsalliance/caliptra-sw/blob/068128e6f16c00f70627ed5d6780a2435dc4d6a5/common/src/hmac_cm.rs#L48

zhalvorsen: Hmm, can you reproduce that hang in a `caliptra-sw` test? The HMAC hardware only supports HMAC-384 and HMAC-512. It should just be returning an error if it is smaller than that though https://github.com/chipsalliance/caliptra-sw/blob/main/common/src/hmac_cm.rs#L36.

As discussed offline this would be easier to test using an AES-256 key

runtime/userspace/api/caliptra-api/src/ml_kem.rs
```diff
@@ -0,0 +1,233 @@
...
+const KEY_GEN_REQ_SIZE: usize = size_of::<MlKemKeyGenReq>();
+
+/// Key generation response: `chksum(4) + fips(4) + encaps_key(1568)`.
+const KEY_GEN_RSP_SIZE: usize = MBOX_RESP_HEADER_SIZE + CMB_MLKEM_ENCAPS_KEY_SIZE;
+
+/// Encapsulate request prefix: `chksum(4) + key_usage(4)`, then
+/// `encaps_key(1568)`.
+#[repr(C)]
+#[derive(FromBytes, IntoBytes, KnownLayout, Immutable, Unaligned)]
+struct MlKemEncapsulateReqPrefix {
```
zhalvorsen, line 58: I know this isn't common practice in this library, but can we reference the structs from `caliptra-sw` directly so we don't worry about bit rot in these definitions? Or connect them with the `caliptra-sw` definitions with compile time checks using `size_of` or `offset_of!`?

See https://github.com/chipsalliance/caliptra-sw/blob/main/api/src/mailbox.rs#L5038 for definitions

embediver: Absolutely, actually prefer it and just did it to mirror the rest of the library.

platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
embediver: This is just a MCU thing, the kernel driver even prints the mailbox error but it doesn't seam to reach the userspace which should exit the test.
Sorry if that was to unclear.

runtime/userspace/api/caliptra-api/src/ml_kem.rs
embediver: Updated it to only use the definitions from caliptra-sw.

platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
embediver: I changed the test to use AES-GCM but that is also not working out of the box since the keys need to go through a HKDF first to have the correct size.
Should we skip the verification of the CMKs in this test (for now) and only verify that the ML-KEM mailbox commands are successful?

zhalvorsen (approved):

platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
embediver: I was curious to see if using a ML-KEM CMK for HKDF would then work and as it seem we have the same limitation there even though that should not be the case I believe.

See here: https://github.com/chipsalliance/caliptra-sw/blob/main/runtime/src/cryptographic_mailbox.rs#L2020 

Documentation states that keys that get imported should be padded to 48 or 64 bytes, but we ~~don't have that option with an encrypted CMK~~ of course can't do that with the negotiated 32 byte ML-KEM key.

parvathib (approved):

platforms/emulator/runtime/userspace/apps/example/src/test_caliptra_crypto.rs
zhalvorsen: > Should we skip the verification of the CMKs in this test (for now) and only verify that the ML-KEM mailbox commands are successful?

That's fine for now, we can add more tests in a follow on

PR #2072, opened 2026-09-03
[spdm] trim transport padding before hashing requests into M1/VCA

TimMarkey:
## Problem
Transport layers (e.g. PCIe VDM) may pad SPDM requests to DWORD granularity. The responder currently hashes the **raw padded request** into the VCA and M1 transcripts. The requester hashes only the exact SPDM message bytes, so the responder's transcript diverges from the requester's and `CHALLENGE_AUTH` signature verification fails once any request arrives padded.

## Fix
Trim each request to its exact SPDM message length before appending it to VCA/M1 in the `spdm-lib` stack:

| Message | Trim length |
|---|---|
| `GET_VERSION` / `GET_DIGESTS` | header + 2 param bytes |
| `GET_CAPABILITIES` | header + `CapabilitiesBody` (18 bytes) |
| `NEGOTIATE_ALGORITHMS` | the `Length` field carried in the fixed body (DSP0274 Table 18) |
| `GET_CERTIFICATE` | header + standard (6) or large (14) request body |
| `CHALLENGE` | header + `ChallengeReqBody` (+ `RequesterContext` in 1.3+) |

The response side already trims via `spdm_len`; this brings the request side to parity.

Also gates the `DIGESTS` `SupportedSlotMask` (Param1) to SPDM 1.3+ per DSP0274 Table 35 — in 1.2 and earlier Param1 is Reserved and shall be 0.

## Why the other transcript-feeding requests don't need this change
This PR covers exactly the request handlers that were appending **raw** request bytes to a transcript. Every other request that contributes to a hashed/signed transcript already trims to the exact SPDM message length before appending, so no change is needed:

| Transcript | Request | Status |
|---|---|---|
| L1 (signed `GET_MEASUREMENTS`) | `GET_MEASUREMENTS` | already trims — `measurements.rs` appends `&req[..spdm_req_len]` |
| TH (`KEY_EXCHANGE_RSP` signature) | `KEY_EXCHANGE` | already trims — `key_exchange.rs` appends `req[..spdm_req_len]` ("use only the actual SPDM bytes, not transport padding") |
| TH (`FINISH` HMAC) | `FINISH` | already trims — `finish.rs` appends `&spdm_msg[..finish_prefix_len]` (and `req_verify_data` is an exact-length slice) |

The in-session `GET_DIGESTS` / `GET_CERTIFICATE` / `GET_MEASUREMENTS` paths dispatch into the **same** `handle_*_req` functions fixed here, so they inherit the trim automatically.

Everything else does not contribute to a signed transcript, so request padding is irrelevant to it: `SET_CERTIFICATE`, `VENDOR_DEFINED_REQUEST`, `END_SESSION`, and `CHUNK_GET` / `CHUNK_SEND` (the reassembled request flows back through the trimmed handlers). `PSK_EXCHANGE` / `PSK_FINISH` are not implemented in this stack yet; when added, their request appends will need the same trim.

## Testing
`cargo check -p caliptra-mcu-spdm-stack` passes cleanly. Verified against the Micron Caliptra Plato SPDM responder + DMTF `spdm-emu` requester with a VDM transport that pads requests: CHALLENGE_AUTH now verifies across SPDM 1.2/1.3.

### DIFF
--- a/runtime/userspace/api/spdm-lib/stack/src/algorithms.rs
+++ b/runtime/userspace/api/spdm-lib/stack/src/algorithms.rs
@@ -103,7 +103,13 @@ pub(crate) async fn handle_negotiate_algorithms<'a, Pal: SpdmPal>(
 
     // SPDM: NEGOTIATE_ALGORITHMS + ALGORITHMS contribute to VCA.
     let head = pal.header_size();
-    state.transcript.append_vca(pal, io, io.request()).await?;
+    // Trim transport padding (e.g. PCIe-VDM DWORD alignment): NEGOTIATE_ALGORITHMS
+    // carries its own total request length (including the SPDM header) in the
+    // fixed body's Length field (DSP0274 Table 18); feed exactly that to VCA so
+    // it matches what the requester hashes.
+    let req_len = fixed.length.get() as usize;
+    let req_msg = req.get(..req_len).ok_or(SPDM_INVALID_REQUEST)?;
+    state.transcript.append_vca(pal, io, req_msg).await?;
     state
         .transcript
         .append_vca(pal, io, &resp[head..head + spdm_len])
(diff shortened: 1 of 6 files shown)


TimMarkey:
Please review
cc @parvathib @RaunakGu @ArthurHeymans @amit-kumarh

Issue #2001, opened 2026-08-18
[Tests]: Fake runtime image


clundin25:
It's unclear that you've successfully recovered to the correct MCU FW image because the fake image doesn't do anything. It would be nice to have a minimal image that emits some UART logs to confirm that it was successfully booted.

PR #2086, opened 2026-09-10
Adding start up message to dummy rom-stub-firmware

v-focutter:
Adding a simple logging message to the dummy firmware rom so will confirm that it was started. 

Task: https://github.com/chipsalliance/caliptra-mcu-sw/issues/2001


xtask/src/rom.rs
```diff
@@ -2,7 +2,24 @@
...
+    u32::to_le(0x0000_0317), // auipc t1, 0
+    u32::to_le(0x0203_0313), // addi t1, t1, 32 (offset from auipc to the string)
+    u32::to_le(0x0003_4383), // lbu t2, 0(t1)
+    u32::to_le(0x0003_8863), // beq t2, zero, 16
+    u32::to_le(0x0072_8023), // sb t2, 0(t0)
+    u32::to_le(0x0013_0313), // addi t1, t1, 1
+    u32::to_le(0xff1f_f06f), // jal zero, -16
+    u32::to_le(0x0000_006f), // jal zero, 0
+];
+static ROM_STUB_START_MESSAGE: [u8; 32] = *b"[mcu-runtime] ROM stub started\n\0";
```
clundin25, line 22: I don't think this is related to ROM. Maybe we can instead say:

`boot sequence complete`

clundin25 (approved):

PR #2089, opened 2026-09-10
spdm: reject LargeCertChain GET_CERTIFICATE without LARGE_RESP_CAP

TimMarkey:
## Summary

A `GET_CERTIFICATE` carrying the `LargeCertChain` request bit (param1) is only defined in SPDM 1.4+ and only when the responder advertised `LARGE_RESP_CAP`. The handler previously gated that bit on the negotiated version alone, so a 1.4 requester that set `LargeCertChain` against a responder which never advertised `LARGE_RESP_CAP` was serviced with a normal `CERTIFICATE` response instead of being rejected.

This makes the handler return `ERROR(UnsupportedRequest)` when `is_large()` is set at 1.4+ without `LARGE_RESP_CAP` in the advertised capabilities. The `< 1.4` path continues to return `ERROR(InvalidRequest)` (the param1 bit is reserved there).

## Conformance

Fixes the `SPDM-Responder-Validator` conformance case `spdm_test_case_certificate_large_invalid_request` (test 5.9), which currently fails against this responder:

```
test assertion 5.9.2 - FAIL response code - 0x02   (got CERTIFICATE, expected ERROR 0x7f)
test assertion 5.9.3 - FAIL response param1 - 0x80  (expected UnsupportedRequest)
```

The suite recently un-skipped this case, so it now fails on every open main-2.1 PR that runs `spdm_validator_tests`.

## Testing

- `cargo test -p caliptra-mcu-spdm-stack` — 40 passed, incl. new `test_get_certificate_v14_large_without_large_resp_cap_returns_unsupported`
- `cargo fmt --check` clean

### DIFF
--- a/runtime/userspace/api/spdm-lib/stack/src/certificate.rs
+++ b/runtime/userspace/api/spdm-lib/stack/src/certificate.rs
@@ -11,7 +11,7 @@
 //! stack-allocated `[u8; N]` array for cert payload.
 
 use caliptra_mcu_spdm_codec::{
-    CertificateLargeRsp, CertificateLargeRspBody, CertificateRsp, CertificateRspBody,
+    CapFlags, CertificateLargeRsp, CertificateLargeRspBody, CertificateRsp, CertificateRspBody,
     GetCertificateParam1, GetCertificateReq, ReqRespCode, ResponseBody, SpdmMsgHdrPdu, SpdmVersion,
     WireWriter,
 };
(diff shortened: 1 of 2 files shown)


TimMarkey:
Please review
cc @parvathib @RaunakGu @ArthurHeymans @amit-kumarh

amit-kumarh (approved):
LGTM. Thanks for the catch!

parvathib (approved):

PR #2094, opened 2026-09-11
[spdm] add support for VDM large fields

### LINKED ISSUE #2075: Complete SPDM 1.4 large certificate-chain support
The current certificate path is limited to  u16::MAX .  MANAGED_MAX_DER_LEN  limits only the stored endorsement DER, while the returned chain also includes the SPDM header, DPE chain, and generated leaf certificate. The composed chain can therefore exceed the current 16-bit encoding limit.

Complete large certificate-chain support end to end:

- Use the SPDM 1.4 four-byte certificate-chain  Length ; retain  Length(u16) + Reserved(u16)  for SPDM 1.3 and earlier.
- Gate large fields on SPDM 1.4 and the negotiated  LARGE_RESP_CAP .
- Complete  LargeOffset ,  LargeLength ,  LargePortionLength , and  LargeRemainderLength  handling for  GET_CERTIFICATE .
- Update certificate hashing to use the same version-appropriate chain header as the emitted chain.
- Support the SPDM 1.4 large-length representation in  SET_CERTIFICATE .
- Retain checked arithmetic and respect platform storage, transfer, and buffering limits rather than assuming the full  u32  range is available.

Expected behavior:

- A standard SPDM 1.4 request for a chain larger than  u16::MAX  receives  DataTooLarge  with  ActualSize , and retrying with large fields succeeds.
- SPDM 1.3 and earlier never truncate an oversized chain.
- Enable the tests related to large fields in validator GET_CERTIFICATE test group and they should run successfully.

amit-kumarh:
This now allows us to advertise LARGE_RESP capabilities. Addresses #2075

### DIFF
--- a/runtime/userspace/api/spdm-lib/codec/src/lib.rs
+++ b/runtime/userspace/api/spdm-lib/codec/src/lib.rs
@@ -73,8 +73,8 @@ pub use vendor_defined::pci_sig::ide_km::{
 };
 pub use vendor_defined::pci_sig::PciSigProtocolHdr;
 pub use vendor_defined::{
-    decode_vendor_defined_req, StandardsBodyId, VendorDefinedReq, VendorDefinedReqPdu,
-    VendorDefinedRspBody, VendorDefinedRspPdu,
+    decode_vendor_defined_req, StandardsBodyId, VendorDefinedParam1, VendorDefinedReq,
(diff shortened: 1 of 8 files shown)


zhalvorsen (approved):
LGTM!

PR #2021, opened 2026-08-25
Handover firmware boot type and ROM capabilities to RT

mlvisaya:
## Summary

Add a versioned ROM-to-Runtime handoff for the MCU firmware boot source and MCU ROM capabilities. Runtime uses this information to select the appropriate image loader and report ROM capabilities through the common `DeviceCapabilities` command.

## Changes

- Extend the handoff ABI to version 1.3 while preserving its existing size and layout constraints.
- Add firmware boot-source values:
  - `Unknown`
  - `Flash`
  - `Streaming`
  - `Network`
- Record the provider that successfully loaded the firmware, including fallback-provider selection.
- Select the Runtime image loader dynamically when both flash and streaming loaders are enabled.
- Select the only available loader at compile time for single-loader builds.
- Preserve compatibility with older ROM handoffs:
  - Pre-v1.2 handoffs use the legacy Streaming loader in dual-loader builds.
  - Invalid boot-source values from v1.2 or newer remain errors.
- Add MCU ROM capability flags:
  - `STREAMING_BOOT_I3C`
  - `FLASH_BOOT`
  - `NETWORK_BOOT`
- Derive ROM capabilities from the configured image providers.
- Expose the boot source and ROM capabilities through System capsule commands.
- Populate `DeviceCapabilities.mcu_rom` using the handoff capability bitmap in big-endian format.
- Report zero ROM capabilities for pre-v1.3 handoffs.
- Update emulator and FPGA handoff checks, documentation, and integration tests.

## Handoff Compatibility

The handoff table remains fixed at 64 bytes, and the complete handoff data remains 260 bytes.

| Handoff Version | Behavior |
| --- | --- |
| Before 1.2 | Boot source unavailable; Runtime uses the legacy Streaming fallback |
| 1.2 | Firmware boot source available |
| 1.3 | Firmware boot source and MCU ROM capabilities available |

Unknown capability bits are ignored, and reserved bits are reported as zero.

PR #2095, opened 2026-09-11
Advertise additional MCU ROM capabilities

mlvisaya:
## Summary

- advertise OCP LOCK, firmware DOT manifest, component SVN manifest, stable owner key, DOT boot, and DOT recovery capabilities
- centralize boot-provider capability detection in the cold-boot capability builder
- document the new capability bits and extend unit and integration coverage

Stacked on #2021. The base should be changed to `main-2.1` after #2021 merges.

## Validation

- `cargo +1.95 xtask precheckin`
- `cargo +1.95 test -p caliptra-mcu-tests-integration test_device_capabilities_cmd -- --nocapture`
- ROM capability tests with default, `stable-owner-key`, and `ocp-lock` feature configurations


rom/src/cold_boot.rs
```diff
@@ -169,6 +169,208 @@ const GCM_TAG_SIZE: usize = 16;
...
+        capabilities |= McuRomCapabilities::STABLE_OWNER_KEY;
+    }
+    let dot_boot_enabled = params.dot_flash.is_some()
+        && params.owner_pk_hash_policy
+            == crate::device_ownership_transfer::OwnerPkHashPolicy::DotThenFuse;
+    if dot_boot_enabled {
+        capabilities |= McuRomCapabilities::DOT_BOOT;
+    }
+    if dot_boot_enabled && !params.dot_locked_recovery_handlers.is_empty() {
+        capabilities |= McuRomCapabilities::DOT_LOCKED_RECOVERY;
```
helloxiling: Should we include `params.dot_recovery_reset_flow` in the `DOT_LOCKED_RECOVERY` predicate?

rom/src/cold_boot.rs
```diff
@@ -169,6 +169,208 @@ const GCM_TAG_SIZE: usize = 16;
...
+    }
+    if params.dot_flash.is_some()
+        && params
+            .i3c_services
+            .is_some_and(|services| services.contains(I3cServicesModes::DOT_RECOVERY))
+    {
```
helloxiling: Could we derive `I3C_DOT_RECOVERY` from an actually reachable I3C recovery path rather than `i3c_services` alone?

rom/src/cold_boot.rs
```diff
@@ -169,6 +169,208 @@ const GCM_TAG_SIZE: usize = 16;
...
+    if cfg!(feature = "stable-owner-key") {
+        capabilities |= McuRomCapabilities::STABLE_OWNER_KEY;
+    }
+    let dot_boot_enabled = params.dot_flash.is_some()
+        && params.owner_pk_hash_policy
+            == crate::device_ownership_transfer::OwnerPkHashPolicy::DotThenFuse;
+    if dot_boot_enabled {
+        capabilities |= McuRomCapabilities::DOT_BOOT;
+    }
+    if dot_boot_enabled && !params.dot_locked_recovery_handlers.is_empty() {
```
helloxiling: Should we include `params.dot_recovery_reset_flow` in the `DOT_LOCKED_RECOVERY` predicate?

mlvisaya: Addressed in c68300e15. DOT_LOCKED_RECOVERY is now advertised when DOT boot is enabled and either dot_recovery_reset_flow is enabled or a locked-state handler is configured. Added focused coverage for the reset-flow-only case.

rom/src/cold_boot.rs
mlvisaya: Addressed in c68300e15. I3C_DOT_RECOVERY is now derived from a reachable path: forced I3C DOT services, or a configured locked-state handler that explicitly reports I3C DOT recovery support. The I3C handler implements that capability; other handlers default to false. Added focused coverage for both paths.

helloxiling (approved):
The latest patch looks good. Please check the CI failures before the merge.

PR #2097, opened 2026-09-14
[caliptra-api] Add owner authorization manifest mailbox support

parvathib:
Add the MCU-side Caliptra API surface for the Owner Authorization Manifest commands.

- SET_OWNER_AUTH_MANIFEST: submit a staged manifest as chksum(4) | manifest_size(4) | manifest bytes, sent as a header plus contiguous payload so only manifest_size bytes reach the wire and the caller's immutable staged buffer is the buffer Caliptra authenticates. The layout is anchored to SetOwnerAuthManifestReq with const asserts.
- AUTHORIZE_AND_STASH: accept IMAGE_AUTHORIZED_OWNER_ONLY behind a new opt-in accept_owner_only parameter, so components expected under an installed owner manifest succeed while base components still reject an unexpected owner-only result.


zhalvorsen (approved):

mhatrevi (commented):
Could we validate that `rsp_len == MBOX_RESP_HEADER_SIZE` before returning success? As written, a truncated response could be treated as a successful manifest installation.

```suggestion
    let rsp_len = crate::wire::mbox_execute_slice(
        CMD_SET_OWNER_AUTH_MANIFEST,
        Some(header.as_bytes()),
        manifest,
        &mut rsp,
    )
    .await?;

    if rsp_len != MBOX_RESP_HEADER_SIZE {
        return Err(mcu_error::codes::INTERNAL_BUG);
    }

    Ok(())
```

If `execute_with_payload_slice` already guarantees the exact response length, could we document that invariant here instead?

parvathib:
> Could we validate that `rsp_len == MBOX_RESP_HEADER_SIZE` before returning success? As written, a truncated response could be treated as a successful manifest installation.
> 
> ```
>     let rsp_len = crate::wire::mbox_execute_slice(
>         CMD_SET_OWNER_AUTH_MANIFEST,
>         Some(header.as_bytes()),
>         manifest,
>         &mut rsp,
>     )
>     .await?;
> 
>     if rsp_len != MBOX_RESP_HEADER_SIZE {
>         return Err(mcu_error::codes::INTERNAL_BUG);
>     }
> 
>     Ok(())
> ```
> 
> If `execute_with_payload_slice` already guarantees the exact response length, could we document that invariant here instead?

Resp len guard is in place now

mhatrevi (approved):

PR #2100, opened 2026-09-14
[spdm] serve per-algorithm chains from managed cert slots

### LINKED ISSUE #2098: Support ML-DSA certificates in managed slots
Managed slots (1 and 2) are hardcoded to ECC-only. Expand them to support ECC and MLDSA banks, similarly to the ReadOnly slot 0.

amit-kumarh:
Managed certificate slots previously only stored a single ECC chain and had no notion of the MLDSA negotiation, which means it was possible to serve ECC bytes in a MLDSA session, producing a garbage cert chain.

This PR separates those managed slots into an ECC bank and an MLDSA bank, like the read-only slot is today. 

Note that the new SPDM 1.4 Slot Management commands are not included - only the currently negotiated algorithm can be managed through the standard commands (GET_CERTIFICATE / SET_CERTIFIACTE).


platforms/emulator/runtime/userspace/apps/user/src/cert_store/mod.rs
```diff
@@ -146,20 +161,23 @@ async fn setup_endorsements<A: ApiAlloc>(store: &SharedCertStore, alloc: &A) ->
...
         {
+            let ecc_base = slot_index * MANAGED_REGIONS_PER_SLOT * MANAGED_SLOT_REGION_SIZE;
             store
                 .set_managed_endorsement(
                     store_slot,
                     spdm_slot,
                     CERT_STORE_PARTITION.driver_num,
-                    region * MANAGED_SLOT_REGION_SIZE,
+                    ecc_base,
+                    Some(ecc_base + MANAGED_SLOT_REGION_SIZE),
```
zhalvorsen: need to take a closer look at this

runtime/userspace/api/spdm-lib/stack/src/set_certificate.rs
```diff
@@ -443,20 +444,49 @@ fn validate_negotiated_set_certificate_algorithms<S, L>(
...
     }
-    if state.negotiated_base_asym_sel != AsymAlgos::ECDSA_ECC_NIST_P384 {
+    // Exactly one signing algorithm is selected per connection: in 1.4
+    // a PQC selection clears `base_asym_sel`. Managed slots back each
+    // algorithm with its own flash region, so either is installable.
+    let ecc = state.negotiated_base_asym_sel == AsymAlgos::ECDSA_ECC_NIST_P384;
+    let mldsa = state
+        .negotiated_pqc_asym_sel
+        .contains(PqcAsymAlgos::ML_DSA_87);
+    if !ecc && !mldsa {
```
zhalvorsen, line 454: It sounds like this should never be the case, but should this check if both are true?

runtime/userspace/api/spdm-lib/stack/src/certificate.rs
```diff
@@ -29,13 +29,46 @@ use crate::error::{
...
+
+/// Encode the leading `Length` field of the SPDM cert-chain header.
+///
+/// Always a little-endian `u32`; for pre-1.4 chains the upper two
+/// bytes are zero, which is exactly the required `Reserved` encoding.
+///
+/// Every producer of the chain header must use this so that the bytes
+/// served by GET_CERTIFICATE and the bytes hashed for GET_DIGESTS
+/// agree.
+pub(crate) fn cert_chain_length_field(total_len: usize) -> mcu_error::McuResult<[u8; 4]> {
```
zhalvorsen, line 84: Should this take an `SpdmVersion` parameter and check the size accordingly?

runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
```diff
@@ -97,32 +97,101 @@ impl SharedCertStore {
...
+        // to a different slot.
+        let ecc_end = base
+            .checked_add(capacity)
+            .ok_or(mcu_error::codes::INVARIANT)?;
+        let mldsa_range = match mldsa_base {
+            Some(mldsa_base) => {
+                let mldsa_end = mldsa_base
+                    .checked_add(capacity)
+                    .ok_or(mcu_error::codes::INVARIANT)?;
+                if base < mldsa_end && mldsa_base < ecc_end {
```
zhalvorsen: nit: It could be worthwhile to add a common helper that takes two `Range`s and checks if they overlap then you could leverage the unit tests from `runtime/userspace/api/spdm-lib/pal/src/cert/endorsement.rs` for all uses

runtime/userspace/api/spdm-lib/stack/src/set_certificate.rs
amit-kumarh: I think we're good here - we (Caliptra responder) control what is the ultimate negotiated algorithm field, and that path is tested to comply with the spec of only allowing a single negotiated algorithm, so that should be an unreachable path.

runtime/userspace/api/spdm-lib/stack/src/certificate.rs
amit-kumarh: The wire representation is the same, as the u32 length consumes the two bytes that were previously reserved - (u32_lower, u32_upper) vs (u16, 0x00).

This clause in `handle_get_certificate` should handle the case where we're requesting a large cert chain without setting the large bit.

```
    let max_chain_len = max_cert_chain_len(state.version).min(req.max_length_cap());
    if total_len_usize > max_chain_len {
        if state.version >= SpdmVersion::V14 {
            let actual_size = u32::try_from(total_len_usize).map_err(|_| SPDM_UNSPECIFIED)?;
            return Err(SPDM_DATA_TOO_LARGE.with_extended_data(actual_size.to_le_bytes()));
        }
        return Err(SPDM_UNSPECIFIED);
    }
```

Otherwise, I don't think we want to prevent, for example, GET_DIGESTS -> CHALLENGE from failing just because the cert chain is > 64K, even if we don't plan on fetching it in its entirety?

runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
amit-kumarh: good idea, done

runtime/userspace/api/spdm-lib/pal/src/cert/endorsement.rs
```diff
@@ -47,30 +49,49 @@ pub const fn slot_index(slot_id: u8) -> Option<usize> {
...
 impl CertSlot {
     pub const fn empty() -> Self {
         Self {
             endorsement: SlotEndorsement::Empty,
-            key_pair_id: None,
-            cert_info: None,
-            write_in_progress: AtomicBool::new(false),
+            key_pair_id: [None; NUM_ASYM_ALGOS],
+            cert_info: [None; NUM_ASYM_ALGOS],
+            write_in_progress: [AtomicBool::new(false), AtomicBool::new(false)],
```
parvathib: Nit: could we use  NUM_ASYM_ALGOS  in the initializer too, so it stays aligned with the array size if more algorithms are added?

platforms/emulator/runtime/userspace/apps/user/src/cert_store/mod.rs
```diff
@@ -73,11 +77,22 @@ static CERT_STORE: SharedCertStore = SharedCertStore::new();
...
+/// Region size. The partition is statically sized, so subdividing it
+/// four ways instead of two costs no additional flash. ML-DSA chains
+/// are roughly an order of magnitude larger than ECC ones, so the
+/// remaining room per region still far exceeds what either needs.
 #[cfg(feature = "test-mctp-spdm-set-certificate")]
 const MANAGED_SLOT_REGION_SIZE: usize = {
-    assert!(MANAGED_SLOT_COUNT != 0);
-    assert!(CERT_STORE_PARTITION.size % MANAGED_SLOT_COUNT == 0);
-    CERT_STORE_PARTITION.size / MANAGED_SLOT_COUNT
+    assert!(MANAGED_REGION_COUNT != 0);
```
parvathib: Should we use separate logical partitions for each managed  (slot, algorithm)  in the reference implementation? Since the layout is platform-owned, this could simplify the cert-store setup and avoid managing subregion offsets here

runtime/userspace/api/spdm-lib/pal/src/cert/endorsement.rs
```diff
@@ -308,15 +400,21 @@ pub struct ManagedEndorsement {
 }
 
 #[cfg(feature = "set-certificate")]
-impl ManagedEndorsement {
-    pub const fn new(slot: u8, driver_num: u32, base: usize, capacity: usize) -> Self {
+impl SingleManagedChain {
```
parvathib: Nit (naming): Consider renaming  `SingleManagedChain`  to  `SingleManagedEndorsement`  (and  `get_chain / set_chain`  to  `get_endorsement / set_endorsement` ). In Caliptra’s AliasCert model, the flash partition stores the endorsement part of cert chain rather than the complete Cert chain, and  `SingleManagedEndorsement` mirrors  `ReadOnlyEndorsement`  more cleanly.

runtime/userspace/api/spdm-lib/pal/src/cert/endorsement.rs
```diff
@@ -287,13 +361,31 @@ pub const fn managed_endorsement_der_capacity(region_size: usize) -> usize {
...
 }
 
+/// Whether two half-open byte ranges intersect.
+///
+/// The single definition of "these managed flash regions collide".
+/// Updating a chain erases its whole region, so any intersection at
+/// all — partial, exact or containing — is destructive; only abutting
+/// ranges are safe. Empty ranges never overlap anything.
+#[cfg(feature = "set-certificate")]
+pub fn ranges_overlap(a: &Range<usize>, b: &Range<usize>) -> bool {
```
parvathib: Could we guard against empty ranges explicitly? An interior empty range currently returns `true` (e.g. `ranges_overlap(&(5..5), &(0..10))` evaluates to `true`), even though the doc comment mentions empty ranges shouldn't overlap:

```
pub fn ranges_overlap(a: &Range<usize>, b: &Range<usize>) -> bool {
    !a.is_empty() && !b.is_empty() && a.start < b.end && b.start < a.end
}

File: runtime/userspace/api/spdm-lib/pal/src/cert/endorsement.rs [hunk 9]
```diff
@@ -611,6 +696,81 @@ impl ManagedEndorsement {
...
+        Self { ecc, mldsa: None }
+    }
+
+    pub const fn with_mldsa(mut self, mldsa: SingleManagedChain) -> Self {
+        self.mldsa = Some(mldsa);
+        self
+    }
+
+    /// The region serving `algo`, or `INVARIANT` if this slot has none.
+    pub fn get_chain(&self, algo: SpdmPalAsymAlgo) -> McuResult<&SingleManagedChain> {
```
parvathib: Could we make `ManagedEndorsement::get_chain` return `Option<&SingleManagedChain>` to mirror `ReadOnlyEndorsement::get_chain`? 
This keeps  `is_provisioned`  uniform across both variants

runtime/userspace/api/spdm-lib/pal/src/cert/store.rs
```diff
@@ -97,32 +97,102 @@ impl SharedCertStore {
 
         let slot = self.cert_slot_mut(idx).ok_or(mcu_error::codes::INVARIANT)?;
         slot.endorsement = SlotEndorsement::ReadOnly(ro);
-        slot.key_pair_id = Some(key_pair_id);
-        slot.cert_info = Some(DEFAULT_CERT_INFO);
+        // Both chains are provisioned together from the same key
```
parvathib, line 100: Ideally, both algorithms should share the same  `KeyPairID`  and  `CertificateInfo` per slot. Since each algorithm is provisioned independently, we currently don't guard against mismatches. No need to address this in this PR, but we should add validation checks in a follow-up to enforce this invariant.

parvathib (approved):
LGTM with some minor comments

platforms/emulator/runtime/userspace/apps/user/src/cert_store/mod.rs
amit-kumarh: Good idea, addressed.

PR #2101, opened 2026-09-14
[auth-cmds] Add SPDM-VDM support for OCP LOCK authorized commands

### LINKED ISSUE #1822: Support MC_OCP_LOCK_ROTATE_HEK in SPDM-VDM

### LINKED ISSUE #1823: Support MC_OCP_LOCK_SET_PERMA_HEK in SPDM-VDM

zhalvorsen:
This adds SPDM-VDM support for `MC_OCP_LOCK_ROTATE_HEK` and `MC_OCP_LOCK_SET_PERMA_HEK`. These commands were already supported in the MCI mailbox. This also adds support to `caliptra-util-host` for sending these commands.


clundin25 (approved):

parvathib:
Hi Zach! 
It looks like 0x12 is already used for Authorized commands, so there seems to be an overlap.
Could we consider having a top-level OCP_LOCK command with subcommands for the individual operations? This would also make it easier to add more operations later without breaking command compatibility.

zhalvorsen:
Yeah, I'll push another change today. I considered that, but didn't know if we would consider it breaking backwards compatibility with the previous mailbox release

PR #2104, opened 2026-09-15
Add caliptra utility host version API

mlvisaya:
Give caliptra-util-host an independent semantic version, expose it through the Rust API with a unit test, and remove the unused SPDM helper that blocked precheckin.


caliptra-util-host/apps/spdm/client/src/validator.rs
```diff
@@ -1223,38 +1223,6 @@ fn signed_fe_prog(
         .map_err(AuthorizedCommandError::Command)
 }
 
-fn send_raw_authorized_command(
```
mlvisaya, line 1228: this is unused code

parvathib (approved):

PR #2105, opened 2026-09-15
mcu_mbox: stage request metadata instead of copying the payload

parvathib:
The mailbox rx buffer maps mailbox SRAM directly, and the sender stays blocked until the command status is set, so a request that arrives while no app is listening is still intact in SRAM. Only the notification is lost, since the cmd-avail interrupt is cleared before dispatch.

Record `{command, dlen}` instead of copying the payload into a per-process grant buffer and read the payload back from SRAM via the new `Mailbox::map_rx_buffer()` when the app arms receive.

App grant drops from 9240 to 2 bytes per process.

Also drops the 9216-byte ceiling: requests larger than that were previously lost silently when no app was listening, hanging the sender until timeout.

Caveat: the staged request shares SRAM with the response buffer, so sending a response overwrites it. Send-response now returns BUSY while a request is staged. This is unlikely to be hit in practice since the mailbox follows a request/response model, where an app reads the request before responding.


caliptra-util-host/apps/spdm/client/src/validator.rs
```diff
@@ -1223,6 +1224,7 @@ fn signed_fe_prog(
         .map_err(AuthorizedCommandError::Command)
 }
 
+#[allow(dead_code)]
 fn send_raw_authorized_command(
```
mlvisaya, line 1228: nit: this function can be removed if not used

mlvisaya (approved):
LGTM

PR #2110, opened 2026-09-17
Add Owner Authorization Manifest builder support

mlvisaya:
Generate owner manifests from independent owner image metadata, expose create and parse commands, package them in flash images with a filename, and include them in SoC boot fixtures.


builder/src/caliptra.rs
```diff
@@ -947,6 +986,42 @@ fn main() -> Result<()> {
...
+                    ecc_priv_key: OWNER_ECC_KEY_PRIVATE,
+                    lms_priv_key: OWNER_LMS_KEY_PRIVATE,
+                    mldsa_priv_key: OWNER_MLDSA_KEY_PRIVATE,
+                }),
+            }
+        };
+        let gen_config = OwnerAuthManifestGeneratorConfig {
+            version: 1,
+            svn,
+            pqc_key_type: FwVerificationPqcKeyType::LMS,
```
parvathib, line 1016: I noticed that PQC key type is currently hardcoded to `LMS` independently when generating the `Caliptra firmware, base SoC manifest, and owner SoC manifest`. 
Since Caliptra uses the firmware’s PQC type to verify both manifests, these artifacts need to use the same type. Should we add `pqc_key_type` to  `CaliptraBuildArgs` and propagate that shared configuration to all three paths? 
We could still keep LMS as the default while allowing ML-DSA configurations.

builder/src/caliptra.rs
```diff
@@ -284,6 +302,27 @@ impl CaliptraBuilder {
         Ok(self.soc_manifest.clone().unwrap())
     }
 
+    /// Builds an Owner Authorization Manifest from `owner_soc_images`.
+    pub fn get_owner_auth_manifest(&mut self, name: Option<&str>) -> Result<PathBuf> {
```
parvathib, line 306: Should we validate the owner-manifest inputs before signing? For example, reject duplicate and out of range fw_ids.
This would prevent the builder from producing manifests that violate the documented namespaces or cannot be installed by Caliptra.

mlvisaya: Addressed in e3d6c0842. Owner manifest inputs are now validated before metadata generation/signing: fw_ids below the documented owner-only namespace (0x0001_0000..=0xFFFF_FFFF) are rejected, and duplicate owner fw_ids are rejected. Added boundary and duplicate regression coverage.

builder/src/caliptra.rs
mlvisaya: Addressed in e3d6c0842. CaliptraBuildArgs now carries a shared optional pqc_key_type (LMS by default), which is propagated through Caliptra firmware generation/re-signing, the base SoC manifest, and the Owner Authorization Manifest. Firmware cache names include the PQC type, and prebuilt firmware is checked for a type mismatch before manifests are generated. Added ML-DSA propagation and mismatch regression coverage.

parvathib (approved):

PR #2119, opened 2026-09-18
Install owner authorization manifest during image loading

mlvisaya:
Stream the packaged owner authorization manifest to Caliptra for flash and PLDM boot paths, and include it in feature-specific test images.


builder/src/all.rs
```diff
@@ -957,6 +982,7 @@ pub fn all_build(args: AllBuildArgs) -> Result<()> {
     let flash_image = create_flash_image(FlashImageBuildArgs {
         caliptra_fw_path: Some(caliptra_fw.clone()),
         soc_manifest_path: Some(soc_manifest.clone()),
+        owner_auth_manifest_path: None,
```
parvathib: Are we deliberately omitting the Owner Authorization manifest here? Is this going to create TOC entry?

builder/src/all.rs
parvathib: I think I'm using fw_id 3 for the test base soc images in the attestation tests. There'll be ID collision. Can we fix them in this PR?

mlvisaya: Addressed in b0c92833f. I moved the attestation fixture fw_ids from 3/4 to dedicated vendor-range IDs 0x2000/0x2001. The owner-only metadata remains in the owner namespace at 0x10000. I also added a regression test that checks the default, attestation, and owner fixture IDs are distinct and in their documented namespaces.

builder/src/all.rs
mlvisaya: Good catch. Addressed in b0c92833f: the base all-build path now generates a default Owner Authorization Manifest and packages it in flash_image.bin, so unconditional set_owner_auth_manifest() has TOC entry 3 available. I verified a generated base image has TOC IDs 0, 1, 2, and 3 and passes `cargo xtask flash-image verify`.

runtime/userspace/api/caliptra-api/src/image_loader/mod.rs
```diff
@@ -103,6 +110,51 @@ impl<T: DmaTransfer> ImageLoader for FlashImageLoader<'_, T> {
...
+        for byte in CommandId::SET_OWNER_AUTH_MANIFEST.0.to_le_bytes() {
+            checksum = checksum.wrapping_add(u32::from(byte));
+        }
+        for byte in req.as_bytes() {
+            checksum = checksum.wrapping_add(u32::from(*byte));
+        }
+        req.chksum = 0u32.wrapping_sub(checksum);
+
+        let response_buffer = &mut [0u8; core::mem::size_of::<MailboxRespHeader>()];
+        loop {
```
parvathib: Should we add `stream.reset();` at the start of each iteration here, matching `PldmImageLoader::set_owner_auth_manifest()` (line 301)?

runtime/userspace/api/caliptra-api/src/image_loader/mod.rs
```diff
@@ -205,6 +257,105 @@ impl<D: DMAMapping + 'static> ImageLoader for PldmImageLoader<'_, D> {
...
+        let (offset, size) = pldm_client::pldm_download_toc(OWNER_AUTH_MANIFEST_IDENTIFIER).await?;
+        if size == 0 || size as usize > OWNER_AUTH_MANIFEST_MAX_SIZE {
+            return Err(ErrorCode::Fail);
+        }
+
+        let mut stream = PldmMailboxPayloadStream::new(offset as usize, size as usize);
+        let mut checksum = 0u32;
+        let mut buffer = [0u8; pldm_context::PLDM_PAYLOAD_CHUNK_SIZE];
+        loop {
+            let bytes_read = stream.read(&mut buffer).await?;
```
parvathib: can we use `stream.get_bytesum()` here instead of computing the checksum manually?

platforms/emulator/runtime/userspace/apps/user/src/image_loader/mod.rs
```diff
@@ -274,6 +278,7 @@ async fn image_loading<D: DMAMapping>(
                 };
                 let pldm_image_loader =
                     PldmImageLoader::new(&fw_params, EXECUTOR.get().spawner(), dma_mapping);
+                pldm_image_loader.set_owner_auth_manifest().await?;
```
parvathib: If  `set_owner_auth_manifest()` fails, returning early with `? `, it won't call `pldm_image_loader.finalize() `  to clean up. This might the PLDM state machine hanging without reporting the failure to the Update Agent.

builder/src/all.rs
```diff
@@ -913,10 +942,13 @@ pub fn all_build(args: AllBuildArgs) -> Result<()> {
     }
 
     let mcu_image_cfg = get_image_cfg_feature(&mcu_cfgs.clone().unwrap_or_default(), "none");
+    let owner_soc_images =
+        create_default_owner_soc_images(effective_soc_images.as_deref().unwrap_or(&[]));
```
parvathib: This call sits directly in the base/production build path (`all_build`), which unconditionally adds dummy 128-byte `0xCC` entries into the Owner SoC manifest. Dummy owner images should only be added when building under test features (`FEATURES_REQUIRING_SOC_IMAGES`).

Also, `AllBuildArgs` doesn't currently accept `owner_soc_images` from the integrator. Is the intention to supply owner SoC images via the common TOML component config introduced in #2108?

In SDK production build, we should pass `owner_soc_images: None` (no owner images). So, when we build the default owner soc manifest, we shall add the owner Auth key, and no other components need to be present.

runtime/userspace/syscall/src/mailbox.rs
```diff
@@ -330,7 +330,27 @@ pub trait PayloadStream {
...
+    fn reset(&mut self);
+
     async fn read(&mut self, buffer: &mut [u8]) -> Result<usize, ErrorCode>;
+
+    /// Returns the wrapping sum of all payload bytes and rewinds the stream.
+    async fn get_bytesum(&mut self) -> u32 {
+        self.reset();
+        let mut sum = 0u32;
+        let mut buffer = [0u8; PAYLOAD_CHUNK_SIZE];
+        while let Ok(bytes_read) = self.read(&mut buffer).await {
```
parvathib: `while let Ok(bytes_read) = self.read(&mut buffer). await` silently terminates on read errors, returning wrong/truncated checksum.

Would it make sense to return `Result<u32, ErrorCode>` so that flash or PLDM I/O errors are propagated rather than masked into later mailbox checksum mismatches.

builder/src/all.rs
mlvisaya: A key-only Owner Authorization Manifest is the right production shape, but Caliptra SW does not currently accept it: `auth-manifest/gen/src/generator.rs::generate_owner()` rejects an empty metadata list, and `runtime/src/set_owner_auth_manifest.rs` rejects `entry_count == 0`. Removing the dummy entry only in this repository therefore makes both bundle generation and OAMN installation fail. I have kept the dummy entry for now and added a comment documenting the protocol constraint. We need a paired caliptra-sw change to allow/sign/accept a zero-entry metadata collection before this path can safely pass `owner_soc_images: None`. The common TOML integration for real owner images should be handled separately once that input contract is available.

runtime/userspace/syscall/src/mailbox.rs
mlvisaya: Addressed in 0a6c60adb. `PayloadStream::get_bytesum()` now returns `Result<u32, ErrorCode>`, propagates the original read error, and rewinds before returning on both success and failure. All callers now use `?`. I also added a unit test that verifies read-error propagation and both resets.

runtime/userspace/api/caliptra-api/src/image_loader/mod.rs
mlvisaya: Addressed in 0a6c60adb. The PLDM owner-manifest path now uses `stream.get_bytesum().await?`, removing the duplicate manual checksum loop and preserving I/O error propagation.

runtime/userspace/api/caliptra-api/src/image_loader/mod.rs
mlvisaya: Addressed in 0a6c60adb. The flash stream is now reset at the start of every mailbox attempt, matching the PLDM path and ensuring a Busy retry resends the full manifest.

platforms/emulator/runtime/userspace/apps/user/src/image_loader/mod.rs
mlvisaya: Addressed in 0a6c60adb. Owner-manifest installation and SoC-image loading now share the failure cleanup path: it reports `VerifyFailedFdSecurityChecks`, waits for the PLDM service to stop when finalize successfully signals an active service, and then returns the original error. The finalize guard avoids waiting forever if PLDM initialization itself failed.

parvathib (approved):

PR #2123, opened 2026-09-21
[ATTESTED CSR] support keypair inventory discovery in MCU and SPDM VDM

parvathib:
Update MCU firmware and host command types to support OCP Device Identity Provisioning (DIP) keypair inventory discovery (key_id == 0) and update caliptra-sw dependencies to merged revision 207835e3.

- Update caliptra-sw dependencies to latest to include DIP changes.
- Define DEVICE_KEY_ID_DISCOVERY (0x0000_0000) for requesting active keypair inventory without generating a CSR.
- Define algorithm constants CSR_ALGO_ECDSA_P384 (1) and CSR_ALGO_MLDSA_87 (2).
- Update ExportAttestedCsrReq wire structures and validation in caliptra-api to allow key_id == 0 discovery passthrough to Core.
- Add unit tests for key ID 0 discovery across mailbox messages, util-host command types, and SPDM VDM handlers.


mhatrevi (changes requested):
runtime/userspace/api/spdm-lib/vdm-handler/src/iana/ocp/caliptra_vdm/mod.rs
Lines: 2152-2167

This test uses a 20-byte response, so it does not exercise the actual ML-DSA-87 discovery response size. The pinned Caliptra implementation produces a 4,627-byte ML-DSA signature alone; after adding the COSE headers and inventory payload, the response exceeds the current 4 KiB limits in both the SPDM VDM and MCU mailbox paths.

This means `key_id = 0, algorithm = 2` is advertised and tested as supported, but cannot succeed end to end. Could we increase and synchronize the response-size limits, then update this test—and add an MCU mailbox equivalent—to use a realistic response larger than 4 KiB?

parvathib:
> runtime/userspace/api/spdm-lib/vdm-handler/src/iana/ocp/caliptra_vdm/mod.rs Lines: 2152-2167
> 
> This test uses a 20-byte response, so it does not exercise the actual ML-DSA-87 discovery response size. The pinned Caliptra implementation produces a 4,627-byte ML-DSA signature alone; after adding the COSE headers and inventory payload, the response exceeds the current 4 KiB limits in both the SPDM VDM and MCU mailbox paths.
> 
> This means `key_id = 0, algorithm = 2` is advertised and tested as supported, but cannot succeed end to end. Could we increase and synchronize the response-size limits, then update this test—and add an MCU mailbox equivalent—to use a realistic response larger than 4 KiB?

Good catch! I've updated the buffer limits across  mcu-mbox ,  vdm-handler, and host command types to 12.8 KiB (matching  caliptra-api ). Currently, this is feature-gated the expanded SPDM scratch pool is applicable under `set-certificate`  so baseline runtime memory remains untouched. However, it'd be ideal to chunk this response since it is a pass through instead of staging in MCU buffer.

mhatrevi (approved):

caliptra-util-host/command-types/src/certificate.rs
```diff
@@ -84,21 +84,32 @@ impl CommandResponse for SetCertificateResponse {}
 // ExportAttestedCsr Command
 // ============================================================================
 
-/// Maximum CSR data size (matches MAX_RESP_DATA_SIZE on MCU side)
-pub const MAX_CSR_DATA_SIZE: usize = 4 * 1024;
+/// Maximum CSR data size (matches MAX_ATTESTED_CSR_RESP_DATA_SIZE on MCU side)
+pub const MAX_CSR_DATA_SIZE: usize =
+    caliptra_mcu_mbox_common::messages::MAX_ATTESTED_CSR_RESP_DATA_SIZE;
```
ArthurHeymans: With this `ExportAttestedCsrResponse` becomes ~12.8 KiB, but `CaliptraSession` still has 8 KiB buffers (`MAX_COMMAND_PACKET_SIZE` in `caliptra-util-host/session/src/lib.rs`). Doesn't this make ExportAttestedCsr from the host fail for all key ids, also P-384? I see it failing on the emulator with `ocp_dev_identity_provision_tool`. Bumping the buffer to 16 KiB works for me, but not sure if that runs into memory budget issues somewhere.

PR #2124, opened 2026-09-21
Make minimum SVN command target-aware

mhatrevi:
## Summary
- rename `MC_FUSE_INCREASE_CALIPTRA_MIN_SVN` to `MC_FUSE_INCREASE_MIN_SVN` while retaining the existing `MCMS` command code (`0x4D43_4D53`), host command ID (`0x8013`), and authorized-subcommand capability bit (`2`)
- extend the request with a target selector and support Caliptra Runtime and SoC Manifest minimum SVN updates
- reserve Owner SoC Manifest as target `2`; requests for it currently return unsupported
- wire the target-aware command through MCI mailbox, SPDM VDM, capability reporting, host clients, validators, and typed OTP access
- route MCI mailbox and SPDM VDM execution through the shared `CaliptraCmdHandler::increase_min_svn` backend while keeping transport framing and authorization separate
- enforce monotonic/range checks, `FW_INFO.fw_svn` for Caliptra Runtime, and `CPTRA_CORE_SOC_MANIFEST_MAX_SVN` for SoC Manifest


caliptra-util-host/apps/spdm/client/src/validator.rs
```diff
@@ -961,29 +963,31 @@ pub fn run_provision_owner_pk_hash(
...
     client: &mut SpdmVdmClient,
     flags: u32,
+    target: u32,
     svn: u32,
     authorizer: Option<&dyn CommandAuthChallengeSigner>,
     _verbose: bool,
 ) -> ValidationResult {
-    let test_name = format!("FuseIncreaseCaliptraMinSvn(svn={svn})");
-    let mut payload = Vec::with_capacity(8);
+    let test_name = format!("FuseIncreaseMinSvn(target={target},svn={svn})");
```
parvathib, line 974: The validator checks Caliptra protocol conformance over a transport.  Ideally it would consume the canonical encoding from a central place rather than build its own copy here. 

We're doing the same in the mailbox validator and for several other commands, so any canonical change today means touching multiple sites and keeping duplicate copies in sync. Worth pulling the Caliptra protocol canonical encode/decode into a common layer above the transports. Not blocking for this PR, but good to have a follow up to cleaning up.

docs/src/caliptra_common_commands.md
```diff
@@ -438,11 +431,26 @@ Provisions the vendor public key hash.
...
+| ------ | ------------------ | ----------- | ---------------------------------- |
+| `0`    | Caliptra Runtime   | Implemented | `CPTRA_CORE_RUNTIME_SVN`           |
+| `1`    | SoC Manifest       | Implemented | `CPTRA_CORE_SOC_MANIFEST_SVN`      |
+| `2`    | Owner SoC Manifest | Reserved    | Not implemented                    |
+
+Unknown targets are invalid. The reserved Owner SoC Manifest target returns
+`UnsupportedOperation`. The SVN must be between 1 and 128 and cannot decrease
+the current fuse floor. The Caliptra Runtime target is additionally bounded by
+the running SVN reported by `FW_INFO`. The SoC Manifest target is bounded by
+`CPTRA_CORE_SOC_MANIFEST_MAX_SVN`. No trusted running SoC Manifest SVN is
```
parvathib, line 449: NIT: Do we have a documented provisioning mechanism for CPTRA_CORE_SOC_MANIFEST_MAX_SVN? Since SoC manifest target SVN is bounded by it, it'd be good to capture how integrators are expected to set it. Or point to it if it's already documented.

parvathib (approved):
LGTM overall. I left a few minor comments that can be addressed in a follow-up.

caliptra-util-host/apps/spdm/client/src/validator.rs
mhatrevi: Filed #2132 to track centralizing canonical Caliptra command encoding and decoding across transports.

docs/src/caliptra_common_commands.md
mhatrevi: Tracked the provisioning documentation follow-up in #2133.

